PR #148, opened 2020-08-19
infra: Automatically publish to PyPi

speller26:
*Issue #, if available:*

*Description of changes:*

*Testing done:*

## Merge Checklist

#### General

#### Tests

### DIFF
--- a/.github/workflows/publish-to-pypi.yml
+++ b/.github/workflows/publish-to-pypi.yml
@@ -1,0 +1,26 @@
+name: Publish distribution to PyPI
+
+on:
+  release:
+    types: [published]
+
+jobs:
+  build-and-publish:
+    name: Build and publish distribution 📦to PyPI and TestPyPI
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@main
+      - name: Set up Python 3.7
+        uses: actions/setup-python@v1
+        with:
+          python-version: 3.7
+      - name: Install wheel
+        run: python -m pip install --user --upgrade wheel
+      - name: Install twine
+        run: python -m pip install --user --upgrade twine
+      - name: Build a binary wheel and a source tarball
+        run: python setup.py sdist bdist_wheel
+      - name: Publish distribution to PyPI
+        uses: pypa/gh-action-pypi-publish@master
+        with:
+          password: ${{ secrets.pypi_token }}
--- a/.github/workflows/publish-to-pypi.yml
+++ b/.github/workflows/publish-to-pypi.yml
@@ -6,7 +6,7 @@ on:
 
 jobs:
   build-and-publish:
-    name: Build and publish distribution 📦to PyPI and TestPyPI
+    name: Build and publish distribution to PyPi
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@main
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-## Amazon Braket Python SDK
+# Amazon Braket Python SDK
 
 [![Latest Version](https://img.shields.io/pypi/v/amazon-braket-sdk.svg)](https://pypi.python.org/pypi/amazon-braket-sdk)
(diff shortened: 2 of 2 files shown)


.github/workflows/publish-to-pypi.yml
```diff
@@ -0,0 +1,26 @@
...
+jobs:
+  build-and-publish:
+    name: Build and publish distribution to PyPi
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@main
+      - name: Set up Python 3.7
+        uses: actions/setup-python@v1
+        with:
+          python-version: 3.7
```
avawang1, line 16: This is still publishing to PyPI for 3.7 and 3.8?

.github/workflows/publish-to-pypi.yml
```diff
@@ -0,0 +1,26 @@
+name: Publish distribution to PyPI
+
+on:
+  release:
```
avawang1, line 4: Just to clarify, this is every time there's a new release in https://github.com/aws/amazon-braket-sdk-python/releases?

speller26: Yep

.github/workflows/publish-to-pypi.yml
speller26: It'll still be compatible with 3.8; I haven't really seen packages on PyPi with multiple tarballs and wheels for a single version

.github/workflows/publish-to-pypi.yml
avawang1: Thanks. How do we know it's successful? Just check the PyPI website, or is there a notification/logs we can view?

speller26: Checking PyPi is the end-all-be-all, but there's also the repo's "Actions" tab that will give a history of every run. I think you can configure your account's notifications for Actions too.

avawang1: Is it possible to run integ/unit tests before the PyPI release? For example, instead of having it in Github Actions, and instead triggered by webhook in codebuild?

speller26: In any normal circumstance, the unit/integ tests _will_ run before the PyPi release. When a new version comes in, tests are run, and a new release is only created if they all pass. It's certainly possible to put the PyPi release in that workflow; however, manual releases, if they're ever needed, will also require a corresponding manual publish to PyPi.

avawang1: There's a wait time for alarms after a commit has been merged to main for the unit/integ tests to run, right? It might be that the release is created before sufficient wait time for a problem to emerge from unit/integ tests.

I'm suggesting if we had a webhook trigger for releases, which ran the unit/integ tests and published to PyPI, so it would work for a manual release or when we have an automatic version updating, the automatic one.

speller26: Are you talking about manual releases causing problems? In the automatic case, the release cannot be created unless the tests all pass.

avawang1: Yes, since we only have manual releases right now. Manual as in we have to do "Create a release" in the UI.

What do you mean by "automatic case"?

Basically the benefit of when there's a release, triggering the integ/unit tests to run, then publishing to PyPI.

avawang1: Also, perhaps this can help with running unit/integ tests: https://github.com/aws-actions/aws-codebuild-run-build

speller26: Right, releases are in fact automatic; as soon as the version number is bumped, a release is automatically created. The main future change would be automatic version bumping.

avawang1: What about the release notes for an automatic update? Doesn't that have to be manually entered?

speller26: Yeah, the way it currently stands, the release notes would have to be post-edited.

avawang1: OK, from the point of view of manual releases, I think once we have integ tests running as part of the PR builds we should be good.

PR #149, opened 2020-08-19
feature: Add noise operations and methods to add noise to circuits

yitinche:
Description of changes: 
Add operations and a result type (density matrix) that are related to noise simulation. 
Add add_noise method to Circuits class.

Build directory: [build_files.tar.gz](https://github.com/aws/amazon-braket-sdk-python/files/5104499/build_files.tar.gz)

## Merge Checklist

#### General

- [ x ] I have read the [CONTRIBUTING](https://github.com/aws/amazon-braket-sdk-python/blob/main/CONTRIBUTING.md) doc
- [ x ] I used the commit message format described in [CONTRIBUTING](https://github.com/aws/amazon-braket-sdk-python/blob/main/CONTRIBUTING.md#commit-your-change)
- [ x ] I have updated any necessary documentation, including [READMEs](https://github.com/aws/amazon-braket-sdk-python/blob/main/README.md) and [API docs](https://github.com/aws/amazon-braket-sdk-python/blob/main/CONTRIBUTING.md#documentation-guidelines) (if appropriate)

#### Tests

- [ x ] I have added tests that prove my fix is effective or that my feature works (if appropriate)
- [ x ] I have checked that my tests are not configured for a specific region or account (if appropriate)

### DIFF
--- a/src/braket/circuits/circuit.py
+++ b/src/braket/circuits/circuit.py
@@ -23,6 +23,8 @@ from braket.circuits.observables import TensorProduct
 from braket.circuits.qubit import QubitInput
 from braket.circuits.qubit_set import QubitSet, QubitSetInput
 from braket.circuits.result_type import ObservableResultType, ResultType
+from braket.circuits.noise import Noise
+from braket.circuits.noise_helpers import _add_noise
 from braket.ir.jaqcd import Program
 
 SubroutineReturn = TypeVar(
@@ -437,6 +439,138 @@ class Circuit:
             self.add_result_type(result_type, target_mapping=target_mapping)
 
         return self
+
+    def add_noise(self,
+                   noise: Noise,
+                   target_gates: Union[str, List[str]] = None,
+                   target_qubits: QubitSetInput = None,
+                   target_times: Union[int, List[int]] = None,
+                   insert_strategy: str = "inclusive",
+    )->Circuit:
+        """
+        Add the provided `noise` to gates, qubits and time specified by `target_gates`,
+         `target_qubits` and `target_times`. When `target_gates`=None, `noise` is added
+        to the specified qubits and time. When `target_qubits`=None, `noise` is added to
+        the specified gates and time at all qubits. When `target_times`=None, `noise` is
+        added to the specified gates and qubits at all time. If all `target_gates`,
+        `target_qubits` and `target_times` are None, `noise` is added to every qubit at
+        every time in a circuit.
+
+        When `target_gates` is not None and `noise.qubit_count`>1, `noise.qubit_count`
+        must be the same as `qubit_count` of gates specified by `target_gates`. When
+        `noise.qubit_count`==1, ie. `noise` is a single-qubit noise, `noise` is added
+        to all qubits in `target_qubits` that interact with the target gates.
+
+        Args:
+            noise (Noise): Noise to be added to the circuit. When `noise.qubit_count`>1,
+                `noise.qubit_count` must be the same as `qubit_count` of gates specified by
+                `target_gates`.
+            targer_gates (Union[str, List[str]): Name or List of name of gates which
+                `noise` is added to. If None, `noise` is added only according to
+                `target_qubits` and `target_times`. None should be used when users want
+                to add `noise` to a ciruit moment that has no gate.
+            target_qubits (QubitSetInput): Index or indices of qubit(s). When `target_gates`
+                is not None, the usage of `target_qubits` is determined by `insert_strategy`.
+                Default=None.
+            target_times (Union[int, List[int]]): Index of indices of time which `noise`
+                is added to. Default=None.
+            insert_strategy (str): Rule of how `target_qubit` is used. `insert_strategy`
+                is usded only when `target_gates` is not None. Default="inclusive".
+                Options:
+                    "strict": Insert noise to a gate when `gate.target` exactly matches
+                        `target_qubits`. Sensitive to the order of qubits.
+                    "inclusive": Insert noise to a gate when `gate.target` is a subset
+                        of `target_qubits`.
+
+        Returns:
+            Circuit: self
+
+        Raises:
+                TypeError: If `noise` is not Noise type, `target_gates` is not str, List[str] or None,
+                `target_times` is not int or List[int].
+
+        Example:
+        >>> circ = Circuit().x(0).y(1).z(0).x(1).cnot(0,1)
+        >>> print(circ)
+            T  : |0|1|2|
+            q0 : -X-Z-C-
+                      |
+            q1 : -Y-X-X-
+            T  : |0|1|2|
+        >>> noise = Noise.Bit_Flip(prob=0.1)
+        >>> circ = Circuit().x(0).y(1).z(0).x(1).cnot(0,1)
+        >>> print(circ.add_noise(noise, target_gates = 'X'))
+            T  : |    0    |    1    |2|
+            q0 : -X-NB(0.1)-Z---------C-
+                                      |
+            q1 : -Y---------X-NB(0.1)-X-
+            T  : |    0    |    1    |2|
+        >>> circ = Circuit().x(0).y(1).z(0).x(1).cnot(0,1)
+        >>> print(circ.add_noise(noise, target_qubits = 1))
+            T  : |    0    |    1    |    2    |
+            q0 : -X---------Z---------C---------
+                                      |
+            q1 : -Y-NB(0.1)-X-NB(0.1)-X-NB(0.1)-
+            T  : |    0    |    1    |    2    |
+        >>> circ = Circuit().x(0).y(1).z(0).x(1).cnot(0,1)
+        >>> print(circ.add_noise(noise, target_gates = 'X', target_qubits = 1))
+            T  : |0|    1    |2|
+            q0 : -X-Z---------C-
+                              |
+            q1 : -Y-X-NB(0.1)-X-
+            T  : |0|    1    |2|
+        >>> circ = Circuit().x(0).y(1).z(0).x(1).cnot(0,1)
+        >>> print(circ.add_noise(noise, target_gates = ['X','Y'], target_qubits = [0,1]))
+            T  : |    0    |    1    |2|
+            q0 : -X-NB(0.1)-Z---------C-
+                                      |
+            q1 : -Y-NB(0.1)-X-NB(0.1)-X-
+            T  : |    0    |    1    |2|
+        >>> circ = Circuit().x(0).y(1).z(0).x(1).cnot(0,1)
+        >>> print(circ.add_noise(noise, target_times=[0,2]))
+            T  : |    0    |1|    2    |
+            q0 : -X-NB(0.1)-Z-C-NB(0.1)-
+                              |
+            q1 : -Y-NB(0.1)-X-X-NB(0.1)-
+            T  : |    0    |1|    2    |
+        >>> circ = Circuit().x(0).y(1).z(0).x(1).cnot(0,1)
+        >>> print(circ.add_noise(noise,
+        ...                      target_gates = ['X','Y'],
+        ...                      target_qubits = [0,1],
+        ...                      target_times=1)
+        ... )
+            T  : |0|    1    |2|
+            q0 : -X-Z---------C-
+                              |
+            q1 : -Y-X-NB(0.1)-X-
+            T  : |0|    1    |2|
+
+        """
+
+        if isinstance(target_gates, str):
+            target_gates = [target_gates]
+        if target_qubits is None:
+            target_qubits = QubitSet(range(self.qubit_count))
+        else:
+            target_qubits = QubitSet(target_qubits)
+        if target_times is None:
+            target_times = range(self.depth)
+        if isinstance(target_times, int):
+            target_times = [target_times]
+
+        if not isinstance(noise, Noise):
+            raise TypeError("noise must be a Noise class")
+        if target_gates and not all(isinstance(s, str) for s in target_gates):
+            raise TypeError(f"all elements in {target_gates} must be str")
+        if not all(isinstance(time, int) for time in target_times):
+            raise TypeError("target_times must be int or List[int]")
+
+        return _add_noise(self, noise,
+                                target_gates,
+                                target_qubits,
+                                target_times,
+                                insert_strategy)
+
 
     def add(self, addable: AddableTypes, *args, **kwargs) -> Circuit:
         """
--- a/src/braket/circuits/circuit.py
+++ b/src/braket/circuits/circuit.py
@@ -13,7 +13,7 @@
 
 from __future__ import annotations
 
-from typing import Callable, Dict, Iterable, List, TypeVar, Union
+from typing import Callable, Dict, Iterable, List, Optional, TypeVar, Union
 
 from braket.circuits import noise_helpers
 from braket.circuits.ascii_circuit_diagram import AsciiCircuitDiagram
@@ -443,11 +443,13 @@ class Circuit:
     def add_noise(
         self,
         noise: Noise,
-        target_gates: Union[str, Iterable[str]] = None,
-        target_qubits: QubitSetInput = None,
-        target_times: Union[int, Iterable[int]] = None,
+        target_gates: Optional[Union[str, Iterable[str]]] = None,
+        target_qubits: Optional[QubitSetInput] = None,
+        target_times: Optional[Union[int, Iterable[int]]] = None,
     ) -> Circuit:
-        """
+        """ Add `noise` to the circuit according to `target_gates`, `target_qubits` and
+        `target_times`.
+
         For any parameter that is None, that specification is ignored (e.g. if 'target_gates'
         and 'target_qubits' are None then the noise is added to all qubits at target_times).
         If 'target_gates', 'target_qubits', and 'target_times' are all None, then 'noise' is
@@ -477,7 +479,9 @@ class Circuit:
 
         Raises:
                 TypeError: If `noise` is not Noise type, `target_gates` is not str,
-                Iterable[str] or None, `target_times` is not int or Iterable[int].
+                    Iterable[str] or None, `target_times` is not int or Iterable[int],
+                    len(target_gates) is not equal to noise.qubit_count when target_gates is
+                    not None
 
         Example:
             >>> circ = Circuit().x(0).y(1).z(0).x(1).cnot(0,1)
@@ -541,7 +545,8 @@ class Circuit:
             target_gates = [target_gates]
         if isinstance(target_times, int):
             target_times = [target_times]
-        target_qubits = QubitSet(target_qubits)
+        if target_qubits is not None:
+            target_qubits = QubitSet(target_qubits)
 
         if not isinstance(noise, Noise):
             raise TypeError("noise must be a Noise class")
--- a/src/braket/circuits/circuit.py
+++ b/src/braket/circuits/circuit.py
@@ -150,9 +150,11 @@ class Circuit:
         added_observables_targets = set()
         for return_type in observable_return_types:
             observable: Observable = return_type.observable
-            targets: List[List[int]] = [list(return_type.target)] if return_type.target else [
-                list([qubit]) for qubit in self._moments.qubits
-            ]
+            targets: List[List[int]] = (
+                [list(return_type.target)]
+                if return_type.target
+                else [list([qubit]) for qubit in self._moments.qubits]
+            )
 
             for target in targets:
                 # only add gates for observables and targets that
@@ -447,7 +449,7 @@ class Circuit:
         target_qubits: Optional[QubitSetInput] = None,
         target_times: Optional[Union[int, Iterable[int]]] = None,
     ) -> Circuit:
-        """ Add `noise` to the circuit according to `target_gates`, `target_qubits` and
+        """Add `noise` to the circuit according to `target_gates`, `target_qubits` and
         `target_times`.
 
         For any parameter that is None, that specification is ignored (e.g. if 'target_gates'
--- a/src/braket/circuits/circuit.py
+++ b/src/braket/circuits/circuit.py
@@ -15,11 +15,11 @@ from __future__ import annotations
 
 from typing import Callable, Dict, Iterable, List, TypeVar, Union
 
+from braket.circuits import noise_helpers
 from braket.circuits.ascii_circuit_diagram import AsciiCircuitDiagram
 from braket.circuits.instruction import Instruction
 from braket.circuits.moments import Moments
 from braket.circuits.noise import Noise
-from braket.circuits import noise_helpers
 from braket.circuits.observable import Observable
 from braket.circuits.observables import TensorProduct
 from braket.circuits.qubit import QubitInput
@@ -566,11 +566,9 @@ class Circuit:
         if not noise_helpers.type_check_target_times(target_times):
             raise TypeError("target_times must be int or Iterable[int]")
 
-        return noise_helpers._add_noise(self, noise,
-                                        target_gates,
-                                        target_qubits,
-                                        target_times,
-                                        insert_strategy)
+        return noise_helpers._add_noise(
+            self, noise, target_gates, target_qubits, target_times, insert_strategy
+        )
 
     def add(self, addable: AddableTypes, *args, **kwargs) -> Circuit:
         """
--- a/src/braket/circuits/circuit.py
+++ b/src/braket/circuits/circuit.py
@@ -15,12 +15,12 @@ from __future__ import annotations
 
 from typing import Callable, Dict, Iterable, List, Optional, Type, TypeVar, Union
 
-from braket.circuits import noise_helpers
 from braket.circuits.ascii_circuit_diagram import AsciiCircuitDiagram
 from braket.circuits.gate import Gate
 from braket.circuits.instruction import Instruction
 from braket.circuits.moments import Moments
 from braket.circuits.noise import Noise
+from braket.circuits.noise_helpers import add_noise_to_gates, add_noise_to_moments
 from braket.circuits.observable import Observable
 from braket.circuits.observables import TensorProduct
 from braket.circuits.qubit import QubitInput
@@ -485,7 +485,7 @@ class Circuit:
             TypeError: If `noise` is not Noise type, `target_gates` is not a Gate class,
                 Iterable[Gate] or None, `target_times` is not int or Iterable[int] or None.
             ValueError:
-                If `noise.qubit_count` is not equal to `len(gate.target)` of gates in
+                If `noise.qubit_count` is not equal to `len(gate.target)` for gates in
                 `target_gates`. If `noise.qubit_count` is not equal to `len(target_qubits)`
                 when `target_gates` is None.
 
@@ -554,7 +554,10 @@ class Circuit:
             ):
                 raise TypeError("all elements in target_times must be int")
 
-        return noise_helpers._add_noise(self, noise, target_gates, target_qubits, target_times)
+        if target_gates is None:
+            return add_noise_to_moments(self, noise, target_qubits, target_times)
+        else:
+            return add_noise_to_gates(self, noise, target_gates, target_qubits, target_times)
 
     def add(self, addable: AddableTypes, *args, **kwargs) -> Circuit:
         """
--- a/src/braket/circuits/circuit.py
+++ b/src/braket/circuits/circuit.py
@@ -13,10 +13,11 @@
 
 from __future__ import annotations
 
-from typing import Callable, Dict, Iterable, List, Optional, TypeVar, Union
+from typing import Callable, Dict, Iterable, List, Optional, Type, TypeVar, Union
 
 from braket.circuits import noise_helpers
 from braket.circuits.ascii_circuit_diagram import AsciiCircuitDiagram
+from braket.circuits.gate import Gate
 from braket.circuits.instruction import Instruction
 from braket.circuits.moments import Moments
 from braket.circuits.noise import Noise
@@ -445,7 +446,7 @@ class Circuit:
     def add_noise(
         self,
         noise: Noise,
-        target_gates: Optional[Union[str, Iterable[str]]] = None,
+        target_gates: Optional[Union[Type[Gate], Iterable[Type[Gate]]]] = None,
         target_qubits: Optional[QubitSetInput] = None,
         target_times: Optional[Union[int, Iterable[int]]] = None,
     ) -> Circuit:
@@ -466,13 +467,11 @@ class Circuit:
             noise (Noise): Noise to be added to the circuit. When `noise.qubit_count` > 1,
                 `noise.qubit_count` must be the same as `qubit_count` of gates specified by
                 `target_gates`.
-            target_gates (Union[str, Iterable[str], optional]): Name or List of name of gates
-                which `noise` is added to. If None, `noise` is added only according to
-                `target_qubits` and `target_times`. None should be used when users want
-                to add `noise` to a ciruit moment that has no gate.
-            target_qubits (Union[QubitSetInput, optional]): Index or indices of qubit(s). When
-                `target_gates` is not None, the usage of `target_qubits` is determined by
-                `insert_strategy`. Default=None.
+            target_gates (Union[Type[Gate], Iterable[Type[Gate]], optional]): Gate class or
+                List of Gate classes which `noise` is added to. If None, `noise` is added only
+                according to `target_qubits` and `target_times`.
+            target_qubits (Union[QubitSetInput, optional]): Index or indices of qubit(s).
+                Default=None.
             target_times (Union[int, Iterable[int], optional]): Index of indices of time which
                 `noise` is added to. Default=None.
 
@@ -480,10 +479,10 @@ class Circuit:
             Circuit: self
 
         Raises:
-                TypeError: If `noise` is not Noise type, `target_gates` is not str,
-                    Iterable[str] or None, `target_times` is not int or Iterable[int],
-                    len(target_gates) is not equal to noise.qubit_count when target_gates is
-                    not None
+            TypeError: If `noise` is not Noise type, `target_gates` is not a Gate class,
+                Iterable[Gate] or None, `target_times` is not int or Iterable[int],
+                len(target_gates) is not equal to noise.qubit_count when target_gates is
+                not None
 
         Example:
             >>> circ = Circuit().x(0).y(1).z(0).x(1).cnot(0,1)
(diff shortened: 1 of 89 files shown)


yitinche:
The CodeBuild CI fails because this repo has dependency on the "noise" branch of Schemas repo. I have checked that the CI will pass with the noise branch of Schemas.

src/braket/circuits/circuit.py
```diff
@@ -438,6 +440,136 @@ def add_circuit(
 
         return self
 
+    def add_noise(
+        self,
+        noise: Noise,
+        target_gates: Union[str, Iterable[str]] = None,
+        target_qubits: QubitSetInput = None,
+        target_times: Union[int, Iterable[int]] = None,
```
licedric, line 448: I'd rename this target_moments, since we have a concept of moments but not time.

src/braket/circuits/circuit.py
```diff
@@ -438,6 +440,136 @@ def add_circuit(
...
+        insert_strategy: str = "inclusive",
+    ) -> Circuit:
+        """
+        Add the provided `noise` to gates, qubits and time specified by `target_gates`,
+         `target_qubits` and `target_times`. When `target_gates`=None, `noise` is added
```
licedric, line 453: I'd simplify this: "For any parameter that is None, that specification is ignored (e.g. if 'target_gates' and 'target_qubits' are None then the noise is added to all qubits at target_times). If 'target_gates', 'target_qubits', and 'target_times' are all None, then 'noise' is added to every qubit at every moment."

src/braket/circuits/circuit.py
```diff
@@ -438,6 +440,136 @@ def add_circuit(
...
+        to the specified qubits and time. When `target_qubits`=None, `noise` is added to
+        the specified gates and time at all qubits. When `target_times`=None, `noise` is
+        added to the specified gates and qubits at all time. If all `target_gates`,
+        `target_qubits` and `target_times` are None, `noise` is added to every qubit at
+        every time in a circuit.
+
+        When `target_gates` is not None and `noise.qubit_count`>1, `noise.qubit_count`
```
licedric, line 460: Format expressions in comment like you would in code: `noise.qubit_count` > 1

src/braket/circuits/circuit.py
```diff
@@ -438,6 +440,136 @@ def add_circuit(
...
+        must be the same as `qubit_count` of gates specified by `target_gates`. When
+        `noise.qubit_count`==1, ie. `noise` is a single-qubit noise, `noise` is added
```
licedric, line 462: i.e. `noise` is single-qubit

src/braket/circuits/circuit.py
```diff
@@ -438,6 +440,136 @@ def add_circuit(
...
+                `noise.qubit_count` must be the same as `qubit_count` of gates specified by
+                `target_gates`.
+            targer_gates (Union[str, Iterable[str]): Name or List of name of gates which
+                `noise` is added to. If None, `noise` is added only according to
+                `target_qubits` and `target_times`. None should be used when users want
+                to add `noise` to a ciruit moment that has no gate.
+            target_qubits (QubitSetInput): Index or indices of qubit(s). When `target_gates`
+                is not None, the usage of `target_qubits` is determined by `insert_strategy`.
+                Default=None.
+            target_times (Union[int, Iterable[int]]): Index of indices of time which `noise`
```
licedric, line 476: Again, use moment here.

src/braket/circuits/circuit.py
```diff
@@ -438,6 +440,136 @@ def add_circuit(
...
+                is added to. Default=None.
+            insert_strategy (str): Rule of how `target_qubit` is used. `insert_strategy`
+                is usded only when `target_gates` is not None. Default="inclusive".
```
licedric, line 479: used

src/braket/circuits/circuit.py
```diff
@@ -438,6 +440,136 @@ def add_circuit(
...
+                Options:
+                    "strict": Insert noise to a gate when `gate.target` exactly matches
+                        `target_qubits`. Sensitive to the order of qubits.
+                    "inclusive": Insert noise to a gate when `gate.target` is a subset
+                        of `target_qubits`.
+
+        Returns:
+            Circuit: self
+
+        Raises:
```
licedric, line 489: What about when qubit count of noise doesn't match target_gates?

src/braket/circuits/noise.py
```diff
@@ -0,0 +1,139 @@
+from typing import Any, List, Sequence
+
+from braket.circuits.operator import Operator
+from braket.circuits.qubit_set import QubitSet
+
+
+class Noise(Operator):
+    """
+    Class `Noise` represents a noise channel that operates on N qubits. Noise are considered as
```
licedric, line 9: "that operates on N qubits" is unnecessary (what's N?)

src/braket/circuits/noise.py
```diff
@@ -0,0 +1,139 @@
...
+
+    def __init__(
+        self, qubit_count: int, ascii_symbols: Sequence[str],
+    ):
+        """
+        Args:
+            qubit_count (int): Number of qubits this noise channel interacts with.
+            ascii_symbols (Sequence[str]): ASCII string symbols for this noise channel. These are
+                used when printing a diagram of circuits. Length must be the same as `qubit_count`
+                , and index ordering is expected to correlate with target ordering on the
```
licedric, line 24: Comma should be on the same line as preceding word

src/braket/circuits/noise.py
```diff
@@ -0,0 +1,139 @@
...
+            ValueError: If the `qubit_count` is less than 1, `ascii_symbols` are `None`, or
+                `ascii_symbols` length != `qubit_count`, `prob` is not `float`, `prob`>1.0, pr
+                `prob`<0.0
+        """
+        super().__init__(qubit_count=qubit_count, ascii_symbols=ascii_symbols)
+
+        if not isinstance(prob, float):
+            raise ValueError("prob must be float type")
+        if not (prob <= 1.0 and prob >= 0.0):
+            raise ValueError("prob must a real number in the interval [0,1]")
```
licedric, line 126: must be

src/braket/circuits/circuit.py
```diff
@@ -438,6 +440,136 @@ def add_circuit(
 
         return self
 
+    def add_noise(
+        self,
+        noise: Noise,
+        target_gates: Union[str, Iterable[str]] = None,
```
licedric, line 446: All of the parameters should be optional (e.g. target_gates: Optional[Union[str, Iterable[str]]]

src/braket/circuits/noise.py
```diff
@@ -0,0 +1,139 @@
...
+    def register_noise(cls, noise: "Noise"):
+        """Register a noise implementation by adding it into the Noise class.
+
+        Args:
+            noise (Noise): Noise class to register.
+        """
+        setattr(cls, noise.__name__, noise)
+
+
+class ProbabilityNoise(Noise):
```
licedric, line 101: I'd name this ProbabilisticNoise to be grammatically correct.

src/braket/circuits/noise.py
```diff
@@ -0,0 +1,139 @@
...
+                index ordering is expected to correlate with the target ordering on the instruction.
+
+        Raises:
+            ValueError: If the `qubit_count` is less than 1, `ascii_symbols` are `None`, or
+                `ascii_symbols` length != `qubit_count`, `prob` is not `float`, `prob`>1.0, pr
+                `prob`<0.0
+        """
+        super().__init__(qubit_count=qubit_count, ascii_symbols=ascii_symbols)
+
+        if not isinstance(prob, float):
```
licedric, line 123: Can you test if this works if I pass it a numpy float type (e.g. np.double or np.float64)?

src/braket/circuits/noise_helpers.py
```diff
@@ -0,0 +1,220 @@
...
+    circ: Circuit,
+    noise: Noise,
+    target_gates: Iterable[str],
+    target_qubits: QubitSetInput,
+    target_times: Iterable[int],
+    insert_strategy: str,
+) -> Circuit:
+    """ Insert noise into circuit.
+    Add the provided `noise` to gates, qubits and time specified by `target_gates`
+    , `target_qubits` and `target_times`. See the description of circuit.add_noise().
```
licedric, line 24: Comma should be on previous line

src/braket/circuits/noise_helpers.py
```diff
@@ -0,0 +1,220 @@
...
+
+    Args:
+        circ (Circuit): A ciruit where `noise` is added to.
```
licedric, line 27: circuit

src/braket/circuits/noise_helpers.py
```diff
@@ -0,0 +1,220 @@
...
+from braket.circuits.qubit_set import QubitSet, QubitSetInput
+
+if TYPE_CHECKING:  # pragma: no cover
+    from braket.circuits.circuit import Circuit
+
+
+def _add_noise(
+    circ: Circuit,
+    noise: Noise,
+    target_gates: Iterable[str],
```
licedric, line 17: These again should be Optional

src/braket/circuits/noise_helpers.py
```diff
@@ -0,0 +1,220 @@
...
+
+def _add_noise_to_qubits(
+    circ: Circuit, noise: Noise, target_qubits: QubitSet, target_times: Iterable[int]
+) -> Circuit:
+    """ Insert noise to the given time and qubits.
+
+    Args:
+        circ (Circuit): A ciruit where `noise` is added to.
+        noise (Noise): A Noise class object to be added to the circuit.
+        target_qubits (QubitSet): Index or indices of qubit. `noise` is added to.
```
licedric, line 64: qubits, not qubit.

src/braket/circuits/circuit.py
licedric: Scratch this, there is time in moments. Weird...

src/braket/circuits/circuit.py
licedric: Again, never mind.

src/braket/circuits/noise_helpers.py
```diff
@@ -0,0 +1,220 @@
...
+    )
+
+
+class NoiseInsertStrategy:
+    """ Base class of noise insert strategy.
+    """
+
+    @staticmethod
+    def _add_1qubit_noise():
+        raise NotImplementedError("_add_1qubit_noise has not been implemented yet.")
```
licedric, line 156: I'd replace "has not been implemented yet" with "is not implemented".

src/braket/circuits/noises.py
```diff
@@ -0,0 +1,252 @@
...
+            Instruction(Noise.Amplitude_Damping(prob=prob), target=qubit)
+            for qubit in QubitSet(target)
+        ]
+
+
+Noise.register_noise(Amplitude_Damping)
+
+
+class Kraus(Noise):
+    """User-defined noise channel
```
licedric, line 183: Can you clarify what the noise channel is, given the input matrices?

src/braket/circuits/noises.py
```diff
@@ -0,0 +1,252 @@
...
+To add a new noise:
+    1. Implement the class and extend `Noise`
+    2. Add a method with the `@circuit.subroutine(register=True)` decorator. Method name
+       will be added into the `Circuit` class. This method is the default way
+       clients add this noise to a circuit.
+    3. Register the class with the `Noise` class via `Noise.register_noise()`.
+"""
+
+
+class Bit_Flip(ProbabilityNoise):
```
avawang1, line 25: Python class naming convention is "BitFlip", same for the other classes

avawang1:
Could you attach a zip file of the build directory, using `tox -e zip-build` to the description of the PR?

yitinche:
> Could you attach a zip file of the build directory, using `tox -e zip-build` to the description of the PR?
Done!

src/braket/circuits/noise.py
yitinche: I refer the convention of Gate() and AngleGate(). Open to change but let's see how other dev think about this.

src/braket/circuits/noise.py
yitinche: yes, it works

src/braket/circuits/noise_helpers.py
yitinche: They are actually not optional for this function. Default values are passed to this function if users don't specify.

src/braket/circuits/noises.py
yitinche: Fixed the class names. How about the circuit method Circuit.bit_flip() when adding this noise to a circuit? Should it be "Circuit().bit_flip()" or "Circuit().bitflip()" ?

src/braket/circuits/circuit.py
yitinche: Just to confirm, do you mean the spaces between "noise.qubit_count", ">", and  "1"?

src/braket/circuits/circuit.py
yitinche: The input is the name of the gates (str type). It is not straightforward to check the qubit_count through their name. An idea I have is using `gate = getattr(Gate, name)` and then check the qubit count. Does this looks okay?

src/braket/circuits/circuit.py
licedric: Yeah.

src/braket/circuits/circuit.py
```diff
@@ -438,6 +440,136 @@ def add_circuit(
...
+        When `target_gates` is not None and `noise.qubit_count`>1, `noise.qubit_count`
+        must be the same as `qubit_count` of gates specified by `target_gates`. When
+        `noise.qubit_count`==1, ie. `noise` is a single-qubit noise, `noise` is added
+        to all qubits in `target_qubits` that interact with the target gates.
+
+        Args:
+            noise (Noise): Noise to be added to the circuit. When `noise.qubit_count`>1,
+                `noise.qubit_count` must be the same as `qubit_count` of gates specified by
+                `target_gates`.
+            targer_gates (Union[str, Iterable[str]): Name or List of name of gates which
```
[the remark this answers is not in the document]
licedric: +1

src/braket/circuits/circuit.py
licedric: As Kshitij suggested above, it's easier to pass in the gate and noise objects directly instead of using strings.

src/braket/circuits/noise_helpers.py
```diff
@@ -0,0 +1,220 @@
...
+            `target_times`. None should be used when users want to add `noise` to
+            a ciruit moment that has no gate.
+        target_qubits (QubitSet): Index or indices of qubit. When `target_gates` is
+            not None, the usage of `target_qubits` is determined by `insert_strategy`.
+        target_times (Iterable[int]): List of time which `noise` is added to.
+        insert_strategy (str): Rule of how `target_qubit` is used. `insert_strategy`
+            is usded only when `target_gates` is not None.
+            "strict": Insert noise to a gate when `gate.target` exactly matches
+                `target_qubits`. Sensitive to order of qubits.
+            "inclusive":Insert noise to a gate when the target of the gate is a subset
```
licedric, line 40: Space after "inclusive"

src/braket/circuits/circuit.py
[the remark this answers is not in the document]
yitinche: For the existing code, users can specify the time and qubit a specific unitary locates. 
The Gate objects do not have unique name. So it is not straightforward to compare to gates in general. For example, when comparing two Rx gates, we can compare their angle. But for for differently parametrized gates, we might want to compare something else. We could always compare the equivalence of matrices, but not sure that leads to the best customer experience because defining a matrix is a bit cumbersome. There are a few other options, but I don't think we  have enough data to determine which behavior fits customer needs the best. So I just kept it simple, and only allow specifying a specific gate through gate name, time and qubit. But we can always add new functionality later.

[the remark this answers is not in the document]
yitinche: Just to confirm, do you mean the class themselves (without initialize, like Noise.BitFlip) or the class object (initialized, like Noise.BitFlip())?

src/braket/circuits/circuit.py
yitinche: Follow-up question: We could allow user pass a gate class (e.g. Gate.Rx) or a gate object (e.g. Gate.Rx(angle=0.1)). In the former option, we can't actually access `qubit_count` from Gate.X. In the later option, do you think user would expect all properties of gate have to match or just the name?

src/braket/circuits/circuit.py
```diff
@@ -438,6 +440,136 @@ def add_circuit(
 
         return self
 
+    def add_noise(
+        self,
+        noise: Noise,
+        target_gates: Union[str, Iterable[str]] = None,
+        target_qubits: QubitSetInput = None,
+        target_times: Union[int, Iterable[int]] = None,
+        insert_strategy: str = "inclusive",
```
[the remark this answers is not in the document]
yitinche: I can work on removing "insert_strategy". Just note that after we remove this, there will be no way to add different noise that is sensitive to qubit order. For example, we won't have the ability to add different noises for CNot(1,0) and CNot(0,1).

src/braket/circuits/noises.py
```diff
@@ -0,0 +1,252 @@
...
+        return [
+            Instruction(Noise.Depolarizing(prob=prob), target=qubit) for qubit in QubitSet(target)
+        ]
+
+
+Noise.register_noise(Depolarizing)
+
+
+class Amplitude_Damping(ProbabilityNoise):
+    """Phase flip noise channel."""
```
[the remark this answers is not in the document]
yitinche: Good point! Will add.

src/braket/circuits/circuit.py
```diff
@@ -438,6 +440,136 @@ def add_circuit(
...
+        if not noise_helpers.type_check_target_gates(target_gates):
+            raise TypeError(f"all elements in {target_gates} must be str")
+        if not noise_helpers.type_check_target_times(target_times):
+            raise TypeError("target_times must be int or Iterable[int]")
+
+        return noise_helpers._add_noise(
+            self, noise, target_gates, target_qubits, target_times, insert_strategy
+        )
+
     def add(self, addable: AddableTypes, *args, **kwargs) -> Circuit:
```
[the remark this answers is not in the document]
yitinche: okay, working in progress

src/braket/circuits/circuit.py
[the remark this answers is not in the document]
kshitijc: I was thinking the class themselves (without initializing, like Gate.X), but hadn't given passing in the objects any thought. Do you see a strong advantage in passing in the objects instead ?

src/braket/circuits/circuit.py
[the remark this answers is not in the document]
kshitijc: Would the use of target_times and target_qubits (as you highlighted for the Unitary case) to differentiate the CNot(0, 1) from CNot(1, 0) be a possibility ?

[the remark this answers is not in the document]
licedric: I synced with Tim offline about this - it looks like order _is_ taken into account when adding noise. For instance, if I add (I X ND) as noise to all CNot gates, then CNot(0, 1) will have ND on qubit 1, while CNot(1, 0) will have ND on qubit 0.

src/braket/circuits/noises.py
avawang1: For method names, it should conform to the Python standard, so bit_flip()

src/braket/circuits/circuit.py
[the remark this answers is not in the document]
avawang1: The gates are also passed in as objects for instructions. The reason they are passed in as objects, is to allow passing in of variables to the object initiation, like the angle to AngledGate.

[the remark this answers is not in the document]
yitinche: I agree, I also need the qubit_count of a gate for validation before actually adding noise. The concern I have is that, when users pass in an object as `target_gates`, they need to specify parameters (e.g. angle) that is not used anywhere because only the type of the gate is compared.

src/braket/circuits/moments.py
```diff
@@ -138,16 +147,35 @@ def add(self, instructions: Iterable[Instruction]) -> None:
...
+                [0,]
+                + [
+                    k.noise_index
+                    for k in self._moments
+                    if (k.time == time and isinstance(k, NoiseMomentsKey))
+                ]
+            )
+
+            noise_key = NoiseMomentsKey(time, instruction.target, max_noise_index + 1)
+            self._moments[noise_key] = instruction
```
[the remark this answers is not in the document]
yitinche: The definition of NoiseMomentKey is to achieve the user experience that adding a noise do not create a new time stamp. In other word, adding noise to a circuit does not change the MomentKey and Instructions of the existing gates. 
Therefore, I use this NoiseMomentKey class which has additional parameter `noise_index` so that a gate and many noises can appear in the same time, as long as their `noise_index` are different.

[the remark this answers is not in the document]
yitinche: Is it possible to change the type hint of the Moments class so that its key can be a set of types?

src/braket/circuits/circuit.py
avawang1: As a user, if the gate object is passed in, I'd expect that all properties would match, not just the class name.

src/braket/circuits/circuit.py
```diff
@@ -438,6 +440,134 @@ def add_circuit(
...
+    def add_noise(
+        self,
+        noise: Noise,
+        target_gates: Union[str, Iterable[str]] = None,
+        target_qubits: QubitSetInput = None,
+        target_times: Union[int, Iterable[int]] = None,
+        insert_strategy: str = "inclusive",
+    ) -> Circuit:
+        """
+        For any parameter that is None, that specification is ignored (e.g. if 'target_gates'
```
speller26, line 452: Add an introductory paragraph that just describes what this method does (adding noise, and what that means)

src/braket/circuits/noise.py
```diff
@@ -0,0 +1,140 @@
...
+
+from braket.circuits.quantum_operator import QuantumOperator
+from braket.circuits.qubit_set import QubitSet
+
+
+class Noise(QuantumOperator):
+    """
+    Class `Noise` represents a noise channel that operates on one or multiple qubits. Noise
+    are considered as building blocks of quantum circuits that simulate noise. It can be
+    used as an operator in an Instruction() object. It appears in the diagram when user prints
```
speller26, line 11: Instruction() -> \`Instruction\`

src/braket/circuits/noise.py
```diff
@@ -0,0 +1,140 @@
...
+
+    @property
+    def qubit_count(self) -> int:
+        """int: Returns number of qubits this quantum operator interacts with."""
+        return self._qubit_count
+
+    @property
+    def ascii_symbols(self) -> List[str]:
+        """List[str]: Returns the ascii symbols for the quantum operator."""
+        return self._ascii_symbols
```
speller26, line 51: No need to redefine these, since they're already in `QuantumOperator`

src/braket/circuits/noise.py
speller26: I agree with Cedric here, `ProbabilisticNoise` works better, especially since there's a `Probabilities` result type, and we need users to know these classes aren't related.

src/braket/circuits/noise.py
```diff
@@ -0,0 +1,140 @@
...
+        Args:
+            noise (Noise): Noise class to register.
+        """
+        setattr(cls, noise.__name__, noise)
+
+
+class ProbabilityNoise(Noise):
+    """
+    Class `ProbabilityNoise` represents a noise channel that operates on N qubits
+    and is parameterized by a probability.
```
speller26, line 104: A noise channel on N qubits parameterized ...
Also, add some more description as to how the probability value is used.

src/braket/circuits/noises.py
```diff
@@ -0,0 +1,307 @@
...
+from braket.circuits.instruction import Instruction
+from braket.circuits.noise import Noise, ProbabilityNoise
+from braket.circuits.quantum_operator_helpers import (
+    is_CPTP,
+    verify_quantum_operator_matrix_dimensions,
+)
+from braket.circuits.qubit_set import QubitSet, QubitSetInput
+
+"""
+To add a new noise:
```
speller26, line 16: To add a new `Noise` implementation:

src/braket/circuits/noises.py
```diff
@@ -0,0 +1,307 @@
...
+            probability=probability,
+            qubit_count=1,
+            ascii_symbols=["NB({:.2g})".format(probability)],
+        )
+
+    def to_ir(self, target: QubitSet):
+        return ir.BitFlip.construct(target=target[0], probability=self.probability)
+
+    def to_matrix(self) -> Iterable[np.ndarray]:
+        K0 = np.sqrt(1 - self.probability) * np.array([[1.0, 0.0], [0.0, 1.0]], dtype=complex)
```
speller26, line 39: For identites, use `np.eye(2)`

src/braket/circuits/circuit.py
licedric: I think a more common use case would be that the user wants to match all gates with a given name (e.g. all XY gates). Do you have a suggestion for a clean way to do this (and also validate qubit_count)?

yitinche: Just has a discussion with Cedric. 
(1) As Cedric pointed above, usually users only want to compare the name of gate. So in this case, passing an instance of Gate seems inappropriate. 
(2) The alternative is to pass class itself without instantiation. In this case, is there a clean way to do type checking? I found that `type(Gate.X)` returns  `<class 'abc.ABCMeta'>` , and `Type[Gate.X]` returns `typing.Type[braket.circuits.gates.X]`. Both of them are not good for type checking. 
(3) Another alternative is just use `str` type as input and user just specify the name of the gates. 

Note that in (2) and (3), we can only check the name of the gate, and have no access to qubit_count.

src/braket/circuits/noise.py
yitinche: each ProbabilisticNoise is a bit different in using the probability, I will add the descriptions to each ProbabilisticNoise class

src/braket/circuits/noises.py
```diff
@@ -0,0 +1,252 @@
...
+    3. Register the class with the `Noise` class via `Noise.register_noise()`.
+"""
+
+
+class Bit_Flip(ProbabilityNoise):
+    """Bit flip noise channel."""
+
+    def __init__(self, prob: float):
+        super().__init__(
+            prob=prob, qubit_count=1, ascii_symbols=["NB({:.2g})".format(prob)],
```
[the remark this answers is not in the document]
yitinche: Google Cirq uses the similar convention, but without prefix "N". I think the idea is to avoid super long circuit printout. I will add the display names to the docstring so that users are not confused.

yitinche:
> Could you please add integration tests against the local simulator as well ?

Added integration tests against the local noise simulator

yitinche:
Thanks for the meeting earlier. I will summarize the few issues left I wanted to bring up in the next few comments. Each issue will be one comment. 
1. Ascii symbol of Noise
Ascii symbol of Noise are "NB" (bit flip), "NP" (phase flip), "NA" (amplitude damping), "ND" (depolarizing) and "NK" (kraus).

There is a request that the ascii symbol need to be more descriptive in the circuit printout. But in the meantime, we need it to be concise otherwise the circuit printout would be too long.  So I think we can use Google's convention but add a prefix "N", which make them slightly more descriptive but are still concise: "NBF" (bit flip), "NPF" (phase flip), "NAD" (amplitude damping), "ND" (depolarizing) and "NK" (kraus). 

Let me know what you think.

src/braket/circuits/noises.py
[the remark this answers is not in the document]
yitinche: @kshitijc I think we also want it to be concise as well. Otherwise the circuit printout would get very long easily. We can use Google's convention but add a prefix "N", which make them slightly more descriptive but are still concise: "NBF" (bit flip), "NPF" (phase flip), "NAD" (amplitude damping), "ND" (depolarizing) and "NK" (kraus). Let me know what you think.

src/braket/circuits/circuit.py
yitinche: As a follow-up, I will use target_gates.__name__ for comparing the target, and issubclass(target_gates, Gate) for type checking.

yitinche: @speller26 Do you know the standard format for the type hint of a class (not an instance)? For example, how do we make it clear that the input should be Gate.X, not Gate.X()?

licedric: I think I didn't understand a different question you asked earlier and accidentally answered this one instead. Check out https://pydantic-docs.helpmanual.io/usage/types/#type
The type hint is probably Type[Gate].

yitinche: okay, thanks! I saw this too but I wasn't sure this is a standard thing or just for pydantic tools.

src/braket/circuits/noise_helpers.py
```diff
@@ -0,0 +1,135 @@
...
+) -> Circuit:
+    """Insert noise to the given time, qubits and gates.
+
+    Args:
+        circuit (Circuit): A ciruit where `noise` is added to.
+        noise (Noise): A `Noise` class object to be added to the circuit.
+        target_gates (Iterable[str]): List of name of gates which `noise` is added to.
+        target_qubits (QubitSet): Index or indices of qubits which `noise` is added to.
+        target_times (Iterable[int]): List of time which `noise` is added to.
+        insert_strategy (str): Rule of how `target_qubit` is used.
```
kshitijc, line 112: No longer used

src/braket/circuits/noises.py
```diff
@@ -0,0 +1,387 @@
...
+       will be added into the `Circuit` class. This method is the default way clients add
+       this noise to a circuit.
+    3. Register the class with the `Noise` class via `Noise.register_noise()`.
+"""
+
+
+class BitFlip(ProbabilisticNoise):
+    """Bit flip noise channel which transforms a density matrix D according to:
+
+    .. math:: D \\Rightarrow (1-p) D + p X D X^{\\dagger}
```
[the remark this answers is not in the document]
yitinche: This is actually correct without sqrt. There is one sqrt(p) associates with each X and there are two X, therefore no sqrt at the outside.

src/braket/circuits/circuit.py
```diff
@@ -438,6 +442,115 @@ def add_circuit(
...
+                isinstance(target_gates, Iterable) and all(isinstance(s, str) for s in target_gates)
+            ):
+                raise TypeError(f"all elements in {target_gates} must be str")
+        if target_times is not None:
+            if not (
+                isinstance(target_times, Iterable) and all(isinstance(t, int) for t in target_times)
+            ):
+                raise TypeError("target_times must be int or Iterable[int]")
+
+        return noise_helpers._add_noise(self, noise, target_gates, target_qubits, target_times)
```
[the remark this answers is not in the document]
yitinche: It seems a bit weird, but moving codes in noise_helpers.py to circuit.py does not sound like a good idea neither.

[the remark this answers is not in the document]
yitinche: Or do you mean I should do `from the noise_helpers import _add_noise` in the beginning of this file, and only do `_add_noise(....)` in this function?

src/braket/circuits/noise_helpers.py
```diff
@@ -0,0 +1,135 @@
...
+        target_qubits = circuit.qubits
+
+    if noise.qubit_count == 1:
+        noise_instructions = [Instruction(noise, qubit) for qubit in target_qubits]
+    else:
+        noise_instructions = [Instruction(noise, target_qubits)]
+
+    new_moments = Moments()
+    time_slices = circuit.moments.time_slices()
+    for time in range(circuit.depth):
```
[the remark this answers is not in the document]
yitinche: Fixed now. The output is now:
```
T  : |        0        |        1        |
                                          
q0 : -X--------NBF(0.2)-C--------NBF(0.2)-
                        |                 
q1 : -NBF(0.2)----------X--------NBF(0.2)-
                                          
q2 : -Y--------NBF(0.2)-NBF(0.2)----------

T  : |        0        |        1        |
```

src/braket/circuits/noise_helpers.py
```diff
@@ -0,0 +1,131 @@
...
+from braket.circuits.instruction import Instruction
+from braket.circuits.moments import Moments
+from braket.circuits.noise import Noise
+from braket.circuits.qubit_set import QubitSet, QubitSetInput
+
+if TYPE_CHECKING:  # pragma: no cover
+    from braket.circuits.circuit import Circuit
+
+
+def _add_noise(
```
speller26, line 15: Helper methods should be public

src/braket/circuits/circuit.py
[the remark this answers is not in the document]
speller26: I think it's fine to just make it a public method in `noise_helpers`

src/braket/circuits/noises.py
```diff
@@ -0,0 +1,391 @@
...
+    1. Implement the class and extend `Noise`
+    2. Add a method with the `@circuit.subroutine(register=True)` decorator. Method name
+       will be added into the `Circuit` class. This method is the default way clients add
+       this noise to a circuit.
+    3. Register the class with the `Noise` class via `Noise.register_noise()`.
+"""
+
+
+class BitFlip(ProbabilisticNoise):
+    """Bit flip noise channel which transforms a density matrix D according to:
```
speller26, line 26: nit: Use the conventional \rho instead of D for density matrix?

src/braket/circuits/circuit.py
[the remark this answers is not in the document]
speller26: I think it's fine to just make it a public method in `noise_helpers`

src/braket/circuits/noises.py
```diff
@@ -0,0 +1,391 @@
...
+        X  =  [0.  1.]
+              [1.  0.]
+
+        Y  =  [0.  -i ]
+              [i    0.]
+
+        Z  =  [1.   0.]
+              [0.  -1.]
+
+        p = probability
```
speller26, line 158: Might as well use LaTeX notation here too, and for other docstrings

src/braket/circuits/noise_helpers.py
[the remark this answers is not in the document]
yitinche: The test for this case (not the exact same circuit, but a similar case) was added.

speller26:
Completed in #212

speller26:
Completed in #212

PR #151, opened 2020-09-01
fix: Call async_result() before calling asyncio.get_event_loop()

licedric:
*Issue #, if available:*

*Description of changes:*
When calling task.result(), task.async_result() was called in the same line and then passed as a parameter. This fix moves the call to async_result() to a preceding line.

Without this fix, calls to task.result() will fail if called from worker threads. The reason is that before this fix, get_event_loop() is called before the event loop is created in async_result(), in case the event loop doesn't already exist. Separating the call to async_result() ensures that the event loop is always created first if necessary.

*Testing done:*
Added integration test that fails before the change and succeeds afterwards.

## Merge Checklist

#### General

#### Tests

### DIFF
--- a/src/braket/aws/aws_quantum_task.py
+++ b/src/braket/aws/aws_quantum_task.py
@@ -236,7 +236,8 @@ class AwsQuantumTask(QuantumTask):
         Consecutive calls to this method return a cached result from the preceding request.
         """
         try:
-            return asyncio.get_event_loop().run_until_complete(self.async_result())
+            async_result = self.async_result()
+            return asyncio.get_event_loop().run_until_complete(async_result)
         except asyncio.CancelledError:
             # Future was cancelled, return whatever is in self._result if anything
             self._logger.warning("Task future was cancelled")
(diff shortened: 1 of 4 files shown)


src/braket/aws/aws_quantum_task.py
```diff
@@ -236,7 +236,8 @@ def result(self) -> Union[GateModelQuantumTaskResult, AnnealingQuantumTaskResult
         Consecutive calls to this method return a cached result from the preceding request.
         """
         try:
-            return asyncio.get_event_loop().run_until_complete(self.async_result())
+            async_result = self.async_result()
```
avawang1, line 239: For learning purposes, what do you mean by "proper setup" in the PR description, what was the error message?

licedric: I got the error message
RuntimeError: There is no current event loop in thread 'ThreadPoolExecutor-0_0'
when calling task.result() from a worker thread.

If there is no existing event loop, then asyncio.get_event_loop() only creates a new one if the OS thread is main (documentation: https://docs.python.org/3/library/asyncio-eventloop.html?highlight=asyncio%20set_event_loop#asyncio.get_event_loop). In the code for async_result() we create a new event loop anyway, but we need to separate that into a separate line to make sure that gets called first.

Updated PR description for clarity.

PR #173, opened 2020-11-19
feature: Add support for batch execution

speller26:
*Issue #, if available:*

*Description of changes:*

Added methods to support running multiple tasks in parallel.

*Testing done:*

## Merge Checklist

#### General

#### Tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,7 +82,7 @@ The code sample imports the Amazon Braket framework, then defines the device to 
 
 ### Running multiple tasks at once
 
-Many quantum algorithms need to run multiple independent circuits, and running the circuits in parallel is much faster than running them one at a time. The following example shows how to run a batch of tasks:
+Many quantum algorithms need to run multiple independent circuits, and submitting the circuits in parallel can be faster than submitting them one at a time. In particular, parallel task processing provides a significant speed up when using simulator devices. The following example shows how to run a batch of tasks on SV1:
 
 ```python
 circuits = [bell for _ in range(5)]
--- a/README.md
+++ b/README.md
@@ -80,6 +80,16 @@ print(task.result().measurement_counts)
 
 The code sample imports the Amazon Braket framework, then defines the device to use (the SV1 AWS simulator). The `s3_folder` statement defines the Amazon S3 bucket for the task result and the folder in the bucket to store the task result. This folder is created when you run the task. It then creates a Bell Pair circuit, executes the circuit on the simulator and prints the results of the job. This example can be found in `../examples/bell.py`.
 
+### Running multiple tasks at once
+
+Many quantum algorithms need to run multiple independent circuits, and running the circuits in parallel is much faster than running them one at a time. The following example shows how to run a batch of tasks:
+
+```python
+circuits = [bell for _ in range(5)]
+batch = device.run_batch(circuits, s3_folder, shots=100)
+print(batch.results()[0].measurement_counts)  # The result of the first task in the batch
+```
+
 ### Available Simulators
 Amazon Braket provides access to two simulators: a fully managed statevector simulator, SV1, and a local simulator that is part of the Amazon Braket SDK.
 
--- a/README.md
+++ b/README.md
@@ -163,7 +163,7 @@ pip install -e "amazon-braket-sdk-python[test]"
 tox -e unit-tests
 ```
 
-You can also pass in various pytest arguments `tox -e integ-tests -- your-arguments` to run selected tests. For more information, please see [pytest usage](https://docs.pytest.org/en/stable/usage.html).
+You can also pass in various pytest arguments `tox -e unit-tests -- your-arguments` to run selected tests. For more information, please see [pytest usage](https://docs.pytest.org/en/stable/usage.html).
 
 
 To run linters and doc generators and unit tests
--- a/src/braket/aws/aws_device.py
+++ b/src/braket/aws/aws_device.py
@@ -18,6 +18,7 @@ from typing import List, Optional, Set, Union
 
 import boto3
 from boltons.dictutils import FrozenDict
+from botocore.config import Config
 from networkx import Graph, complete_graph, from_edgelist
 
 from braket.annealing.problem import Problem
@@ -155,8 +156,9 @@ class AwsDevice(Device):
         task_specifications: List[Union[Circuit, Problem]],
         s3_destination_folder: AwsSession.S3DestinationFolder,
         shots: Optional[int] = None,
-        max_parallel: Optional[int] = None,
-        poll_interval_seconds: Optional[int] = DEFAULT_RESULTS_POLL_INTERVAL,
+        max_parallel: int = AwsQuantumTaskBatch.MAX_PARALLEL_DEFAULT,
+        max_connections: int = AwsQuantumTaskBatch.MAX_CONNECTIONS_DEFAULT,
+        poll_interval_seconds: int = DEFAULT_RESULTS_POLL_INTERVAL,
         *aws_quantum_task_args,
         **aws_quantum_task_kwargs,
     ) -> AwsQuantumTaskBatch:
@@ -171,8 +173,10 @@ class AwsDevice(Device):
             max_parallel (int): The maximum number of tasks to run on AWS in parallel.
                 Batch creation will fail if this value is greater than the maximum allowed
                 concurrent tasks on the device. Default: 10
-            poll_interval_seconds (int): The polling interval for results and
-                and, in seconds. Default: 1 second.
+            max_connections (int): The maximum number of connections in the Boto3 connection pool.
+                Also the maximum number of thread pool workers for the batch. Default: 100
+            poll_interval_seconds (int): The polling interval for results in seconds.
+                Default: 1 second.
             *aws_quantum_task_args: Variable length positional arguments for
                 `braket.aws.aws_quantum_task.AwsQuantumTask.create()`.
             **aws_quantum_task_kwargs: Variable length keyword arguments for
@@ -185,12 +189,13 @@ class AwsDevice(Device):
             `braket.aws.aws_quantum_task_batch.AwsQuantumTaskBatch`
         """
         return AwsQuantumTaskBatch(
-            self._aws_session,
+            AwsDevice._copy_aws_session(self._aws_session, max_connections=max_connections),
             self._arn,
             task_specifications,
             s3_destination_folder,
             shots if shots is not None else self._default_shots,
             max_parallel=max_parallel,
+            max_workers=max_connections,
             poll_interval_seconds=poll_interval_seconds,
             *aws_quantum_task_args,
             **aws_quantum_task_kwargs,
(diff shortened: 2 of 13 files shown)


src/braket/aws/aws_quantum_task_batch.py
```diff
@@ -0,0 +1,231 @@
...
+from typing import List, Set, Union
+
+from braket.annealing import Problem
+from braket.aws.aws_quantum_task import AwsQuantumTask
+from braket.aws.aws_session import AwsSession
+from braket.circuits import Circuit
+
+
+class AwsQuantumTaskBatch:
+    """Executes and a batch of quantum tasks in parallel.
```
licedric, line 27: Remove "and", unless you were planning to add something else.

speller26: Dangit

src/braket/aws/aws_quantum_task_batch.py
```diff
@@ -0,0 +1,231 @@
...
+                for index 0 and key for index 1, that specifies the Amazon S3 bucket and folder
+                to store task results in.
+            shots (int): The number of times to run the task on the device. If the device is a
+                simulator, this implies the state is sampled N times, where N = `shots`.
+                `shots=0` is only available on simulators and means that the simulator
+                will compute the exact results based on the task specification.
+            max_parallel (int): The maximum number of tasks to run on AWS in parallel.
+                Batch creation will fail if this value is greater than the maximum allowed
+                concurrent tasks on the device. Default: 10
+            poll_interval_seconds (float): The polling interval for result(), default is 0.25
```
licedric, line 72: Nit: I'd remove the description for the default. Otherwise if we decide to update the default later we'd have to track down every comment we mentioned it in.

src/braket/aws/aws_quantum_task_batch.py
```diff
@@ -0,0 +1,231 @@
...
+        task_specification,
+        s3_destination_folder,
+        shots,
+        poll_interval_seconds,
+        max_parallel,
+        *args,
+        **kwargs,
+    ):
+        while len(executing) >= max_parallel:
+            time.sleep(poll_interval_seconds)
```
licedric, line 145: I'd shorten this. We aren't actually polling anything so 0.25s seems too long.

test/integ_tests/gate_model_device_testing_utils.py
```diff
@@ -443,3 +441,15 @@ def run_circuit(circuit):
         assert np.allclose(result.measurement_probabilities["00"], 0.5, **tol)
         assert np.allclose(result.measurement_probabilities["11"], 0.5, **tol)
         assert len(result.measurements) == shots
+
+
+def batch_bell_pair_testing(device: AwsDevice, run_kwargs: Dict[str, Any]):
```
licedric, line 446: Can we add an extra assert in here testing that we can still call task.result() on individual tasks after batch execute is done?

test/unit_tests/braket/aws/test_aws_quantum_task_batch.py
```diff
@@ -0,0 +1,81 @@
...
+from common_test_utils import MockS3
+
+from braket.aws import AwsQuantumTaskBatch, AwsSession
+from braket.circuits import Circuit
+from braket.tasks import GateModelQuantumTaskResult
+
+S3_TARGET = AwsSession.S3DestinationFolder("foo", "bar")
+
+
+@patch("braket.aws.aws_quantum_task.AwsQuantumTask.create")
```
licedric, line 28: Can we have a test for larger batch sizes (e.g. 50), for when we actually have to wait for tasks to complete?

src/braket/aws/aws_quantum_task_batch.py
```diff
@@ -0,0 +1,231 @@
...
+
+    Using this class can yield vast speedups over executing tasks sequentially,
+    and is particularly useful for computations that can be parallelized,
+    such as calculating quantum gradients or statistics of terms in a Hamiltonian.
+
+    Note: there is no benefit to using this method with QPUs outside of their execution windows,
+    since results will not be available until the window opens.
+    """
+
+    MAX_WORKERS = 100
```
licedric, line 37: Should we try to feed this into the max_pool_size for the configs for the s3 and braket client as well?

test/integ_tests/gate_model_device_testing_utils.py
speller26: Since the tasks are Mock objects, this wouldn't be particularly useful as the unit test; it belongs more in the integ test

test/unit_tests/braket/aws/test_aws_quantum_task_batch.py
speller26: This is actually already done in the batch integ test, where `max_parallel` is dropped to 5 and there are 15 tasks in the batch

src/braket/aws/aws_quantum_task_batch.py
lundql: Customers always want to know the default.

licedric: The default is `AwsQuantumTask.DEFAULT_RESULTS_POLL_INTERVAL`, as can be seen from the signature. I'd rather mention that as the default instead of 0.25 seconds.

lundql:
Can one of you give me a brief explanation--from a customer perspective--use case and benefits--when would I want to use this feature? Is it a way to launch tasks in parallel? Does it make tasks run faster? Is it something the customer has to set up using these scripts or notebooks? Is it available on all devices for Braket, or which ones? Thank you very much, I am sorry to be so uninformed, I am just getting up to speed on all of your team's endeavors, and turning right around to try to explain them for our customers. You can send email if you don't want to reply here. Many thanks!

speller26:
Yep, it's a way to launch tasks in parallel. For background, let's say you want to calculate something that requires 10 tasks, but the circuits in the tasks are independent of one another. You shouldn't have to wait for one task to be created before you move on the the next on. This is available on all devices, but most useful for the simulator, which can also run things concurrently.

lundql:
Ok, so this feature is independent of device?  It is actually a NEW feature of Braket we are releasing with others on 12/7?

src/braket/aws/aws_device.py
```diff
@@ -137,23 +139,68 @@ def run(
...
         )
 
+    def run_batch(
+        self,
+        task_specifications: List[Union[Circuit, Problem]],
+        s3_destination_folder: AwsSession.S3DestinationFolder,
+        shots: Optional[int] = None,
+        max_parallel: int = AwsQuantumTaskBatch.MAX_PARALLEL_DEFAULT,
+        max_connections: int = AwsQuantumTaskBatch.MAX_CONNECTIONS_DEFAULT,
+        poll_interval_seconds: int = DEFAULT_RESULTS_POLL_INTERVAL,
```
licedric, line 161: Should we add something for poll_timeout_seconds too?

speller26: Those are in the `*args` and `**kwargs`; `poll_interval_seconds` is only here because it's actually used by AwsQuantumTaskBatch.

src/braket/aws/aws_quantum_task_batch.py
```diff
@@ -0,0 +1,236 @@
...
+    def __init__(
+        self,
+        aws_session: AwsSession,
+        device_arn: str,
+        task_specifications: List[Union[Circuit, Problem]],
+        s3_destination_folder: AwsSession.S3DestinationFolder,
+        shots: int,
+        max_parallel: int = MAX_PARALLEL_DEFAULT,
+        max_workers: int = MAX_CONNECTIONS_DEFAULT,
+        poll_interval_seconds: float = AwsQuantumTask.DEFAULT_RESULTS_POLL_INTERVAL,
```
licedric, line 53: As mentioned in aws_session.py, consider adding poll_timeout_seconds to match the behavior for AwsQuantumTask.

test/unit_tests/braket/aws/test_aws_device.py
```diff
@@ -305,9 +319,9 @@ def test_aws_session_in_another_qpu_region(
     aws_session.get_device.return_value = MOCK_GATE_MODEL_QPU_1
 
     creds = Mock()
-    creds.access_key = "access key"
-    creds.secret_key = "secret key"
-    creds.token = "token"
+    creds.access_key = "foo"
```
licedric, line 322: Wondering what's up with this change?

licedric:
Looks good to me, thanks!

licedric (approved):

src/braket/aws/aws_quantum_task.py
```diff
@@ -331,8 +340,8 @@ async def _wait_for_completion(
                     f" now."
                 )
                 continue
-            current_metadata = self.metadata()
-            task_status = current_metadata["status"]
+            task_status = self._status()
```
licedric, line 343: One thing to consider is checking the cached result (if self.metadata() already exists) and checking whether it's a terminal state; if it is, then we save a call to GetQuantumTask.

test/unit_tests/braket/aws/test_aws_device.py
speller26: I added creds to the fixture too, and just wanted to distinguish the ones here. Even though the fixture creds aren't used in this test, in case it gets modified down the line, I don't want a pass simply because the creds are coicidentally the same.

src/braket/aws/aws_quantum_task.py
speller26: Good idea

src/braket/aws/aws_quantum_task_batch.py
```diff
@@ -98,15 +102,16 @@ def _execute(
...
         max_parallel,
+        max_workers,
         poll_interval_seconds,
         *args,
         **kwargs,
     ):
-        # These two lists track the number of remaining and currently executing tasks
-        # Since appending and popping are atomic, they serve as locks
+        # Tracks the number of remaining and currently executing tasks
+        # Since appending and popping are atomic, it serve as a lock too
```
lundql, line 114: # Because appending and popping are atomic, it serves as a lock, too

(Note to writer: It had a typo, it needed a coma, and you really meant "because" as in causality, not "since" as in a period of time has elapsed)

test/unit_tests/braket/aws/test_aws_device.py
lundql: Believe it or not, AWS Style advises not to use "foo" and those other tokens, due to their unsavory origins in military profanity.  I realize they are almost universal in the industry. If you can possibly think of others, that would be better, because customers will see this example, correct?

lundql (changes requested):
Helpful example in the README, thank you!

I had two other comments just based on AWS style and grammar.

src/braket/aws/aws_quantum_task_batch.py
lundql: needed a comma, not a coma, hahah!

README.md
```diff
@@ -80,6 +80,16 @@ print(task.result().measurement_counts)
 
 The code sample imports the Amazon Braket framework, then defines the device to use (the SV1 AWS simulator). The `s3_folder` statement defines the Amazon S3 bucket for the task result and the folder in the bucket to store the task result. This folder is created when you run the task. It then creates a Bell Pair circuit, executes the circuit on the simulator and prints the results of the job. This example can be found in `../examples/bell.py`.
 
+### Running multiple tasks at once
+
+Many quantum algorithms need to run multiple independent circuits, and running the circuits in parallel is much faster than running them one at a time. The following example shows how to run a batch of tasks:
```
emkessler, line 85: Need to distinguish between sims and QPUs:

Many quantum algorithms need to run multiple independent circuits. Submitting the circuits in parallel can be faster than submitting them one at a time. In particular, if you use simulator devices, parallel processing of your tasks provide a significant speed up.

lundql (approved):
Nice work.

PR #174, opened 2020-11-23
Correctly cache status

speller26:
*Issue #, if available:*

*Description of changes:*

*Testing done:*

## Merge Checklist

#### General

#### Tests

### DIFF
--- a/src/braket/aws/aws_quantum_task.py
+++ b/src/braket/aws/aws_quantum_task.py
@@ -219,15 +219,16 @@ class AwsQuantumTask(QuantumTask):
 
         Args:
             use_cached_value (bool, optional): If `True`, uses the value most recently retrieved
-                from the Amazon Braket `GetQuantumTask` operation. If `False`, calls the
-                `GetQuantumTask` operation  to retrieve metadata, which also updates the cached
-                value. Default = `False`.
+                from the Amazon Braket `GetQuantumTask` operation, if it exists; if not,
+                `GetQuantumTask` will be called to retrieve the metadata. If `False`, always calls
+                `GetQuantumTask`, which also updates the cached value. Default: `False`.
         Returns:
             Dict[str, Any]: The response from the Amazon Braket `GetQuantumTask` operation.
             If `use_cached_value` is `True`, Amazon Braket is not called and the most recently
-            retrieved value is used.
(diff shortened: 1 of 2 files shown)


src/braket/aws/aws_quantum_task.py
```diff
@@ -270,7 +275,7 @@ def result(self) -> Union[GateModelQuantumTaskResult, AnnealingQuantumTaskResult
             if the task completed successfully; returns None if the task did not complete
             successfully or the future timed out.
         """
-        if self._result or self._status(True) in self.NO_RESULT_TERMINAL_STATES:
+        if self._result or self._status in self.NO_RESULT_TERMINAL_STATES:
```
licedric, line 278: What's up with this change? I couldn't find self._status anywhere.

src/braket/aws/aws_quantum_task.py
```diff
@@ -255,6 +256,10 @@ def _status(self, use_cached_value=False):
             self._logger.warning(f"Task is in terminal state {status} and no result is available")
         return status
 
+    def _update_status_if_nonterminal(self):
```
licedric, line 259: This is still not ideal: consider the case where the metadata is empty and the status is "RUNNING".
Then first self._status(True) is called, filling the metadata and returning the "RUNNING" status.
Since it's not a terminal state, self._status(False) will be called again. This uses two GetQuantumTask calls.

src/braket/aws/aws_quantum_task.py
speller26: My bad, I tried introducing a `_status` property and removed it

src/braket/aws/aws_quantum_task.py
speller26: However, there will be at most one extra call to GetQuantumTask, and since the status isn't terminal, I don't think it hurts to make the extra call; it's going to have to poll again regardless. Easy enough to get rid of it though.

licedric: In the scenario I described it's two successive calls to GetQuantumTask, so the second call will likely return the same result and be wasted. In any case, thanks for the fix.

licedric (approved):

PR #189, opened 2020-12-22
fix: Get regions for QPUs instead of providers

speller26:
*Description of changes:*

Overhauls get_devices logic to filter regions based on ARNs instead of providers. This future-proofs the SDK for QPUs added to new regions.

By getting rid of the region lists, this also fixes a bug where the wrong devices would be returned for regions other than us-west-2, which was the first region in the list.

*Testing done:*

[build_files.tar.gz](https://github.com/aws/amazon-braket-sdk-python/files/5727106/build_files.tar.gz)

## Merge Checklist

#### General

#### Tests

### DIFF
--- a/src/braket/aws/aws_device.py
+++ b/src/braket/aws/aws_device.py
@@ -14,7 +14,7 @@
 from __future__ import annotations
 
 from enum import Enum
-from typing import List, Optional, Set, Union
+from typing import FrozenSet, List, Optional, Set, Union
 
 import boto3
 from boltons.dictutils import FrozenDict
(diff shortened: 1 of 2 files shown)


src/braket/aws/aws_device.py
```diff
@@ -320,33 +327,31 @@ def _aws_session_for_qpu(device_arn: str, aws_session: Optional[AwsSession]) ->
...
+        region: Optional[str] = None,
         max_connections: Optional[int] = None,
     ) -> AwsSession:
         config = Config(max_pool_connections=max_connections) if max_connections else None
         if aws_session:
             session_region = aws_session.boto_session.region_name
-            new_regions = regions or [session_region]
-            if session_region in new_regions and not config:
+            new_region = region or session_region
+            if session_region == new_region and not config:
```
licedric, line 342: I'm actually wondering if this is desired behavior here - do we want to enable refreshing credentials for a new AwsSession? (Probably not in this PR, but worth thinking about later.)

src/braket/aws/aws_device.py
```diff
@@ -397,60 +402,55 @@ def get_devices(
...
-        )
-        if provider_names:
-            provider_region_set = set()
-            for provider in provider_names:
-                for key in AwsDevice.DEVICE_REGIONS:
-                    if key in provider.lower():
-                        provider_region_set.add(AwsDevice.DEVICE_REGIONS[key][0])
-                        break
-            device_regions_set = device_regions_set.intersection(provider_region_set)
+        device_regions_set = set(AwsDevice.REGIONS)
```
licedric, line 443:
```
if types == {AwsDeviceType.SIMULATOR}:
    device_regions_set = {current_region}
else:
    device_regions_set = set(AwsDevice.REGIONS)
```

This has the added benefit of allowing bad region names in device_regions_set for simulator, which means it'll throw an error (as it should).

src/braket/aws/aws_device.py
speller26: That's something for a future PR; all this change does is move from lists of regions to just individual regions.

licedric (approved):

src/braket/aws/aws_device.py
```diff
@@ -397,60 +402,55 @@ def get_devices(
...
+        if types == {AwsDeviceType.SIMULATOR}:
+            device_regions_set &= {current_region}
```
speller26, line 445:
```suggestion
        device_regions_set = (
            {current_region} if types == {AwsDeviceType.SIMULATOR} else set(AwsDevice.REGIONS)
        )
```

PR #195, opened 2021-01-07
change: update result_types to use hashing

ajberdy:
use a dict to hold the result types to speed up checking if a result type is in _result_types. result_types property still interfaces as a list. Python 3.7+ officially guarantees order for dicts, so result_types is expected to be in order

also edit gate_model_quantum_task_result to add a dict mapping result type to index, as well as ensuring consistency regarding result type in the implementation

*Issue #, if available:*

*Description of changes:*

*Testing done:*

## Merge Checklist

#### General

#### Tests

### DIFF
--- a/src/braket/circuits/result_type.py
+++ b/src/braket/circuits/result_type.py
@@ -125,6 +125,9 @@ class ResultType:
     def __repr__(self) -> str:
         return f"{self.name}()"
 
+    def __hash__(self) -> int:
+        return hash(repr(self))
+
 
 class ObservableResultType(ResultType):
     """
@@ -197,3 +200,6 @@ class ObservableResultType(ResultType):
 
     def __copy__(self) -> ObservableResultType:
         return type(self)(observable=self.observable, target=self.target)
+
+    def __hash__(self) -> int:
+        return super().__hash__()
--- a/src/braket/circuits/result_types.py
+++ b/src/braket/circuits/result_types.py
@@ -65,6 +65,11 @@ class StateVector(ResultType):
     def __copy__(self) -> StateVector:
         return type(self)()
 
+    # must redefine __hash__ since __eq__ is overwritten
+    # https://docs.python.org/3/reference/datamodel.html#object.__hash__
+    def __hash__(self) -> int:
+        return super().__hash__()
+
 
 ResultType.register_result_type(StateVector)
 
@@ -135,6 +140,9 @@ class Amplitude(ResultType):
     def __copy__(self):
         return type(self)(state=self.state)
 
+    def __hash__(self) -> int:
+        return super().__hash__()
+
 
 ResultType.register_result_type(Amplitude)
 
@@ -174,7 +182,8 @@ class Probability(ResultType):
 
     def to_ir(self) -> ir.Probability:
         if self.target:
-            return ir.Probability.construct(targets=list(self.target))
+            # convert qubits to int as required by the ir type
+            return ir.Probability.construct(targets=[int(qubit) for qubit in self.target])
         else:
             return ir.Probability.construct()
 
@@ -206,6 +215,9 @@ class Probability(ResultType):
 
     def __copy__(self) -> Probability:
         return type(self)(target=self.target)
+
+    def __hash__(self) -> int:
+        return super().__hash__()
 
 
 ResultType.register_result_type(Probability)
@@ -251,7 +263,7 @@ class Expectation(ObservableResultType):
     def to_ir(self) -> ir.Expectation:
         if self.target:
             return ir.Expectation.construct(
-                observable=self.observable.to_ir(), targets=list(self.target)
+                observable=self.observable.to_ir(), targets=[int(qubit) for qubit in self.target]
             )
         else:
             return ir.Expectation.construct(observable=self.observable.to_ir())
@@ -318,7 +330,7 @@ class Sample(ObservableResultType):
     def to_ir(self) -> ir.Sample:
         if self.target:
             return ir.Sample.construct(
-                observable=self.observable.to_ir(), targets=list(self.target)
+                observable=self.observable.to_ir(), targets=[int(qubit) for qubit in self.target]
             )
         else:
             return ir.Sample.construct(observable=self.observable.to_ir())
@@ -386,7 +398,7 @@ class Variance(ObservableResultType):
     def to_ir(self) -> ir.Variance:
         if self.target:
             return ir.Variance.construct(
-                observable=self.observable.to_ir(), targets=list(self.target)
+                observable=self.observable.to_ir(), targets=[int(qubit) for qubit in self.target]
             )
         else:
             return ir.Variance.construct(observable=self.observable.to_ir())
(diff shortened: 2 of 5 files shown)


src/braket/tasks/gate_model_quantum_task_result.py
```diff
@@ -31,6 +32,10 @@
 T = TypeVar("T")
 
 
+def result_type_hash(rt_type):
+    return repr(dict(sorted(dict(rt_type).items(), key=lambda x: x[0])))
```
speller26, line 36: Can you add unit tests asserting that result types are hashed properly, particularly arbitrary Hermitian observables? Also, this can be private, and moved to the bottom of the class.

src/braket/tasks/gate_model_quantum_task_result.py
```diff
@@ -332,6 +349,7 @@ def _calculate_result_types(
                 value = GateModelQuantumTaskResult._probability_from_measurements(
                     measurements, measured_qubits, targets
                 )
+                casted_result_type = Probability(targets=targets)
```
speller26, line 352: What is this casting for?

src/braket/tasks/gate_model_quantum_task_result.py
```diff
@@ -79,8 +84,8 @@ class GateModelQuantumTaskResult:
 
     task_metadata: TaskMetadata
     additional_metadata: AdditionalMetadata
-    result_types: List[Dict[str, str]]
-    values: List[Any]
+    result_types: List[ResultTypeValue] = None
+    values: List[Any] = None
```
speller26, line 88: Why are there default values now? It doesn't look like the constructor is being called differently anywhere.

src/braket/circuits/result_type.py
```diff
@@ -197,3 +200,6 @@ def __repr__(self) -> str:
 
     def __copy__(self) -> ObservableResultType:
         return type(self)(observable=self.observable, target=self.target)
+
+    def __hash__(self) -> int:
+        return super().__hash__()
```
speller26, line 205: Would this already inherit the superclass method?

src/braket/circuits/result_types.py
```diff
@@ -174,7 +182,8 @@ def target(self, target: QubitSetInput) -> None:
 
     def to_ir(self) -> ir.Probability:
         if self.target:
-            return ir.Probability.construct(targets=list(self.target))
+            # convert qubits to int as required by the ir type
+            return ir.Probability.construct(targets=[int(qubit) for qubit in self.target])
```
speller26, line 186: This breaks with `Qubit`? Surprising consider `Qubit` implements `int`

src/braket/tasks/gate_model_quantum_task_result.py
ajberdy: result_type is a dict and ResultTypeValue.type expects an ir Result

src/braket/tasks/gate_model_quantum_task_result.py
ajberdy: Since the type annotation states that these values are lists and doesn't provide a default, you would expect them to always be of type list. That isn't true, since a supported behavior is passing None for these values-- if these values are in fact optional, it's good form to indicate that by providing a default value of None, rather than explicitly passing None when a list is expected.

Alternatively, we can stop supporting None and insist on passing an empty list to initialize with no results, but that could break existing usages

src/braket/circuits/result_type.py
ajberdy: __hash__ is not inherited if __eq__ is overwritten. Dropped a link in the code in case anyone has the same question in the future: https://docs.python.org/3/reference/datamodel.html#object.__hash__

src/braket/circuits/result_types.py
ajberdy: Indeed. Since the result_type hash uses __repr__, '1' != 'Qubit(1)'. It's possible to move the casting into the hash function if you think that would be cleaner, but since ir typically has raw ints, I feel it could make sense for the to_ir function to cast to int

src/braket/tasks/gate_model_quantum_task_result.py
ajberdy: On it

test/unit_tests/braket/tasks/test_gate_model_quantum_task_result.py
```diff
@@ -373,3 +373,35 @@ def test_calculate_ir_results_value_error():
     measured_qubits = [0]
     measurements = np.array([[0]])
     GateModelQuantumTaskResult._calculate_result_types(ir_string, measurements, measured_qubits)
+
+
+def test_hash_result_types():
```
speller26, line 378: Nice! Try using `pytest.mark.parametrize`, like what's done [here](https://github.com/aws/amazon-braket-sdk-python/blob/main/test/unit_tests/braket/circuits/test_circuit.py#L734-L760). This way, you can use the same test code and just add more cases.

test/unit_tests/braket/tasks/test_gate_model_quantum_task_result.py
```diff
@@ -373,3 +373,35 @@ def test_calculate_ir_results_value_error():
...
+    expectation_2 = jaqcd.Expectation(observable=["x"], targets=[1, 2])
+    expectation_3 = jaqcd.Expectation(observable=["x"], targets=[2, 1])
+    expectation_4 = jaqcd.Expectation(
+        observable=[[[[0, 0], [1, 0]], [[1, 0], [0, 0]]], [[[1, 0], [1, 0]], [[1, 0], [-1, 0]]]],
+        targets=[1, 2],
+    )
+    expectation_5 = jaqcd.Expectation(
+        observable=[[[[0, 0], [1, 0]], [[1, 0], [0, 0]]], [[[1, 0], [1, 0]], [[1, 0], [-1, 0]]]],
+        targets=[1, 2],
+    )
```
speller26, line 389: Probably being paranoid, but also add a test case for Hermitians with a a bunch of digits after the decimal, just for peace of mind.

src/braket/circuits/result_types.py
speller26: Interesting, what I'm seeing is
```
In [4]: 1 == Qubit(1)                                                                   
Out[4]: True
```

ajberdy: I use the repr function in the hash

```
In: repr(Qubit(1)) == repr(1)
Out: False
```

test/unit_tests/braket/tasks/test_gate_model_quantum_task_result.py
ajberdy: done

test/unit_tests/braket/tasks/test_gate_model_quantum_task_result.py
ajberdy: changed!

speller26 (approved):

PR #196, opened 2021-01-08
add placeholder methods for result types to enable ide compatibility

ajberdy:
*Issue #, if available:*

*Description of changes:*

Previously, ResultType.StateVector() for example would not be recognized as valid by the IDE, since result type constructors as ResultType classmethods are defined dynamically. Adding placeholder classmethods for established result types so the IDE recognizes them as valid.

*Testing done:*

existing tox and manual verification that result types work as previously (no logic was changed)

## Merge Checklist

#### General

#### Tests

### DIFF
--- a/src/braket/circuits/result_type.py
+++ b/src/braket/circuits/result_type.py
@@ -125,6 +125,26 @@ class ResultType:
     def __repr__(self) -> str:
         return f"{self.name}()"
 
+    @classmethod
+    def StateVector(cls):
+        raise NotImplementedError("Result type not registered")
+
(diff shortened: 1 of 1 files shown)


speller26 (commented):
I think we need to be careful about how we do this; my concern is that we're introducing these methods in a different class from where their actual definitions reside. Currently, we only need to define these in one place, but with this, we'll have to remember to add a new method to `ResultType` every time a new result type is defined in the `result_types` module. And if we do this here, we'd want to do it for each of the gates too, and that quickly gets messy.

PR #201, opened 2021-01-16
fix: ensure AngledGate casts its angle argument to float so it can be…

ajberdy:
… serialized to json

*Issue #, if available:*

https://github.com/aws/amazon-braket-sdk-python/issues/200

*Description of changes:*

Cast angle argument to float so it can be serialized to json (previously np.float32 would throw an error)

*Testing done:*

Added unit test to test_angled_gate where angle is instantiated from np.float32

## Merge Checklist

#### General

#### Tests

### DIFF
--- a/test/unit_tests/braket/circuits/test_angled_gate.py
+++ b/test/unit_tests/braket/circuits/test_angled_gate.py
@@ -11,6 +11,8 @@
 # ANY KIND, either express or implied. See the License for the specific
 # language governing permissions and limitations under the License.
 
+import re
+
 import numpy as np
 import pytest
 from pydantic import BaseModel
@@ -50,7 +52,7 @@ def test_angle_setter(angled_gate):
 
 def test_np_float_angle_json():
     angled_gate = AngledGate(angle=np.float32(0.15), qubit_count=1, ascii_symbols=["foo"])
-    assert (
-        BaseModel.construct(target=[0], angle=angled_gate.angle).json()
-        == '{"target": [0], "angle": 0.15000000596046448}'
-    )
+    angled_gate_json = BaseModel.construct(target=[0], angle=angled_gate.angle).json()
+    match = re.match(r'\{"target": \[0], "angle": (\d*\.?\d*)}', angled_gate_json)
+    angle_value = float(match.group(1))
+    assert np.isclose(angle_value, 0.15)
--- a/test/unit_tests/braket/circuits/test_angled_gate.py
+++ b/test/unit_tests/braket/circuits/test_angled_gate.py
@@ -11,7 +11,9 @@
 # ANY KIND, either express or implied. See the License for the specific
 # language governing permissions and limitations under the License.
 
+import numpy as np
 import pytest
+from pydantic import BaseModel
 
 from braket.circuits import AngledGate, Gate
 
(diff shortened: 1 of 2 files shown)


test/unit_tests/braket/circuits/test_angled_gate.py
```diff
@@ -44,3 +46,11 @@ def test_getters():
 @pytest.mark.xfail(raises=AttributeError)
 def test_angle_setter(angled_gate):
     angled_gate.angle = 0.14
+
+
+def test_np_float_angle_json():
+    angled_gate = AngledGate(angle=np.float32(0.15), qubit_count=1, ascii_symbols=["foo"])
+    assert (
+        BaseModel.construct(target=[0], angle=angled_gate.angle).json()
```
speller26, line 54: Instead of using `BaseModel`, prefer using [`BraketSchemaBase.parse_raw_schema`](https://github.com/aws/amazon-braket-schemas-python/blob/main/src/braket/schema_common/schema_base.py#L58):

```python
# Something like this
assert(
    angled_gate == BraketSchemaBase.parse_raw_json(angled_gate.json())
)
```

test/unit_tests/braket/circuits/test_angled_gate.py
```diff
@@ -44,3 +46,11 @@ def test_getters():
...
+        == '{"target": [0], "angle": 0.15000000596046448}'
```
speller26, line 55: Is this approximation guaranteed?

test/unit_tests/braket/circuits/test_angled_gate.py
ajberdy: Hey, the issue here is that AngledGate is an abstract class with no implementation for json().

Any specific AngledGate subclass has a to_ir function along the lines of the following:

```python
def to_ir(self, target: QubitSet):
    return ir.Rx.construct(target=target[0], angle=self.angle)
```

Here, ir.Rx is a subclass of BaseModel,  so instead of BaseModel.construct, I could use any specific ir version of an Angled Gate class, but I didn't want to rely on a subclass for testing here.

speller26: Ugh, that was me getting IR and SDK classes mixed up. Cool.

test/unit_tests/braket/circuits/test_angled_gate.py
speller26: To add on to this, rather than asserting the JSON string, can you convert it back and assert the round-trip value is the same? That way, we don't have to hardcode values, and we test the important piece of functionality.

ajberdy: I believe so. Nevertheless, I replaced the equality check with a regex match and used the np isclose() method to compare the two values.

test/unit_tests/braket/circuits/test_angled_gate.py
```diff
@@ -44,3 +48,11 @@ def test_getters():
...
 def test_angle_setter(angled_gate):
     angled_gate.angle = 0.14
+
+
+def test_np_float_angle_json():
+    angled_gate = AngledGate(angle=np.float32(0.15), qubit_count=1, ascii_symbols=["foo"])
+    angled_gate_json = BaseModel.construct(target=[0], angle=angled_gate.angle).json()
+    match = re.match(r'\{"target": \[0], "angle": (\d*\.?\d*)}', angled_gate_json)
+    angle_value = float(match.group(1))
+    assert np.isclose(angle_value, 0.15)
```
speller26, line 58:
```suggestion
    assert np.isclose(angle_value, angled_gate.angle)
```

speller26 (approved):
LGTM, just one minor comment!

test/unit_tests/braket/circuits/test_angled_gate.py
ajberdy: I like the suggestion. Following this path, the rounding error is already accounted for in the float32 -> float conversion, so we would expect angle_value to match angle_gate.angle exactly, I'll push one more commit removing isclose()

speller26 (approved):

PR #226, opened 2021-03-11
fix: check for task completion before entering async event loop

ajberdy:
*Issue #, if available:*

*Description of changes:*

Add check for completeness in task.result() and directly download the result if available

Will be adding more commits to this pull request for related changes

*Testing done:*

unit/integ

## Merge Checklist

#### General

#### Tests

### DIFF
--- a/src/braket/aws/aws_quantum_task.py
+++ b/src/braket/aws/aws_quantum_task.py
@@ -272,6 +272,8 @@ class AwsQuantumTask(QuantumTask):
         """
         if self._result or self._status(True) in self.NO_RESULT_TERMINAL_STATES:
             return self._result
+        if self._status(True) in self.RESULTS_READY_STATES:
+            return self._download_result()
         try:
             async_result = self.async_result()
             return async_result.get_loop().run_until_complete(async_result)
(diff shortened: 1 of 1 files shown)


speller26 (dismissed):
Interesting that the current unit tests already cover this

src/braket/aws/aws_quantum_task.py
```diff
@@ -272,6 +272,8 @@ def result(self) -> Union[GateModelQuantumTaskResult, AnnealingQuantumTaskResult
         """
         if self._result or self._status(True) in self.NO_RESULT_TERMINAL_STATES:
             return self._result
+        if self._status(True) in self.RESULTS_READY_STATES:
```
licedric, line 275: Note this has one of the issues being fixed in https://github.com/aws/amazon-braket-sdk-python/pull/234/files .

licedric, line 275: This will call get_quantum_task if we've never polled before. Is this intended? 
I was trying to fix this issue two lines above in https://github.com/aws/amazon-braket-sdk-python/pull/226/files

ajberdy: I think that is reasonable, since there is no need for polling if a task is already done. Looks like we need this call to check status.

licedric: The issue with this currently is if we've never polled before (self._metadata is empty), then this will poll immediately to full self._metadata and then possibly proceed to the async event loop, where it could poll immediately again. That would be an extra poll.

ajberdy, line 275:
```suggestion
        if self._metadata and self._status(True) in self.RESULTS_READY_STATES:
```

licedric (approved):

PR #250, opened 2021-06-11
fix: Get qubit count without instantiating op

speller26:
*Issue #, if available:*

*Description of changes:*

Introduced `fixed_qubit_count` so customers don't need to instantiate a gate or noise operation to see its qubit count. Also fixed a bug when checking compatibility of gate and noise qubit counts, where the gate could always be instantiated with no arguments, which obviously fails for parametrized gates.

*Testing done:*

## Merge Checklist

#### General

#### Tests

### DIFF
--- a/src/braket/circuits/gates.py
+++ b/src/braket/circuits/gates.py
@@ -42,6 +42,8 @@ To add a new gate:
 class H(Gate):
     """Hadamard gate."""
 
+    qubit_count = 2
+
     def __init__(self):
         super().__init__(qubit_count=1, ascii_symbols=["H"])
 
--- a/src/braket/circuits/gates.py
+++ b/src/braket/circuits/gates.py
@@ -43,7 +43,7 @@ class H(Gate):
     """Hadamard gate."""
 
     def __init__(self):
-        super().__init__(qubit_count=self.fixed_qubit_count(), ascii_symbols=["H"])
+        super().__init__(qubit_count=None, ascii_symbols=["H"])
 
     def to_ir(self, target: QubitSet):
         return ir.H.construct(target=target[0])
--- a/src/braket/circuits/gates.py
+++ b/src/braket/circuits/gates.py
@@ -42,16 +42,18 @@ To add a new gate:
 class H(Gate):
     """Hadamard gate."""
 
-    qubit_count = 2
-
     def __init__(self):
-        super().__init__(qubit_count=1, ascii_symbols=["H"])
+        super().__init__(qubit_count=self.fixed_qubit_count(), ascii_symbols=["H"])
 
     def to_ir(self, target: QubitSet):
         return ir.H.construct(target=target[0])
 
     def to_matrix(self) -> np.ndarray:
         return 1.0 / np.sqrt(2.0) * np.array([[1.0, 1.0], [1.0, -1.0]], dtype=complex)
+
+    @staticmethod
+    def fixed_qubit_count() -> int:
+        return 1
 
     @staticmethod
     @circuit.subroutine(register=True)
@@ -68,7 +70,7 @@ class H(Gate):
             >>> circ = Circuit().h(0)
             >>> circ = Circuit().h([0, 1, 2])
         """
-        return [Instruction(Gate.H(), target=qubit) for qubit in QubitSet(target)]
+        return [Instruction(H(), target=qubit) for qubit in QubitSet(target)]
 
 
 Gate.register_gate(H)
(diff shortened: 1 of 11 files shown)


src/braket/circuits/gates.py
```diff
@@ -43,14 +43,18 @@ class H(Gate):
     """Hadamard gate."""
 
     def __init__(self):
-        super().__init__(qubit_count=1, ascii_symbols=["H"])
+        super().__init__(qubit_count=self.fixed_qubit_count(), ascii_symbols=["H"])
```
licedric, line 46: Why is this different from all the other gates? is `self.fixed_qubit_count()` or `None` the correct choice here?

speller26: Oops, missed an undo. Should be None, feel free to suggest and commit that change.

licedric: Probably faster just to ask you - what happens when qubit_count is `None`? Does it automatically get redirected to `fixed_qubit_count`?

speller26: So it's defined in `QuantumOperation`: if `fixed_qubit_count` is defined, then the argument `qubit_count` is ignored altogether, hence `None`. I would e like to move the argument to the end and set `None` as default, but that would be breaking.

speller26, line 46:
```suggestion
        super().__init__(qubit_count=None, ascii_symbols=["H"])
```

src/braket/circuits/noise_helpers.py
```diff
@@ -62,10 +62,11 @@ def check_noise_target_gates(noise: Noise, target_gates: Iterable[Type[Gate]]):
 
     if noise.qubit_count > 1:
         for g in target_gates:
-            if g != Gate.Unitary and g().qubit_count != noise.qubit_count:
+            fixed_qubit_count = g.fixed_qubit_count()
+            if g != Gate.Unitary and fixed_qubit_count != noise.qubit_count:
```
licedric, line 66: If Unitary is an input, should we raise an error?

src/braket/circuits/quantum_operator.py
```diff
@@ -35,25 +40,50 @@ def __init__(self, qubit_count: int, ascii_symbols: Sequence[str]):
...
                 `ascii_symbols` length != `qubit_count`
         """
 
-        if qubit_count < 1:
-            raise ValueError(f"qubit_count, {qubit_count}, must be greater than zero")
-        self._qubit_count = qubit_count
+        fixed_qubit_count = self.fixed_qubit_count()
+        self._qubit_count = (
+            qubit_count if fixed_qubit_count is NotImplemented else fixed_qubit_count
+        )
```
kshitijc, line 51: Would it be better to delegate this to the subclasses instead ? That is, each class passes in the `qubit_count=self.fixed_qubit_count()` instead of `qubit_count=None`. This would be much simpler and make it easier to handle other such cases in the future, if any. Trying to handle other scenarios in the QuantumOperator class could get challenging.

speller26: Thing is, I want to enforce `GateClass.fixed_qubit_count() == GateClass(**kwargs).qubit_count`, and this was the only real way to do this in a backwards-compatible way. If each subclass does this work instead,
1. We lose that equality contract; there's no stopping the class and instants qubit counts from being different (I use "stop" loosely, as this is Python we're talking about and you can technically do anything), and
2. We basically leave it to the implementer to pass `fixed_qubit_count()` into the class' own constructor, which isn't immediately obvious

kshitijc: It's a little confusing that we simply ignore the `qubit_count` value passed in, if `fixed_qubit_count` has been implemented. If we wish to enforce that `GateClass.fixed_qubit_count() == GateClass(**kwargs).qubit_count`, then could we explicitly add that check in the constructor and raise an error informing the user that fixed qubit count and qubit count values don't match, but we expect them to ?

speller26: I can add a check to ensure that `qubit_count`, if passed in, is equal to `fixed_qubit_count()`. However, I still want the work required to implement a fixed qubit count gate to be limited to implementing `fixed_qubit_count`, and not passing that number in as well. Ideally, `qubit_count` would be the second parameter, with a default value, but that would be a breaking change for the anyone who creates new operations and uses positional arguments.

kshitijc: Yes, that's true. I was thinking that if we require customers to pass in qubit_count=None, we might as well have them pass in the actual fixed qubit count. It seemed simpler, and anyone referencing the existing implementation should be able to get it right. This is not a blocker though, if you feel it won't work well.

speller26: I think restricting `qubit_count` to equal `fixed_qubit_count` if both are nonnull is sufficient to erase the confusion. Instead of always ignoring the parameter remains optional, but still holds meaning if passed.

kshitijc (approved):

PR #256, opened 2021-07-02
Add circuit loader methods: from_ir, from_repr and from_diagram

kjacky:
*Issue #, if available:*

*Description of changes:* Add three circuit loader methods to the Circuit class:

- `from_ir(program: Program) -> Circuit` - Create Circuit from Program object.
- `from_repr(repr_str: str) -> Circuit` - Create Circuit from a string produced by repr(circuit).
- `from_diagram(diagram_str: str) -> Circuit` - Create Circuit from a circuit diagram string produced by f'{circuit}'.

These methods are added to the Circuit class as a subroutine using `@circuit.subroutine(register=True)`.

*Testing done:*

- `./test/unit_tests/braket/circuits/test_from.py` to test 30 test cases of circuit ascii diagrams covering quantum gates, noise operators and result types. (The ascii diagram files are stored under from/input/test_cases.txt.)
- `./test/unit_tests/braket/circuits/test_from_matrics.py` to test the unitary gate, the Kraus noise operator and the Hermitian result type.
- `./test/unit_tests/braket/circuits/test_from_xfail.py` to test error cases.

## Merge Checklist

#### General

#### Tests


kshitijc:
Thanks for the PR @kjacky! This would be a very useful addition. Following are some initial thoughts:
* It'd be nice to break out the PR to focus on a single type of translation. This would help make it easier to review the changes. I'd recommend starting with `from_ir()` for the following reasons:
   * The IR contains the complete representation of the circuit. Other representations would truncate information (such as the exact value for an angle in an angled gate) or mask it (Unitary gate is represented as "Unitary" in an ASCII diagram and doesn't contain the information about the unitary matrix being used). This would cause the `from_repr` and `from_diagram` to not work in all cases.
   * The IR can always be retrieved post-execution (from the task execution results). The circuit repr and diagram are not stored by default. Hence, a customer won't be able to use the `from_repr` and `from_diagram` unless they explicitly store these, and if they already have a way to reconstruct the circuit from the stored representation (IR) it's questionable what value these methods would add.
   * Methods such as `from_diagram` can be very brittle since they rely on correctly parsing extraneous formatting of the circuit layout. The circuit formatting can change frequently causing these methods to break. 

* The current `from_ir` implementation is too complex (which you might have seen the build failure complaining about as well). We should try to think about how to simplify this. Here are a couple of ideas which might be worth considering:
   * The current implementation aggregates all the logic across multiple object types in a single method. This would be hard to maintain and reason about as we add in new object types. One way to solve this would be to delegate the responsibility to the different object types themselves. You might have seen that each class (Circuit, Instruction, Gate, ResultType) have a `to_ir()` method which contains logic on how that specific object can be converted to the IR representation, and the classes delegate this conversion where possible (Circuit delegates the IR conversion for instructions to the Instruction class, which in turn delegates the conversion for the quantum operator to the specific operator class). Similarly, each class could have a `from_ir()` method which knows how to construct an instance of the specific class from the IR representation.
   ```
   class Circuit:
     
     @classmethod
     def from_ir(cls, program: Program) -> Circuit:
        instructions = [Instruction.from_ir(program.instructions)]
        results = [ResultType.from_ir(program.results)]
        return cls(instructions + results)
   ```
  * We are creating some duplicated mappings `obj_class_defs`, `gate_defs`, `result_type_defs`, `observable_defs`, `gate_to_observable`. Consider how we could reuse the existing mappings to serve this use case. It might help to look at how some of the classes are storing this information using the `register` methods. (See [Gate](https://github.com/aws/amazon-braket-sdk-python/blob/main/src/braket/circuits/gate.py#L63), [Noise](https://github.com/aws/amazon-braket-sdk-python/blob/main/src/braket/circuits/noise.py#L81), [ResultType](https://github.com/aws/amazon-braket-sdk-python/blob/main/src/braket/circuits/result_type.py#L117))

kjacky:
Thank you very much @kshitijc for your guidance. I have removed `from_diagram` and `from_repr`, then refactored `from_ir` into individual `braket.circuits` classes. All `*_defs` and conversions were removed.

Each of the 16 `from_ir()` class methods returns an object of the `cls` type. All are 1-to-3 liners, except for `Instruction::from_ir()` with 15 non-comment lines implementing the core logic. Tests and coverage were passed 100%.

I've added two utility functions for internal use: `complex_matrices()` and `_attr_dict()` (near the bottom of `circuits/result_type.py` for lack of a better place I found - feel free to move them).

Looking forward to your feedback @kshitijc .

src/braket/circuits/result_types.py
```diff
@@ -44,6 +44,19 @@ def __init__(self):
     def to_ir(self) -> ir.StateVector:
         return ir.StateVector.construct()
 
+    @classmethod
+    def from_ir(cls, ir_result) -> StateVector:
```
kshitijc, line 48: Could we please add the type annotations to the parameters for these `from_ir` methods across all the files? 
(`def from_ir(cls, ir_result: ir.StateVector)`)

test/unit_tests/braket/circuits/test_from_ir.py
```diff
@@ -0,0 +1,86 @@
...
+    output_dir = f"{test_dir}/output_ir.cache"
+    if not os.path.isdir(output_dir):
+        os.mkdir(output_dir, 0o755)
+    result_dir = f"{test_dir}/result_ir.cache"
+    if not os.path.isdir(result_dir):
+        os.mkdir(result_dir, 0o755)
+
+    test_specs = read_file(f"{test_dir}/test_cases_ir.txt").strip("\n").split("\n")
+
+    for test_spec in test_specs:
```
kshitijc, line 53: Instead of looping in a single test (which would make it hard to detect which case failed), consider how these tests can be executed as separate cases (`pytest.mark.parametrize` is one possibility).

test/unit_tests/braket/circuits/test_from_ir.py
```diff
@@ -0,0 +1,86 @@
...
+    if not os.path.isdir(check_dir):
+        os.mkdir(check_dir, 0o755)
+    output_dir = f"{test_dir}/output_ir.cache"
+    if not os.path.isdir(output_dir):
+        os.mkdir(output_dir, 0o755)
+    result_dir = f"{test_dir}/result_ir.cache"
+    if not os.path.isdir(result_dir):
+        os.mkdir(result_dir, 0o755)
+
+    test_specs = read_file(f"{test_dir}/test_cases_ir.txt").strip("\n").split("\n")
```
kshitijc, line 51: It's hard to review the binary files since I can't read the content. Would it be possible to store these in a more readable format ?

test/unit_tests/braket/circuits/test_from_ir.py
```diff
@@ -0,0 +1,86 @@
...
+        write_file(f"{output_path}.ir_out", f"{ir_out}")
+        assert f"{ir_inp}" == f"{ir_out}", f"Failed {test_spec}: {test_case}.ir_inp != ir_out"
+
+        # Run the circuit created by from_ir() to make sure it is runnable
+        try:
+            task = devices[sim_dev].run(circ_from_ir, shots=shots)
+            result = task.result()
+            counts = result.measurement_counts
+            # Save the run result for manual checking if required
+            write_file(f"{result_path}.result_from_ir", f"{sort_key(counts)}")
```
kshitijc, line 84: This is much bigger than a unit test. A unit test is supposed to test the smallest piece of code that can be logically isolated in a system. A good way to think about how the tests should be built up is the [Test Pyramid](https://martinfowler.com/bliki/TestPyramid.html). Unit tests for each specific method which can test all the possible cases, then integration tests to test integration across multiple classes etc.

src/braket/circuits/instruction.py
```diff
@@ -89,6 +90,36 @@ def to_ir(self):
...
+
+        Returns:
+            Instruction: The instruction object created
+        """
+        type_name = type(ir_instruction).__name__
+        op_class = None
+        if hasattr(Gate, type_name):
+            op_class = getattr(Gate, type_name)
+        elif hasattr(Noise, type_name):  # pragma: no branch
+            op_class = getattr(Noise, type_name)
```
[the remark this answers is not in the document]
kjacky: Agree. Initially I had an else raising an error if `type_name` is neither a Gate or Noise. However, Coverage gave me a "miss" on the error raising line as it is never executed. Any suggestion to xfail test (with invalide `type_name`) in this case?

src/braket/circuits/gates.py
```diff
@@ -26,6 +28,7 @@
 )
 from braket.circuits.qubit import QubitInput
 from braket.circuits.qubit_set import QubitSet, QubitSetInput
+from braket.circuits.result_type import _attr_dict, complex_matrices
```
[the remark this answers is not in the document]
kjacky: Agree. I was tossing between `circuit_helpers.py` and `result_type.py`. Should I move these two functions to `circuit_helpers.py` or create a new file?

src/braket/circuits/instruction.py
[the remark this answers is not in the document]
kjacky: I have extended `InstructionOperator = Union[Gate, Noise]` and use `InstructionOperator.__args__` to iterate through instead of hardcoding Gate and Noise. If there're no matches, it will raise a `ValueError`.

A new unit test has been created under `test_instruction_from_ir.py` to loop through all types from the schemas asserting the type names: `from braket.ir.jaqcd.program_v1 import Results, _valid_gates, _valid_noise_channels`.

src/braket/circuits/instruction.py
```diff
@@ -89,6 +90,36 @@ def to_ir(self):
...
+        qubit_set = []
+        # Controls must come before targets
+        for attr_name in ["control", "controls", "target", "targets"]:
+            if hasattr(ir_instruction, attr_name):
+                attr = getattr(ir_instruction, attr_name)
+                # Flatten the list
+                if type(attr) == list:
+                    qubit_set += attr
+                else:
+                    qubit_set.append(attr)
```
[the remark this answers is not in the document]
kjacky: This block has been removed and delegated to individual gates, noise channels and result types. As a result, the `from_ir()` in gate and noise channel classes have been renamed to `ir_instr_to_op()` and `ir_instr_to_qubit_set()`. For result types, the name remains `from_ir()` as the `targets` argument is with the object.

src/braket/circuits/noise.py
```diff
@@ -378,3 +420,18 @@ def probability(self) -> float:
     def __repr__(self):
         return f"{self.name}('gamma': {self.gamma}, 'probability': {self.probability}, \
 'qubit_count': {self.qubit_count})"
+
+    @classmethod
+    def from_ir(cls, ir_instruction) -> GeneralizedAmplitudeDampingNoise:
```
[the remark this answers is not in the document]
kjacky: Yes, resolved as in my previous comment.

src/braket/circuits/gates.py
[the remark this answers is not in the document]
kjacky: I have created `circuit_utils.py` for these functions (as `circuit_helpers.py` include `Circuit` and therefore created a circular import).

src/braket/circuits/instruction.py
```diff
@@ -89,6 +90,36 @@ def to_ir(self):
...
+            ir_instruction: The IR instruction to create the Instruction object from
+
+        Returns:
+            Instruction: The instruction object created
+        """
+        type_name = type(ir_instruction).__name__
+        op_class = None
+        if hasattr(Gate, type_name):
+            op_class = getattr(Gate, type_name)
+        elif hasattr(Noise, type_name):  # pragma: no branch
```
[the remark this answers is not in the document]
kjacky: Yes, as commented in my previous comment.

src/braket/circuits/result_types.py
kjacky: Yes, added type annotations to all `ir_result` arguments.

test/unit_tests/braket/circuits/test_from_ir.py
kjacky: Yes, there're some cache directories to echo the circuit diagram for tester to check. In this new version, I've moved them to `./.cache/from_ir_test/check`. There're also IR output in `./.cache/from_ir_test/output` and run result in `./.cache/from_ir_test/result`.

test/unit_tests/braket/circuits/test_from_ir.py
kjacky: Yes, incorporated `pytest.mark.parametrize` to loop through the test instead of looping inside the test. Thanks for your advice @kshitijc 👍

test/unit_tests/braket/circuits/test_from_ir.py
kjacky: Yes, I've renamed `test_from_ir.py` to `test_circuit_from_ir.py` and created `test_instruction_from_ir.py`, which include `from braket.ir.jaqcd.program_v1 import Results, _valid_gates, _valid_noise_channels` and loop through each gate and noise channel. For `Result` I use a small 2-qubit circuit (Bell) and append different result types to it. I've kept both of them under `unit_tests` though. Do you think we need something in `integ_tests`?

kshitijc:
Closing PR since there has been no activity on this for a while. Please feel free to reopen if you wish to continue working on this. Thanks!

PR #260, opened 2021-07-29
fix: Pass along endpoint url when copying an AwsSession

kshyatt-aws:
Change `_copy_aws_session` to also track the endpoint URL of the original session to its clone. Need some help writing a nice test for this as my knowledge of `botocore` is rudimentary.

## Merge Checklist

#### General

#### Tests

### DIFF
--- a/src/braket/aws/aws_device.py
+++ b/src/braket/aws/aws_device.py
@@ -325,6 +325,7 @@ class AwsDevice(Device):
         session_region = aws_session.boto_session.region_name
         new_region = region or session_region
         creds = aws_session.boto_session.get_credentials()
+        client = aws_session.braket_client
         if creds.method == "explicit":
             boto_session = boto3.Session(
                 aws_access_key_id=creds.access_key,
@@ -334,7 +335,8 @@ class AwsDevice(Device):
             )
         else:
             boto_session = boto3.Session(region_name=new_region)
-        return AwsSession(boto_session=boto_session, config=config)
+        braket_client = boto_session.client('braket', endpoint_url=client._endpoint.host)
+        return AwsSession(boto_session=boto_session, braket_client=braket_client, config=config)
 
     def __repr__(self):
         return "Device('name': {}, 'arn': {})".format(self.name, self.arn)


src/braket/aws/aws_device.py
```diff
@@ -334,7 +335,8 @@ def _copy_aws_session(
             )
         else:
             boto_session = boto3.Session(region_name=new_region)
-        return AwsSession(boto_session=boto_session, config=config)
+        braket_client = boto_session.client('braket', endpoint_url=client._endpoint.host)
```
kshitijc, line 338: * It's generally advisable to not rely on private attributes. It might be a better to idea to access this as: `client.meta.endpoint_url`.
* The existing flow [would have constructed](https://github.com/aws/amazon-braket-sdk-python/blob/main/src/braket/aws/aws_session.py#L44) the client with the `config` on #324. In order to avoid any regressions, it might be a good idea to pass that in while constructing the client here.

kshitijc (commented):
Thank you for this fix!

* Please run `tox -e linters` to fix the build failures due to formatting.
* [This commit](https://github.com/aws/amazon-braket-sdk-python/commit/b06e04e03c8889d7da7e36c06fcb57c9749965a9) should help provide an idea on writing unit tests for this change.

kshyatt-aws:
Fixed the linting issues and copied the config to the client. I have a bunch of other tasks I should be working on and found the new triggered failures really opaque, so if someone wants to add commits to this, please do.

src/braket/aws/aws_device.py
```diff
@@ -334,7 +335,10 @@ def _copy_aws_session(
             )
         else:
             boto_session = boto3.Session(region_name=new_region)
-        return AwsSession(boto_session=boto_session, config=config)
+        braket_client = boto_session.client(
```
krneta, line 338: I'm concerned because the new session may be in a different region, so when you specify the endpoint url there might be a mismatch between the region and the url you're specifying. I think it (kind of) defeats the purpose of having this copy function in the first place. Can you explain the actual use case you have and what this code change is trying to fix?

As an aside, I think we really have to rethink this copy function. Technically, your code is making this function better! That is, if we need the aws session to really be copied, then we should copy the client including all the possible params to the client as well (such as api version, ssl and verify settings). But we hit a wall with that really quickly since we don't actually know what we should and shouldn't copy. For example, we probably don't want to copy the auth params - or maybe we only want to copy them if they were manually passed in in the first place. Then there are the concerns of having conflicts between the session and the client (as above).

PR #267, opened 2021-08-20
feature: Calculate arbitrary observables when `shots=0`

speller26:
Allows arbitrary observables to be attached to a circuit, allowing, for example, the simultaneous calculation of all terms of a qubit Hamiltonian when `shots=0`. However, running such a circuit with `shots>0`, will throw an exception, as observables must commute to be measured (sampled) simultaneously; in fact, here the observables must be either identical or the identity.

*Issue #, if available:*

*Description of changes:*

*Testing done:*

## Merge Checklist

#### General

#### Tests

### DIFF
--- a/src/braket/circuits/circuit.py
+++ b/src/braket/circuits/circuit.py
@@ -123,8 +123,9 @@ class Circuit:
         self._moments: Moments = Moments()
         self._result_types: Dict[ResultType] = {}
         self._qubit_observable_mapping: Dict[Union[int, Circuit._ALL_QUBITS], Observable] = {}
-        self._qubit_target_mapping: Dict[int, Tuple[int]] = {}
+        self._qubit_observable_target_mapping: Dict[int, Tuple[int]] = {}
         self._qubit_observable_set = set()
+        self._noncommuting_observables = False
 
         if addable is not None:
             self.add(addable, *args, **kwargs)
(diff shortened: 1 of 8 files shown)


src/braket/circuits/circuit.py
```diff
@@ -912,6 +901,20 @@ def as_unitary(self) -> np.ndarray:
 
         return calculate_unitary(qubit_count, self.instructions)
 
+    @property
+    def has_noncommuting_observables(self) -> bool:
```
licedric, line 905: I do wonder about this name, since we actually only support identical observables, not commuting observables. At the very least this should be renamed to non_qubitwise_commuting, but that's not completely accurate either. Probably need a second opinion from someone else.

licedric:
> Allows arbitrary observables to be attached to a circuit, allowing, for example, the simultaneous calculation of all terms of a qubit Hamiltonian when shots=0.

Clarify this a bit more - it should be clear what happens. Namely, the user can create a circuit with arbitrary observables, but an error is returned if it is run with shots > 0.

src/braket/circuits/circuit.py
speller26: How about I flip the truth value and call it `observables_simultaneously_measurable`?

licedric: Oh, that could work. Thanks!

licedric (approved):

src/braket/circuits/circuit.py
```diff
@@ -194,7 +198,7 @@ def add_result_type(
         self,
         result_type: ResultType,
         target: QubitSetInput = None,
-        target_mapping: Dict[QubitInput, QubitInput] = {},
+        target_mapping: Dict[QubitInput, QubitInput] = None,
```
licedric, line 201: Why?

speller26: https://florimond.dev/en/posts/2018/08/python-mutable-defaults-are-the-source-of-all-evil/#:~:text=In%20Python%2C%20when%20passing%20a,or%20even%20a%20class%20instance.

Will never forget the headaches this has caused me in the past

PR #291, opened 2021-12-07
feature: Add parameterized circuits

math411:
*Issue #, if available:*

*Description of changes:*
feature: Add parameterized circuits. Parameterized circuits cannot be executed. Instead binding **all** parameters with the function ```make_bound_circuit()``` or the __call__ function will give a circuit that is then able to be ran.

How to use:

Creation:
```
    theta = FreeParameter("theta")
    # Note, equality is based off of the name so assigning values will assign for both obj alpha and alpha2.
    alpha = FreeParameter("alpha")
    alpha2 = FreeParameter("alpha")

    circ = Circuit().ry(angle=theta, target=0).ry(angle=alpha2, target=1).ry(angle=theta, target=2)

    # Parameterized instructions can be added in later to a circuit.
    circ.add_instruction(Instruction(Gate.Ry(alpha), 3))
```

Binding parameters:
```
    theta = FreeParameter("theta")
    alpha = FreeParameter("alpha")
    input_val = np.pi
    circ = Circuit().ry(angle=theta, target=0).ry(angle=theta, target=1).ry(angle=alpha, target=2)

    # A new circuit with all alpha and theta both bound to pi. The new circuit reflects these changes.
    circ_new = circ.make_bound_circuit({"theta": input_val, "alpha": input_val})
```

```
    # This use case is demonstrating the way to use __call__ for a circuit. It supports a single arg to bulk assign parameters
    # or kwargs which assign based on the keyword provided. Kwargs supersedes arg bulk assignment.

    alpha = FreeParameter("alpha")
    beta = FreeParameter("beta")
    theta = FreeParameter("theta")
    gamma = FreeParameter("gamma")
    circ = (
        Circuit()
        .h(0)
        .rx(angle=theta, target=1)
        .ry(angle=alpha, target=0)
        .ry(angle=beta, target=2)
        .rx(angle=gamma, target=1)
        .add_result_type(prob)
    )

   # Kwargs of theta and alpha get set to their respective values. Any other parameters in a circuit get bound to pi.
    new_circ = circ(np.pi, theta=1, alpha=0)
```

*Testing done:*
```tox```

## Merge Checklist

#### General

- [ x ] I have read the [CONTRIBUTING](https://github.com/aws/amazon-braket-sdk-python/blob/main/CONTRIBUTING.md) doc
- [ x ] I used the commit message format described in [CONTRIBUTING](https://github.com/aws/amazon-braket-sdk-python/blob/main/CONTRIBUTING.md#commit-your-change)
- [ x ] I have updated any necessary documentation, including [READMEs](https://github.com/aws/amazon-braket-sdk-python/blob/main/README.md) and [API docs](https://github.com/aws/amazon-braket-sdk-python/blob/main/CONTRIBUTING.md#documentation-guidelines) (if appropriate)

#### Tests

- [ x ] I have added tests that prove my fix is effective or that my feature works (if appropriate)
- [ x ] I have checked that my tests are not configured for a specific region or account (if appropriate)

### CODE REVIEW

File: src/braket/circuits/circuit.py [hunk 1]
```diff
@@ -412,10 +434,30 @@ def add_instruction(
...
+
+        Returns: True if the angle is not a number. False otherwise.
+        """
+        # Check for Angled Gates.
+        if (
+            issubclass(type(instruction.operator), QuantumOperator)
+            and instruction.operator.parameterized
+        ):
+            return True
+        return False
```
speller26, line 459: Another reason why `parameter` should live with AngledGate rather than QuantumOperator.

src/braket/circuits/parameterized_operator.py
```diff
@@ -0,0 +1,46 @@
...
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+
+from __future__ import annotations
+
+from braket.circuits.free_parameter import FreeParameter
+
+
+class ParameterizedOperator:
```
speller26, line 19: I don't think "operator" should be in the name of the class; after all, there might be non-operator classes that need parametrization.

speller26: Also, how will this deal with classes that take multiple parameters, like some Pauli noise?

src/braket/circuits/circuit.py
```diff
@@ -196,6 +199,27 @@ def qubits(self) -> QubitSet:
         """QubitSet: Get a copy of the qubits for this circuit."""
         return QubitSet(self._moments.qubits.union(self._qubit_observable_set))
 
+    @property
+    def is_parameterized(self) -> bool:
```
speller26, line 203: I think this name is ambiguous; it's a mouthful, but maybe something like `has_free_parameters`

src/braket/circuits/angled_gate.py
```diff
@@ -41,21 +48,31 @@ def __init__(self, angle: float, qubit_count: Optional[int], ascii_symbols: Sequ
...
         return self._angle
 
     def __eq__(self, other):
         if isinstance(other, AngledGate):
-            return self.name == other.name and math.isclose(self.angle, other.angle)
+            if isinstance(self.angle, FreeParameter):
+                return self.name == other.name and self.angle == other.angle
+            else:
+                return self.name == other.name and math.isclose(self.angle, other.angle)
         return NotImplemented
```
speller26, line 76: I know it's not part of the original scope this PR, but might as well fix it; `NotImplemented` is a truthy value, so it should actually be `False`.

src/braket/circuits/circuit.py
```diff
@@ -196,6 +199,27 @@ def qubits(self) -> QubitSet:
...
+            bool: True if the circuit has any free parameters.
+        """
+        return self._parameterized
+
+    @property
+    def parameters(self) -> set:
+        """
+        Gets a set of the parameters in the Circuit.
+        Returns:
+            set: A set of FreeParameters in the Circuit.
```
speller26, line 219: Narrow down the signature (even if Python doesn't enforce it)
```suggestion
    def parameters(self) -> Set[FreeParameter]:
        """
        Gets a set of the parameters in the Circuit.
        
        Returns:
            Set[FreeParameter]: The `FreeParameters` in the Circuit.
```

src/braket/circuits/circuit.py
```diff
@@ -412,10 +436,32 @@ def add_instruction(
...
+        Returns:
+            bool: Whether an object is parameterized.
+        """
+        # Check for Angled Gates.
+        if (
+            issubclass(type(instruction.operator), ParameterizedOperator)
+            and instruction.operator.parameterized
+        ):
+            return True
+        return False
```
speller26, line 463:
```suggestion
        return (
            issubclass(type(instruction.operator), ParameterizedOperator)
            and instruction.operator.parameterized
        )
```

src/braket/circuits/circuit.py
```diff
@@ -773,6 +822,16 @@ def apply_initialization_noise(
 
         return apply_noise_to_moments(self, noise, target_qubits, "initialization")
 
+    def set_parameter_values(self, param_values: Dict):
+        """
+        Sets FreeParameters based upon values passed in.
+
+        Args:
+            param_values:  A mapping of FreeParameters to a list of values to assign to them.
```
speller26, line 830:
```suggestion
    def set_parameter_values(self, param_values: Dict[FreeParameter, float]) -> None:
        """
        Sets FreeParameters based upon values passed in.

        Args:
            param_values Dict[FreeParameter, float]:  A mapping of FreeParameters
                to a list of values to assign to them.
```

src/braket/circuits/free_parameter.py
```diff
@@ -0,0 +1,114 @@
...
+        self._name = name
+        self._symbol = parameter_symbol
+        self._parameter_expression = parameter_symbol
+        self._parameter_values = []
+
+    @property
+    def name(self) -> str:
+        """
+        Returns:
+            str: Name for :class:'FreeParameter'.
```
speller26, line 47: property docstrings are a bit different:
```suggestion
        str: Name of this parameter.
```
Same goes for other properties

src/braket/circuits/free_parameter.py
```diff
@@ -0,0 +1,114 @@
...
+
+    @property
+    def symbol(self) -> Symbol:
+        """
+        Returns:
+            Symbol: Symbol for :class:'FreeParameter'.
+        """
+        return self._symbol
+
+    def fix_values(self, param_values):
```
speller26, line 67: Type annotations:
```suggestion
    def fix_values(self, param_values: ?) -> None:
```

src/braket/circuits/free_parameter.py
```diff
@@ -0,0 +1,114 @@
...
+        """
+        Sets the values for a :class:'FreeParameter'.
+
+        Args:
+            param_values: The values to be set for the object.
```
speller26, line 72: How are circuits run if multiple values are fixed? Maybe change this to just a single value for now.

src/braket/circuits/free_parameter.py
```diff
@@ -0,0 +1,114 @@
...
+    def __str__(self):
+        return self.name
+
+    def __hash__(self) -> int:
+        return hash(tuple(self._name))
+
+    def __eq__(self, other):
+        if isinstance(other, FreeParameter):
+            return self.name == other.name
+        return NotImplemented
```
speller26, line 106:
```suggestion
        return False
```

src/braket/aws/aws_quantum_task.py
```diff
@@ -133,6 +133,10 @@ def create(
         )
         if tags is not None:
             create_task_kwargs.update({"tags": tags})
+        if isinstance(task_specification, Circuit) and task_specification.is_parameterized:
+            raise ValueError(
+                f"Cannot execute parameterized circuit. " f"Invalid circuit: ({task_specification})"
```
speller26, line 138: Maybe list out the unbound parameters

src/braket/circuits/angled_gate.py
```diff
@@ -41,22 +48,32 @@ def __init__(self, angle: float, qubit_count: Optional[int], ascii_symbols: Sequ
...
         """
         Returns the angle for the gate
 
         Returns:
-            angle (float): The angle of the gate in radians
+            angle (Union[FreeParameter, float]): The angle of the gate in radians
         """
         return self._angle
 
     def __eq__(self, other):
```
licedric, line 70: Can you describe in a comment what equality does here?
Also, is this really what we want?

math411: This checks the angled gate is equal based on the angle based upon our previous definition. As the angle can be a float or free parameter, equality is checked for the two branches.

licedric: what if self.angle is a float but other.angle is a FreeParameter?

src/braket/circuits/circuit.py
```diff
@@ -773,6 +837,17 @@ def apply_initialization_noise(
 
         return apply_noise_to_moments(self, noise, target_qubits, "initialization")
 
+    def set_parameter_value(self, param_values: Dict[FreeParameter, Number]):
```
licedric, line 840: No way to unset parameters right now? What if we want to run a parametrized circuit multiple times with different parameter values?

math411: Can be called in a loop with parameter values used for execution and sending the circuits with these vals

math411: The free parameter value would be overridden with the next value with subsequent set_parameter_value calls

src/braket/circuits/angled_gate.py
math411:
```
self.angle == other.angle
```
Would run a check of a free parameter against a float here. FP has a check of the type for the equality operator so it would be false in this case.

src/braket/circuits/free_parameter.py
```diff
@@ -0,0 +1,112 @@
...
+                f"Invalid inputs: {param_value}"
+            )
+
+    def __str__(self):
+        return self.name
+
+    def __hash__(self) -> int:
+        return hash(tuple(self._name))
+
+    def __eq__(self, other):
```
licedric, line 101: Is this what we expect? What does SymPy do if two variables have been assigned the same value?

src/braket/circuits/free_parameter.py
```diff
@@ -0,0 +1,112 @@
...
+
+    @property
+    def symbol(self) -> Symbol:
+        """
+        Returns:
+            Symbol: The symbol for a parameter.
+        """
+        return self._symbol
+
+    def fix_value(self, param_value: Number) -> None:
```
licedric, line 68: Would it make sense just to assign the symbol instead? Otherwise it doesn't look like you're using SymPy functionality at all.

src/braket/circuits/angled_gate.py
licedric: And this is true even if the free parameter has been set to the same value as the float?

src/braket/circuits/free_parameter.py
math411: The symbol would be used in conjunction with the expression for future use cases similar to the physics use cases captured outside of this PR. Also, later support for multiple values would be beneficial to retain this.

src/braket/circuits/free_parameter.py
math411: https://github.com/sympy/sympy/blob/master/sympy/core/symbol.py#L264
Equality is similarly down with the name attributed.

src/braket/circuits/angled_gate.py
math411: Yes FreeParameter equality is checked based on the name and the instance type. The check would return false as the comparison would be between a FreeParameter and a float

src/braket/circuits/parameterized_operator.py
```diff
@@ -0,0 +1,46 @@
...
+    """A parameterized operator is the abstract definition of a
+    parameterizable operation for a quantum device."""
```
yitchen-tim, line 21: Is there a use case of `parameterized_operator`? I don't see an example in the test case. Is it not true that we can always define a operator with `free_parameter` without the need of `parameterized_operator`?

src/braket/circuits/free_parameter.py
```diff
@@ -0,0 +1,112 @@
...
+            return self.name == other.name
+        return False
+
+    def __repr__(self):
+        """
+        The representation of the :class:'FreeParameter'.
+
+        Returns: The name of the class:'FreeParameter' to represent the class.
+        """
+        return self.name
```
yitchen-tim, line 112: It might be niche, but there are use cases that the angle of a gate is a product of two variables. Would it easy to also add multiplication of `free_parameter`? 

A use case is when users try to implement a variation of [this circuit](https://pennylane.readthedocs.io/en/stable/code/api/pennylane.IQPEmbedding.html). Without multiplication, user need to define `x1*x2` as a free parameter in addition to `x1` and `x2`.

yitchen-tim (commented):
Thanks Abe `free_parameter` is a cool feature! Some comments about AHS work stream: In AHS, SymPy is used to represented complex mathematical form that consist of many variables. The variables are coded in SymPy primarily because we need to use the `sympy.simplify` features. It is very different from here where SmyPy is only used to represent standalone trainable parameters. I don't think AHS classes are able to use the `free_parameter` class due to lack of required features. But I also don't think we necessarily need to change/expand `free_parameter` class. We can just define different classes in AHS.

src/braket/circuits/angled_gate.py
licedric: Wanted to stress again that many users are accustomed to SymPy - what's the behavior there? I'm hoping anything we do here will not be too surprising.

src/braket/circuits/free_parameter.py
licedric: I actually thought the consensus was to make FreeParameter a wrapper around SymPy, but as it stands it doesn't look like you're using SymPy at all - could have just used a string for a name and a float for a value instead.

src/braket/circuits/free_parameter.py
math411: Currently, without expressions acting on free_parameter, this would have to be done by using multiplication for the values of `x1` and `x2`. In the future, Something such as `rX(x1x2)` would be supported instead of having to pass values in manually.

src/braket/circuits/parameterized_operator.py
math411: This is currently being used in the class structure for `AngledGate`. You can make the a `free_parameter` without a `parameterized_operator`.

src/braket/circuits/free_parameter.py
math411: Fixed to a singular value but the strategy for multiple value input will be implemented with the execution PR

src/braket/aws/aws_quantum_task.py
math411: Going to update this code for the execution PR. For now, as we are entirely blocking execution and just producing the object code, the message shall indicate that the execution of parameterized circuits is not supported.

src/braket/circuits/angled_gate.py
```diff
@@ -41,22 +48,32 @@ def __init__(self, angle: float, qubit_count: Optional[int], ascii_symbols: Sequ
         super().__init__(qubit_count=qubit_count, ascii_symbols=ascii_symbols)
         if angle is None:
             raise ValueError("angle must not be None")
-        self._angle = float(angle)  # explicit casting in case angle is e.g. np.float32
+        if isinstance(angle, FreeParameter):
+            self._angle = angle
+            self._parameterized = True
```
krneta, line 53: Should we actually need to keep track of this, or should we just return isinstance of angle in the property?

src/braket/circuits/angled_gate.py
krneta: Yeah, while I agree that we shouldn't do stuff that seems weird for the customers, we don't want to be restricted either if what SymPy does isn't right for us. However, I completely agree that we should understand what SymPy does and make a conscious decision about what path we'll take.

src/braket/circuits/circuit.py
```diff
@@ -196,6 +201,28 @@ def qubits(self) -> QubitSet:
         """QubitSet: Get a copy of the qubits for this circuit."""
         return QubitSet(self._moments.qubits.union(self._qubit_observable_set))
 
+    @property
+    def has_free_parameters(self) -> bool:
```
krneta, line 205: I'm worried that we're going to have trouble tracking this. If this is just an optimization, how expensive would it be to instead check the circuit before compilation?

src/braket/circuits/circuit.py
```diff
@@ -196,6 +201,28 @@ def qubits(self) -> QubitSet:
...
+
+    @property
+    def parameters(self) -> Set[FreeParameter]:
+        """
+        Gets a set of the parameters in the Circuit.
+
+        Returns:
+            Set[FreeParameter]: The `FreeParameters` in the Circuit.
+        """
+        return self._parameters
```
krneta, line 224: Is self._parameters == (self._parameters != None) ?

krneta: Can you explain why we actually need to keep track of this? I believe it's not just for optimization purposes, right?

src/braket/circuits/circuit.py
```diff
@@ -484,6 +530,9 @@ def add_circuit(
             values = target
             target_mapping = dict(zip(keys, values))
 
+        if circuit.has_free_parameters:
+            self._parameterized = True
```
krneta, line 534: Wouldn't the "add_instruction" call already do this? Why do we need to do it twice?

src/braket/circuits/circuit.py
```diff
@@ -773,6 +822,40 @@ def apply_initialization_noise(
...
+
+        Args:
+            param_values Dict[str, Number]:  A mapping of FreeParameter names
+                to a value to assign to them.
+
+        Raises:
+            ValueError: If there are no parameters that match the key for the arg
+                param_values.
+        """
+        for param in param_values:
```
krneta, line 837: This function makes me think that maybe we want self._parameters to a dictionary (with a different name) and for self.parameters to return the dictionary values? Maybe?

src/braket/circuits/circuit.py
```diff
@@ -773,6 +822,40 @@ def apply_initialization_noise(
 
         return apply_noise_to_moments(self, noise, target_qubits, "initialization")
 
+    def set_parameter_values(self, param_values: Dict[str, Number]):
+        """
+        Sets FreeParameters based upon their name and values passed in.
```
krneta, line 827: We might want to mention that if a circuit has more than one FreeParameter with the same name, all of them will be set to the value.

src/braket/circuits/circuit.py
```diff
@@ -773,6 +822,40 @@ def apply_initialization_noise(
...
+        """
+        Populates a dictionary of the FreeParameters to their values in this circuit.
+
+        Returns:
+            Dict[FreeParameter, Number]: Returns a mapping of FreeParameter and
+                the corresponding value.
+        """
+        param_values = {}
+        for param in self.parameters:
+            param_values[param] = param.parameter_value
```
krneta, line 856: What happens if two parms (with the same name) have two different values? Do we ever check for that?

src/braket/circuits/free_parameter.py
```diff
@@ -0,0 +1,101 @@
...
+    Class 'FreeParameter'
+    """
+
+    def __init__(self, name: str):
+        """
+        Initializes a new :class:'FreeParameter' object.
+        Free parameters can be used in parameterized circuits.
+
+        Args:
+            name: Name of the :class:'FreeParameter'. Can be a unicode value.
```
krneta, line 30: shouldn't this just be a str?

src/braket/circuits/free_parameter.py
```diff
@@ -0,0 +1,101 @@
...
+
+        Examples:
+            >>> param1 = FreeParameter("theta")
+            >>> param1 = FreeParameter("\u03B8")
+        """
+        self._name = Symbol(name)
```
krneta, line 36: Why is _name a Symbol here? Why not just a str (for now)?

src/braket/circuits/free_parameter.py
```diff
@@ -0,0 +1,101 @@
...
+
+        """
+        if not isinstance(param_value, Number):
+            raise ValueError(
+                f"Parameters value assignment can only take numeric values. "
+                f"Invalid inputs: {param_value}"
+            )
+
+    def __str__(self):
+        return self.name
```
krneta, line 85: This would expose the name as a Symbol, wouldn't it?

src/braket/circuits/free_parameter.py
```diff
@@ -0,0 +1,101 @@
...
+
+    def __hash__(self) -> int:
+        return hash(tuple(self.name))
```
krneta, line 88: This would also rely on Symbol functionality, right?

src/braket/circuits/parameterized.py
```diff
@@ -0,0 +1,45 @@
...
+from __future__ import annotations
+
+from braket.circuits.free_parameter import FreeParameter
+
+
+class Parameterized:
+    """A parameterized object is the definition of an object that can take in FreeParameters"""
+
+    def __init__(self):
+        self._parameterized = False
```
krneta, line 23: I don't see why we need both of these. What will we do if they conflict (it says it's not parameterized and has a _parameter value, or vise-versa)?

test/unit_tests/braket/aws/test_aws_device.py
```diff
@@ -238,6 +238,36 @@ def test_gate_model_sim_schema():
 )
 
 
+@pytest.fixture
+def quantum_task(aws_session, s3_destination_folder):
```
krneta, line 242: can you call this parameterized_task?

test/unit_tests/braket/aws/test_aws_device.py
```diff
@@ -238,6 +238,36 @@ def test_gate_model_sim_schema():
...
+        aws_session=aws_session,
+        poll_timeout_seconds=2,
+        task_specification=circ,
+        shots=10,
+        s3_destination_folder=s3_destination_folder,
+    )
+
+
+@pytest.fixture
+def quantum_task_batch(aws_session, s3_destination_folder):
```
krneta, line 256: can you call this parameterized_task_batch?

test/unit_tests/braket/aws/test_aws_device.py
krneta: Won't this always fail?

test/unit_tests/braket/aws/test_aws_device.py
krneta: Won't this always fail?

test/unit_tests/braket/circuits/test_angled_gate.py
```diff
@@ -61,6 +61,26 @@ def test_equality(angled_gate):
...
 
 
+def test_symbolic_equality():
+    symbol1 = FreeParameter("theta")
+    symbol2 = FreeParameter("phi")
+    gate1 = AngledGate(angle=symbol1, qubit_count=1, ascii_symbols=["bar"])
+    gate2 = AngledGate(angle=symbol1, qubit_count=1, ascii_symbols=["foo"])
+    other_gate = AngledGate(angle=symbol2, qubit_count=1, ascii_symbols=["foo"])
+
+    assert gate1 == gate2
```
krneta, line 71: Can you add tests/assertions where the FreeParameter objects are different, but their names are the same? And tests/assertions for what happens when you fix parameters (with same/different values).

test/unit_tests/braket/circuits/test_ascii_circuit_diagram.py
```diff
@@ -27,6 +35,35 @@ def test_one_gate_one_qubit():
...
+    expected = "\n".join(expected)
+
+    assert AsciiCircuitDiagram.build_diagram(circ) == expected
+
+
+def test_one_gate_one_qubit_rotation_with_unicode():
+    theta = FreeParameter("\u03B8")
+    circ = Circuit().rx(angle=theta, target=0)
+    # Column formats to length of the gate plus the ascii representation for the angle.
+    expected = ("T  : |  0  |", "            ", "q0 : -Rx(θ)-", "", "T  : |  0  |")
```
krneta, line 61: OK, this is pretty cool. :)

test/unit_tests/braket/circuits/test_circuit.py
```diff
@@ -1306,3 +1327,102 @@ def test_diagram(h):
 
     assert h.diagram(mock_diagram) == expected
     mock_diagram.build_diagram.assert_called_with(h)
+
+
+def test_add_parameterized_check_true():
```
krneta, line 1332: can you please name the function based on what you're testing. I think you're testing "has_free_parameters" here, right?

test/unit_tests/braket/circuits/test_circuit.py
```diff
@@ -1306,3 +1327,102 @@ def test_diagram(h):
...
+
+def test_get_parameter_value_multiple():
+    theta = FreeParameter("theta")
+    alpha = FreeParameter("alpha")
+    input_val = np.pi
+    circ = Circuit().ry(angle=theta, target=0).ry(angle=theta, target=1).ry(angle=alpha, target=2)
+    circ.set_parameter_values({"theta": input_val, "alpha": input_val})
+
+    expected = {theta: input_val, alpha: input_val}
+    assert circ.get_parameter_values() == expected
```
krneta, line 1428: What do we need to test as far as Circuit mutation is concerned? For example, when a circuit has noise added to it, or it is mutated in some other way?

test/unit_tests/braket/circuits/test_free_parameter.py
```diff
@@ -0,0 +1,72 @@
...
+@pytest.fixture
+def free_parameter():
+    return FreeParameter("theta")
+
+
+def test_is_free_param(free_parameter):
+    assert isinstance(free_parameter, FreeParameter)
+
+
+def test_equality():
```
krneta, line 34: You might want to do this with fixtures.

krneta: I think some of the same comments I had about testing Circuits apply here.

test/unit_tests/braket/aws/test_aws_device.py
```diff
@@ -501,6 +531,36 @@ def test_run_no_extra(aws_quantum_task_mock, device, circuit):
     )
 
 
+@pytest.mark.xfail(raises=ValueError)
```
krneta, line 534: Where are we actually testing the passing version of this?

math411: Right now execution with free parameters is blocked. In a subsequent PR, the execution will be enabled.

src/braket/circuits/angled_gate.py
math411: This is used as a check across objects that also use `Parameterized`. As some parameterized objects in the future may not be `AngledGates`, there is a need for a generalized way of tracking this for different instructions.

src/braket/circuits/circuit.py
math411: Depends on the circuit size. This check is being used to note the existence of FreeParameters in the circuit. The plan for execution is to have the values replace at the time of execution. This will serve as a means of going through circuits that need this replacement.

src/braket/circuits/circuit.py
math411: This is for tracking the parameters to update later on. Additionally, customers may try and access circuit free parameters later on. It can be that the set is empty if there are no free parameters in the circuit.

src/braket/circuits/free_parameter.py
math411: As the representation for the Symbol is a str and the plan in the future is to start using expressions, I opted to use Symbol now so it is one less area to worry about refactoring in the future.

src/braket/circuits/free_parameter.py
math411: Yes, changed to use str.

src/braket/circuits/free_parameter.py
math411: As sympy returns a string for this, this will behave the same as if we were also using a string.

test/unit_tests/braket/aws/test_aws_device.py
math411: This will always for fail for this PR. Changed to parameterized_quantum_task

test/unit_tests/braket/circuits/test_circuit.py
math411: These are test cases I will capture in the execution side of things. I was worried that this already long PR, may have too much to keep track of with added mutations.

src/braket/circuits/circuit.py
```diff
@@ -773,6 +819,41 @@ def apply_initialization_noise(
 
         return apply_noise_to_moments(self, noise, target_qubits, "initialization")
 
+    def set_parameter_values(self, param_values: Dict[str, Number]):
```
krneta, line 822: What about a strict parameter?

math411: Added in.

test/unit_tests/braket/circuits/test_free_parameter.py
math411: Add in equality tests

test/unit_tests/braket/circuits/test_angled_gate.py
math411: Added in different equality checks in the free parameter checks

test/unit_tests/braket/aws/test_aws_device.py
math411: Yes, for now. Adding in description around the changes for the subsequent PR

src/braket/circuits/parameterized.py
math411: Remove

src/braket/circuits/circuit.py
math411: Removed function.

src/braket/circuits/circuit.py
math411: Removed parameterized from circuit

src/braket/circuits/gates.py
```diff
@@ -454,11 +455,13 @@ class Rx(AngledGate):
...
     Args:
-        angle (float): angle in radians.
+        angle (Union[FreeParameter, float]): angle in radians.
     """
 
-    def __init__(self, angle: float):
-        super().__init__(angle=angle, qubit_count=None, ascii_symbols=["Rx({:.3g})".format(angle)])
+    def __init__(self, angle: Union[FreeParameter, float]):
+        super().__init__(
+            angle=angle, qubit_count=None, ascii_symbols=["Rx({:})".format(str(angle))]
```
licedric, line 463: Why get rid of formatting?

math411: Angle can be a string or a float. This is to accompany for the longer strings.

src/braket/circuits/angled_gate.py
```diff
@@ -41,22 +48,30 @@ def __init__(self, angle: float, qubit_count: Optional[int], ascii_symbols: Sequ
         super().__init__(qubit_count=qubit_count, ascii_symbols=ascii_symbols)
         if angle is None:
             raise ValueError("angle must not be None")
-        self._angle = float(angle)  # explicit casting in case angle is e.g. np.float32
+        if isinstance(angle, FreeParameter):
+            self._angle = angle
+            self._parameter = angle
```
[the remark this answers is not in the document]
math411: There was a type casting to float before so leaving that in and then handling the other case where we have a FreeParam

src/braket/circuits/circuit.py
```diff
@@ -773,6 +805,31 @@ def apply_initialization_noise(
...
+    def set_parameter_values(self, param_values: Dict[str, Number], strict: bool = False):
+        """
+        Sets FreeParameters based upon their name and values passed in. If parameters
+        share the same name, all the parameters of that name will be set to the mapped value.
+
+        Args:
+            param_values (Dict[str, Number]):  A mapping of FreeParameter names
+                to a value to assign to them.
+
+            strict (bool, optional): A flag to only accept FreeParameters in the circuit
```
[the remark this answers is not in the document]
math411: Updated

src/braket/circuits/angled_gate.py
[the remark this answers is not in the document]
speller26: Still feels redundant. An angle should just be a union of float and FreeParameter; once the parameter is bount, you have a new circuit with no free parameters.

src/braket/circuits/circuit.py
```diff
@@ -412,10 +426,28 @@ def add_instruction(
...
 
         return self
 
+    def _check_for_params(self, instruction: Instruction) -> bool:
+        """
+        This checks for free parameters in an :class:{Instruction}. Checks children classes of
+        :class:{Parameterizable}.
+
+        Args:
+            instruction: The instruction to check for a :class:{FreeParameter}.
```
speller26, line 441: Annotate with type; applies everywhere else too:
```suggestion
            instruction (Instruction): The instruction to check for a :class:{FreeParameter}.
```

src/braket/circuits/free_parameter.py
speller26: I generally agree, if Freeparameter is a wrapper around sympy, we should make use of sympy.

My question is, what happens to the symbol after it's assigned? Does it stick around or is it replaced wholesale? Also, how does this fit into the idea that fixing parameters in a circuit creates a new circuit with fewer FreeParameter objects? Finally, how does this handle parallelism, eg. if we want to run the circuit with multiple values at once, say for a gradient calculation?

src/braket/circuits/free_parameter.py
```diff
@@ -0,0 +1,101 @@
...
+    Class 'FreeParameter'
+    """
+
+    def __init__(self, name: str):
+        """
+        Initializes a new :class:'FreeParameter' object.
+        Free parameters can be used in parameterized circuits.
+
+        Args:
+            str: Name of the :class:'FreeParameter'. Can be a unicode value.
```
speller26, line 30:
```suggestion
            name (str): Name of the :class:'FreeParameter'. Can be a unicode value.
```

src/braket/circuits/angled_gate.py
```diff
@@ -41,22 +48,30 @@ def __init__(self, angle: float, qubit_count: Optional[int], ascii_symbols: Sequ
...
 
     @property
-    def angle(self) -> float:
+    def angle(self) -> Union[FreeParameter, float]:
         """
         Returns the angle for the gate
 
         Returns:
-            angle (float): The angle of the gate in radians
+            angle (Union[FreeParameter, float]): The angle of the gate in radians
```
speller26, line 64: Returns aren't named:
```suggestion
            Union[FreeParameter, float]: The angle of the gate in radians
```

src/braket/circuits/circuit.py
```diff
@@ -412,10 +426,28 @@ def add_instruction(
...
+
+        Returns:
+            bool: Whether an object is parameterized.
+        """
+        return (
+            issubclass(type(instruction.operator), Parameterizable)
+            and instruction.operator.parameter
```
math411, line 448: isinstance(instruction.operator.parameter, FreeParameter) change for Angled classes as well

src/braket/circuits/parameterized_operator.py
math411: Fixed this

src/braket/aws/aws_quantum_task.py
math411: Ascii output will include parameters that are not set.

src/braket/circuits/circuit.py
```diff
@@ -773,6 +805,56 @@ def apply_initialization_noise(
...
+        Returns:
+            Circuit: Returns a circuit with all present parameters fixed to their respective
+                values.
+
+        Raises:
+            ValueError: If there are no parameters that match the key for the arg
+                param_values.
+        """
+        fixed_circ = Circuit()
+        if strict:
```
krneta, line 831: Apparently I commented on the diff rather than the pull request. Please see my comments in the diff.

math411: Ack

krneta:
> Will update with docs. This was a quick way to set values. As setting values now returns a new circuit, it will return the fixed value circuit.

But it's really not intuitive to me...

src/braket/aws/aws_quantum_task_batch.py
```diff
@@ -121,6 +121,12 @@ def _execute(
         *args,
         **kwargs,
     ):
+        for task_specification in task_specifications:
+            if isinstance(task_specification, Circuit) and task_specification.parameters:
+                raise ValueError(
+                    f"Cannot execute parameterized circuit. "
+                    f"Invalid circuit: ({task_specification})"
```
speller26, line 128: Same here

src/braket/circuits/circuit.py
```diff
@@ -1044,6 +1126,15 @@ def __eq__(self, other):
             )
         return NotImplemented
 
+    def __call__(self, arg=False, **kwargs) -> Circuit:
```
krneta, line 1129: Why False and not None?

src/braket/circuits/ascii_circuit_diagram.py
```diff
@@ -84,6 +84,10 @@ def build_diagram(circuit) -> str:
         if additional_result_types:
             lines.append(f"\nAdditional result types: {', '.join(additional_result_types)}")
 
+        # A list of parameters in the circuit to the currently assigned values.
+        for parameter in circuit.parameters:
+            lines.append(f"\n{parameter} has no assigned value.")
```
speller26, line 89: Probably want something like this; otherwise you might end up with many more lines than is readable.
```suggestion
        lines.append(f"Unassigned parameters: {circuit.parameters}")
```

src/braket/circuits/circuit.py
```diff
@@ -773,6 +805,56 @@ def apply_initialization_noise(
...
+        self, param_values: Dict[str, Number], strict: bool = False
+    ) -> Circuit:
+        """
+        Sets FreeParameters based upon their name and values passed in. If parameters
+        share the same name, all the parameters of that name will be set to the mapped value.
+
+        Args:
+            param_values (Dict[str, Number]):  A mapping of FreeParameter names
+                to a value to assign to them.
+
```
speller26, line 818: nit: extra space

src/braket/circuits/circuit.py
```diff
@@ -1044,6 +1126,15 @@ def __eq__(self, other):
             )
         return NotImplemented
 
+    def __call__(self, arg=False, **kwargs) -> Circuit:
+        param_values = dict()
+        if arg:
```
speller26, line 1131: What if `arg==0`?

src/braket/circuits/circuit.py
```diff
@@ -1044,6 +1126,15 @@ def __eq__(self, other):
...
+            for param in self.parameters:
+                param_values[str(param)] = arg
+        for key, val in kwargs.items():
+            param_values[str(key)] = val
```
speller26, line 1135: Beautiful

src/braket/circuits/parameterizable.py
```diff
@@ -0,0 +1,36 @@
...
+
+
+class Parameterizable:
+    """A parameterized object is the definition of an object that can take in FreeParameters"""
+
+    def __init__(self):
+        self._parameters = list()
+
+    @property
+    def parameters(self) -> List[Union[FreeParameter, float]]:
```
speller26, line 28: Might be worth making the class abstract and this an abstract property.

src/braket/circuits/circuit.py
```diff
@@ -13,12 +13,14 @@
 
 from __future__ import annotations
 
-from typing import Callable, Dict, Iterable, List, Optional, Tuple, Type, TypeVar, Union
+from numbers import Number
```
speller26, line 16: Why only use `Number` here and `float` elsewhere?

src/braket/circuits/parameterizable.py
speller26, line 28: Also, not sure if it's useful to return parameters if they're already bound; after all the symbol is gone, and it's not clear what the number even means.

speller26 (commented):
Looking great! Just a few more minor comments.

src/braket/circuits/circuit.py
math411: Fixed

src/braket/aws/aws_quantum_task_batch.py
math411: Fixed

src/braket/circuits/circuit.py
math411: Fixed

src/braket/circuits/parameterizable.py
math411: The parametric circuit retains the parameterized objects. The new circuit will have a copy of objects now parameterized or a copy of the same type (ex. a copy of an Rx gate) with the parameters fixed.

src/braket/circuits/circuit.py
math411: We later cast to a float within an angle. This is used for a general check to see that the value passed in is any sort of number that can be made to a float later.

src/braket/circuits/parameterizable.py
speller26: In that case, it's probably more useful to return a dict

src/braket/circuits/parameterizable.py
```diff
@@ -0,0 +1,36 @@
...
+        """
+        Returns the free parameters associated with the object.
+
+        Returns:
+            Union[FreeParameter, float]: Returns the free parameters or fixed value
+                associated with the object.
```
speller26, line 34: nit: Return docstrings are actually indented to the same level
```suggestion
            Union[FreeParameter, float]: Returns the free parameters or fixed value
            associated with the object.
```

src/braket/circuits/free_parameter.py
```diff
@@ -0,0 +1,60 @@
...
+            >>> param1 = FreeParameter("theta")
+            >>> param1 = FreeParameter("\u03B8")
+        """
+        self._name = name
+
+    @property
+    def name(self) -> str:
+        """
+        Returns:
+            str: Name of this parameter.
```
speller26, line 38: Property docstrings don't need the `Returns`
```suggestion
        str: Name of this parameter.
```

src/braket/circuits/parameterizable.py
speller26: Just a bump for this comment

src/braket/aws/aws_quantum_task.py
```diff
@@ -133,6 +133,11 @@ def create(
         )
         if tags is not None:
             create_task_kwargs.update({"tags": tags})
+        if isinstance(task_specification, Circuit) and task_specification.parameters:
+            raise ValueError(
+                f"Cannot execute parameterized circuit. "
+                f"Parameters found are: {task_specification.parameters}"
```
speller26, line 139:
```suggestion
                f"Cannot execute circuit with unbound parameters: "
                f"{task_specification.parameters}"
```

src/braket/circuits/parameterizable.py
math411: Making this abstract

test/unit_tests/braket/circuits/test_circuit.py
```diff
@@ -1306,3 +1369,159 @@ def test_diagram(h):
...
+        .ry(angle=theta, target=3)
+    )
+    expected = set()
+    expected.add(theta)
+
+    assert circ.parameters == expected
+
+
+def test_add_parameterized_instr_parameterized_circ_check_true():
+    theta = FreeParameter("theta")
```
krneta, line 1390: can you also add an alpha2 that is alpha2=FreeParameter("alpha") and add it to the circuit? What is expected for the parameters?

test/unit_tests/braket/circuits/test_circuit.py
```diff
@@ -104,6 +105,41 @@ def test_equality():
...
+    expected = (
+        Circuit()
+        .h(0)
+        .rx(angle=1, target=1)
+        .ry(angle=0, target=0)
+        .ry(angle=np.pi, target=2)
+        .rx(angle=np.pi, target=1)
+    )
+    assert new_circ == expected
+
```
krneta, line 141: what can we assert about new_circ.parameters ?

test/unit_tests/braket/circuits/test_circuit.py
```diff
@@ -104,6 +105,41 @@ def test_equality():
...
 
+def test_call():
+    alpha = FreeParameter("alpha")
+    theta = FreeParameter("theta")
+    circ = Circuit().h(0).rx(angle=theta, target=1).ry(angle=alpha, target=0)
+    circ_new = circ(theta=1, alpha=0)
+    expected = Circuit().h(0).rx(angle=1, target=1).ry(angle=0, target=0)
+
+    assert circ_new == expected
+
```
krneta, line 116: What can we assert about circ_new.parameters?

src/braket/circuits/parameterizable.py
krneta: I don't follow why it's more useful as a dict?

speller26: Mainly around the confusion of a heterogeneous list, where the order of the bound values isn't really meaningful. In fact, the order of this list isn't particularly meaningful either.

krneta: I guess it depends on the key and if the "keys" (or what's used as the keys) can be duplicated or not in this list.

test/unit_tests/braket/circuits/test_circuit.py
math411: Parameters remain unchanged. Adding in this test case.

test/unit_tests/braket/circuits/test_circuit.py
math411: Added in the assertion that this is empty.

test/unit_tests/braket/circuits/test_circuit.py
math411: Added in the assertion that this is empty.

math411: Also added a new test case for where we still have an unassigned parameter.

src/braket/circuits/circuit.py
```diff
@@ -773,6 +805,91 @@ def apply_initialization_noise(
...
+                params = instruction.operator.parameters
+                values = list(
+                    map(lambda x: param_values[str(x)] if str(x) in param_values else x, params)
+                )
+                for val in values:
+                    if not isinstance(val, FreeParameter):
+                        self._validate_parameter_value(val)
+                fixed_circ.add(
+                    Instruction(type(instruction.operator)(*values), target=instruction.target)
+                )
```
speller26, line 871: I'm a bit iffy that

```python
Instruction(type(instruction.operator)(*values), target=instruction.target)
```

relies on the order of `instruction.operator.parameters` matching that of the arguments of `instruction.operator`.

Thinking on it a bit more, I'm thinking that base parameterizable classes, like `AngledGate` and [`PauliNoise`](https://github.com/aws/amazon-braket-sdk-python/blob/main/src/braket/circuits/noise.py#L216), should have specific methods to bind parameters and create new, bound instances. With such a method, let's call it `bind_values`, this would look something like

```python
fixed_circ.add(Instruction(instruction.operator.bind_values(...)), instruction.target)
```

krneta: I like the idea that we want an explicit method, but it should probably create a new Instruction rather than mutating the existing one.

math411: Yes, this would return a new object of the type with the new values kept in. We need the existing object to remain not mutated in the circuit so that later on it can be reused.

speller26: Yep, by fixing parameters, I'm assuming a new object is created

speller26: I'm usually thinking functionally, so I'll explicitly mention mutation. In this case, it becomes a nice extension of [currying](https://en.wikipedia.org/wiki/Currying)

speller26: Now that `bind_values` exists, this entire block can be simplified:
```suggestion
                fixed_circ.add(
                    Instruction(
                        instruction.operator.bind_value(*param_values), target=instruction.target
                    )
                )
```
or something along these lines.

math411: fixed

licedric:
Can you update a description of this PR? Also make it a bit more descriptive in how to use the parametric circuit.

src/braket/circuits/circuit.py
```diff
@@ -773,6 +805,98 @@ def apply_initialization_noise(
...
+        Returns:
+            Circuit: A Circuit with specified parameters swapped for their
+            values.
+
+        """
+        fixed_circ = Circuit()
+        for instruction in self.instructions:
+            if self._check_for_params(instruction):
+                params = instruction.operator.parameters
+                bound_mapping = dict(
```
licedric, line 863: Maybe you should move this outside the loop?

math411: This is dependent on the instruction.

licedric (approved):

src/braket/circuits/circuit.py
```diff
@@ -773,6 +805,98 @@ def apply_initialization_noise(
...
+                    map(
+                        lambda x: (str(x), param_values[str(x)])
+                        if str(x) in param_values
+                        else (str(x), x),
+                        params,
+                    )
+                )
```
speller26, line 870: Why does this need to happen per operation? From what I can tell, this is what's happening:
1. `param_values` is filtered to include parameters that apply to the current operation
2. The parameters _not_ being assigned values from `param_values` are thrown in as well and mapped to their current values.
3. The resulting `bound_mapping` dict, containing every parameter on the `operation`, has its values validated and is then bound to the operation.

Why can't all the values in `parameter_values` be validated just once, and then for each operation, the entire `parameter_values` dict is bound? After all, only the relevant parameters are selected.

src/braket/circuits/circuit.py
```diff
@@ -773,6 +805,97 @@ def apply_initialization_noise(
...
+                fixed_circ.add(
+                    Instruction(
+                        instruction.operator.bind_values(**bound_mapping), target=instruction.target
+                    )
+                )
```
speller26, line 877: Since the validation is happening outside the loop, why is `bound_mapping` still necessary? Why not just
```suggestion
                fixed_circ.add(
                    Instruction(
                        instruction.operator.bind_values(**param_values), target=instruction.target
                    )
                )
```
since `bind_values` already filters `param_values` to only take the parameters in `instruction.operation`?

speller26 (approved):
Great work!

speller26 (dismissed):

src/braket/circuits/free_parameter_expression.py
```diff
@@ -0,0 +1,110 @@
...
+    def __sub__(self, other):
+        if issubclass(type(other), FreeParameterExpression):
+            return FreeParameterExpression(self.expression - other.expression)
+        else:
+            return FreeParameterExpression(self.expression - other)
+
+    def __rsub__(self, other):
+        return FreeParameterExpression(other - self.expression)
+
+    def __mul__(self, other):
```
speller26, line 89: Also worth implementing `__pow__`

speller26 (commented):
Looks good, just minor stuff

src/braket/circuits/free_parameter_expression.py
speller26: And `__neg__`

math411: Added both

src/braket/circuits/angled_gate.py
```diff
@@ -12,20 +12,29 @@
...
-class AngledGate(Gate):
+class AngledGate(Gate, Parameterizable):
     """
     Class `AngledGate` represents a quantum gate that operates on N qubits and an angle.
     """
 
-    def __init__(self, angle: float, qubit_count: Optional[int], ascii_symbols: Sequence[str]):
+    def __init__(
+        self,
+        angle: Union[FreeParameterExpression, FreeParameter, float],
```
speller26, line 30: nit: I don't think `FreeParameter` is needed here. It _is_, however, needed for the `List` type parameters, because `List` is parameter invariant.

src/braket/circuits/angled_gate.py
```diff
@@ -12,20 +12,29 @@
...
+        qubit_count: Optional[int],
+        ascii_symbols: Sequence[str],
+    ):
         """
         Args:
-            angle (float): The angle of the gate in radians.
+            angle (Union[FreeParameterExpression,FreeParameter, float]): The angle of the
+                gate in radians or expression representation.
```
speller26, line 37:
```suggestion
            angle (Union[FreeParameterExpression, float]): The angle of the gate in radians
                or expression representation.
```

src/braket/circuits/angled_gate.py
math411: Updated

speller26 (dismissed):
Awesome!

src/braket/circuits/angled_gate.py
krneta, line 30: Since FreeParameter is a child of FreeParameterExpression, you don't need both here. You can just have FreeParameterExpression and FreeParameter is implicitly valid.

src/braket/circuits/angled_gate.py
```diff
@@ -12,20 +12,29 @@
...
+    def __init__(
+        self,
+        angle: Union[FreeParameterExpression, FreeParameter, float],
+        qubit_count: Optional[int],
+        ascii_symbols: Sequence[str],
+    ):
         """
         Args:
-            angle (float): The angle of the gate in radians.
+            angle (Union[FreeParameterExpression, float]): The angle of the gate in radians
```
krneta, line 36: This should be consistent with the above. So if we decide to keep the above as is, this should be changed.

src/braket/circuits/angled_gate.py
```diff
@@ -41,22 +50,51 @@ def __init__(self, angle: float, qubit_count: Optional[int], ascii_symbols: Sequ
         super().__init__(qubit_count=qubit_count, ascii_symbols=ascii_symbols)
         if angle is None:
             raise ValueError("angle must not be None")
-        self._angle = float(angle)  # explicit casting in case angle is e.g. np.float32
+        if issubclass(type(angle), FreeParameterExpression):
```
krneta, line 53: why not isinstance instead?

src/braket/circuits/angled_gate.py
```diff
@@ -41,22 +50,51 @@ def __init__(self, angle: float, qubit_count: Optional[int], ascii_symbols: Sequ
...
+            self._parameters = [angle]
+        else:
+            self._parameters = [float(angle)]  # explicit casting in case angle is e.g. np.float32
 
     @property
-    def angle(self) -> float:
+    def parameters(self) -> List[Union[FreeParameterExpression, FreeParameter, float]]:
```
krneta, line 59: As above. I won't repeat this comment, but I believe if we decide to change it, it will be in multiple places.

src/braket/circuits/free_parameter_expression.py
```diff
@@ -0,0 +1,122 @@
...
+
+class FreeParameterExpression:
+    """
+    Class 'FreeParameterExpression'
+    """
+
+    def __init__(self, expression: Expr):
+        """
+                Initializes a FreeParameterExpression. Best practice is to initialize using
+                FreeParameters and Numbers. Not meant to be initialized directly.
```
krneta, line 28: Why is it not meant to be initialized directly?

src/braket/circuits/free_parameter_expression.py
```diff
@@ -0,0 +1,122 @@
...
+
+from sympy import Expr, sympify
+
+
+class FreeParameterExpression:
+    """
+    Class 'FreeParameterExpression'
+    """
+
+    def __init__(self, expression: Expr):
```
krneta, line 25: Doesn't exposing the sympy object here defeat the purpose of wrapping it?

src/braket/circuits/free_parameter_expression.py
```diff
@@ -0,0 +1,122 @@
...
+                    expression (Expr): The expression to use.
+
+                Examples:
+                    >>> expression_1 = FreeParameter("theta") * FreeParameter("alpha")
+                    >>> expression_2 = 1 + FreeParameter("beta") + 2 * FreeParameter("alpha")
+        ="""
+        self._expression = expression
+
+    @property
+    def expression(self) -> Expr:
```
krneta, line 40: ...and here. It seems like we have a hard coupling with sympy.

krneta: Maybe it should be private?

src/braket/circuits/gates.py
```diff
@@ -597,18 +663,33 @@ def to_ir(self, target: QubitSet):
...
+
+        Args:
+            **kwargs: The parameters that are being assigned.
+
+        Returns:
+            Gate.PhaseShift: A new Gate of the same type with the requested
+            parameters bound.
+
+        """
+        new_angle = self.angle if str(self.angle) not in kwargs else kwargs[str(self.angle)]
```
krneta, line 678: There are at least 9 repeats of this code. Can we put it in a superclass or helper function?

test/unit_tests/braket/circuits/test_free_parameter_expression.py
```diff
@@ -0,0 +1,123 @@
...
+    assert mul_1 == mul_2
+
+
+def test_add():
+    add_expr = FreeParameter("theta") + FreeParameter("alpha")
+    expected = FreeParameterExpression(FreeParameter("theta")) + FreeParameterExpression(
+        FreeParameter("alpha")
+    )
+
+    assert add_expr == expected
```
krneta, line 56: The assertions here bug me a little bit, because you're verifying that the two calls are identical (using the add operator or adding two free parameter expressions) but you're not actually testing that the "add" operation is correct.

src/braket/circuits/circuit.py
```diff
@@ -773,6 +810,88 @@ def apply_initialization_noise(
...
+        Args:
+            param_values (Dict[str, Number]): A mapping of FreeParameter names
+                to a value to assign to them.
+
+        Returns:
+            Circuit: A Circuit with specified parameters swapped for their
+            values.
+
+        """
+        fixed_circ = Circuit()
```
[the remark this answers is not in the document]
math411: L876 we copy the result type. This method should be similar to the copy method on L1124 with the exception that parameters need to be fixed which is what happens in the loops.

What do you mean when you say all the places where we copy the circuit?

src/braket/circuits/free_parameter_expression.py
math411: I think instead of making it private, I will return `repr(self._expression)` to make it less dependent on the package.

src/braket/circuits/gates.py
math411: Will add this in the next commit.

test/unit_tests/braket/circuits/test_free_parameter_expression.py
```diff
@@ -0,0 +1,123 @@
...
+#
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+
+import pytest
+
+from braket.circuits import FreeParameter, FreeParameterExpression
+
```
krneta, line 17: Can you please add constructor tests to see what are valid/invalid values? For example, can I pass in FreeParameterExpression(1+1) or FreeParameterExpression(sympy.Expr(...)), etc.

We also need to determine what we want to actually do about sympy (do we want to automatically convert or not, etc.)

src/braket/circuits/free_parameter_expression.py
speller26: Rather than initialize with a Sympy expression, we want `FreeParameterExpression`s to be built up algebraically from other `FreeParameterExpression`s, with `FreeParameter`s at the bottom.

src/braket/circuits/angled_gate.py
math411: Ack, will simplify it to just `FPE`

src/braket/circuits/angled_gate.py
speller26: Good catch; why wasn't this
```suggestion
        if isinstance(angle, FreeParameterExpression):
```
from the beginning? Seems more intuitive to write.

src/braket/circuits/circuit.py
```diff
@@ -13,12 +13,15 @@
```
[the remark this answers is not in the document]
math411: Added in a circuit test.

test/unit_tests/braket/circuits/test_free_parameter_expression.py
math411: Added constructor tests for FreeParameter. Currently, parameter expresssions can take in anything for the constructor. FreeParameterExpressions should not be constructed directly with __init__.

src/braket/circuits/free_parameter_expression.py
math411: Removing the typing for the object.

src/braket/circuits/circuit.py
```diff
@@ -773,6 +810,88 @@ def apply_initialization_noise(
...
+            param_values.
+        """
+        parameter_strings = set()
+        for parameter in self.parameters:
+            parameter_strings.add(str(parameter))
+        for param in parameter_values:
+            if param not in parameter_strings:
+                raise ValueError(f"No parameter in the circuit named: {param}")
+
+    def _use_parameter_value(self, param_values: Dict[str, Number]) -> Circuit:
```
krneta, line 851: I can't find where this is used.

krneta (dismissed):

test/unit_tests/braket/circuits/test_free_parameter_expression.py
math411: Added coverage of the object to be only a FreeParameterExpression, Number, Expr

src/braket/circuits/free_parameter.py
```diff
@@ -0,0 +1,76 @@
...
+        return self._name.name
+
+    def subs(self, parameter_values):
+        """
+        Substitutes a value in if the parameter exists within the mapping.
+
+        Args:
+            parameter_values: A mapping of parameter to its corresponding value.
+
+        Returns: The substituted value otherwise returns the object.
```
speller26, line 53: Needs type hints. Also, the `Returns` docstring should read something like
> The substituted value if this parameter is in parameter_values, otherwise returns self

src/braket/circuits/free_parameter_expression.py
```diff
@@ -0,0 +1,122 @@
...
+        """
+        Similar to a substitution in Sympy. Parameters are swapped for corresponding values or
+        expressions from the dictionary.
+
+        Args:
+            parameter_values: A mapping of parameters to their corresponding values to
+            be assigned.
+
+        Returns: A numerical value if there are no symbols left in the expression otherwise
+        returns a new FreeParameterExpression.
```
speller26, line 63: Add type hints and fix docstring spacing

speller26 (commented):
minor comments

src/braket/circuits/free_parameter.py
```diff
@@ -0,0 +1,79 @@
...
+        super().__init__(expression=self._name)
+
+    @property
+    def name(self) -> str:
+        """
+        str: Name of this parameter.
+        """
+        return self._name.name
+
+    def subs(self, parameter_values: Number):
```
speller26, line 48: `Number` seems to  be at odds with the actual type, which acts like `Dict[str, Number]`

src/braket/circuits/free_parameter.py
```diff
@@ -0,0 +1,79 @@
...
+from numbers import Number
+
+from sympy import Symbol
+
+from braket.circuits.free_parameter_expression import FreeParameterExpression
+
+
+class FreeParameter(FreeParameterExpression):
+    """
+    Class 'FreeParameter'
```
speller26, line 23: More description here

src/braket/circuits/free_parameter_expression.py
```diff
@@ -0,0 +1,129 @@
...
+        """
+                Initializes a FreeParameterExpression. Best practice is to initialize using
+                FreeParameters and Numbers. Not meant to be initialized directly.
+
+                Args:
+                    expression: The expression to use.
+
+                Examples:
+                    >>> expression_1 = FreeParameter("theta") * FreeParameter("alpha")
+                    >>> expression_2 = 1 + FreeParameter("beta") + 2 * FreeParameter("alpha")
```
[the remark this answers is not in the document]
math411: Fixed

src/braket/circuits/free_parameter_expression.py
```diff
@@ -0,0 +1,129 @@
...
+
+    @property
+    def expression(self):
+        """
+        Returns:
+            The expression for the FreeParameterExpression.
+        """
+        return self._expression
+
+    def subs(self, parameter_values: Dict):
```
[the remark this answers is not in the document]
math411: Fixed

src/braket/circuits/free_parameter.py
math411: Added

src/braket/circuits/free_parameter_expression.py
```diff
@@ -0,0 +1,129 @@
...
+
+from numbers import Number
+from typing import Dict
+
+from sympy import Expr, sympify
+
+
+class FreeParameterExpression:
+    """
+    Class 'FreeParameterExpression'
```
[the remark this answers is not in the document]
math411: Added

krneta (dismissed):

src/braket/circuits/free_parameter.py
speller26: I mean to put the description here, in the class-level docstring rather than `__init__`. Same goes for FreeParameterExpression.

speller26 (approved):

PR #309, opened 2022-03-04
feature: support local simulators for jobs

ajberdy:
*Issue #, if available:*

*Description of changes:*

Support device arg for job creation of the form "local:provider.device.name".

*Testing done:*

Tox, added unit tests, ran a job and verified AMZN_BRAKET_DEVICE_ARN is as expected

## Merge Checklist

#### General

#### Tests


src/braket/aws/aws_quantum_job.py
```diff
@@ -531,6 +535,8 @@ def __hash__(self) -> int:
     @staticmethod
     def _initialize_session(session_value, device, logger):
         aws_session = session_value or AwsSession()
+        if re.match("^local:[a-zA-Z0-9-.]+.[a-zA-Z0-9-.]+$", device):
```
kshitijc, line 538: This regex would also match something like `local:.....`. Is that acceptable?

test/unit_tests/braket/aws/test_aws_quantum_job.py
```diff
@@ -985,3 +991,26 @@ def test_exceptions_in_all_device_regions(aws_session):
...
+        "local:provider.device",
+        "local:provider.de-vice.name",
+    ),
+)
+@patch("braket.aws.aws_quantum_job.AwsSession")
+def test_initialize_session_local_device(mock_new_session, aws_session, device_str):
+    """
+    aws_session = session_value or AwsSession()
+    if re.match('^local:[a-zA-Z0-9-.]+.[a-zA-Z0-9-.]+$', device):
+        return aws_session
```
kshitijc, line 1010: Is this an intentional comment ?

src/braket/aws/aws_quantum_job.py
```diff
@@ -531,6 +535,9 @@ def __hash__(self) -> int:
     @staticmethod
     def _initialize_session(session_value, device, logger):
         aws_session = session_value or AwsSession()
+        local_device_pattern = re.compile(r"^local:[a-zA-Z0-9-_]+\.([a-zA-Z0-9-_]+\.)*[a-zA-Z0-9-_]+$")
```
kshitijc, line 538: * Does this match what the service has ? How would we ensure that it stays in sync with the service ? 
* Do we really need the entire regex match or is it sufficient for us to detect `device.startswith("local")` ? If we intend for this to be a validation (and we should think through if this adds value), what's the experience if I supply an invalid local arn (say `local:default`)

ajberdy: That makes sense, since validation happens on the service side and this is just for region switching logic.

src/braket/aws/aws_quantum_job.py
```diff
@@ -84,8 +85,11 @@ def create(
         """Creates a job by invoking the Braket CreateJob API.
 
         Args:
-            device (str): ARN for the AWS device which is primarily
-                accessed for the execution of this job.
+            device (str): ARN for the AWS device which is primarily accessed for the execution
+                of this job. Alternatively, a string of the format
+                "local:<provider>.<simulator>.<name>" for using a local simulator for the job.
```
krneta, line 90: Without being able to be more specific for the "provider", "simulator", and "name", I'm not sure what use we get out of having the format. It seems like we'll just confuse customers and (since it's not standardized) customers may just send in "a.a.a". 

What about making this the same as our LocalSimulator "name" parameter:
The name of the simulator backend or
                the actual simulator instance to use for simulation. Defaults to the
                `default` simulator backend name.

krneta: Or similar, if not the same.

krneta: I see you updated the check - we should also update the docs.

ajberdy: Format and docs updated to match the decision to use `"local:provider/simulator"` convention

kshitijc (approved):

ajberdy:
Tested e2e:

```
def run_job():
    print(os.environ.get("AMZN_BRAKET_DEVICE_ARN"))
    device = LocalSimulator()

    bell = Circuit().h(0).cnot(0, 1)
    num_tasks = 10
    results = []

    for i in range(num_tasks):
        task = device.run(bell, shots=100)
        result = task.result().measurement_counts
        results.append(result)
        print(f"iter {i}: {result}")

    save_job_result({"results": results})


if __name__ == "__main__":
    job = AwsQuantumJob.create(
        device="local:provider/simulator",
        source_module="job.py",
        entry_point="job:run_job",
        wait_until_complete=True,
    )
    print(job.result())
```

kshitijc (approved):

math411 (approved):

virajvchaudhari (approved):

PR #311, opened 2022-03-10
feature: Gate and Circuit inversion

speller26:
Introduces the ability to take adjoints of Gates, Instructions and Circuits.

*Issue #, if available:*

*Description of changes:*

*Testing done:*

## Merge Checklist

#### General

#### Tests


src/braket/circuits/compiler_directive.py
```diff
@@ -41,6 +43,21 @@ def ascii_symbols(self) -> Tuple[str, ...]:
         """Tuple[str, ...]: Returns the ascii symbols for the compiler directive."""
         return self._ascii_symbols
 
+    def opposite(self) -> CompilerDirective:
```
math411, line 46: Maybe opposite_directive? `opposite` seems a bit too general of a function name.

src/braket/circuits/instruction.py
```diff
@@ -81,6 +83,24 @@ def target(self) -> QubitSet:
...
+    def adjoint(self) -> Instruction:
+        """Returns an Instruction with the adjoint of this instruction's own operator
+
+        This operation only works on Gate operators.
+
+        Returns:
+            Instruction: A new instruction whose operator is the adjoint of `operator`
+
+        Raises:
+            TypeError: If `operator` is not of type `Gate`
```
math411, line 95:
```suggestion
            NotImplementedError: If `operator` is not of type `Gate` or `CompilerDirective`
```

test/unit_tests/braket/circuits/test_angled_gate.py
```diff
@@ -50,6 +50,12 @@ def test_angle_setter(angled_gate):
     angled_gate.angle = 0.14
 
 
+def test_adjoint_list(angled_gate):
```
math411, line 53: Can you add a test with an expression included?
```suggestion
def test_adjoint_list_parameterized():
    theta = FreeParameter("theta")
    AngledGate(angle=(theta + 1), qubit_count=1, ascii_symbols=["foo"])
    assert angled_gate.adjoint_list() == [
        AngledGate(angle=-(theta + 1), qubit_count=1, ascii_symbols=["foo"])
    ]
```

math411: This would be in addition to the current test. The above is not something to commit

src/braket/circuits/circuit.py
```diff
@@ -1030,6 +1029,12 @@ def _flatten(addable):
 
         return self
 
+    def adjoint(self) -> Circuit:
+        circ = Circuit([instr.adjoint() for instr in reversed(self.instructions)])
+        for result_type in self._result_types:
+            circ.add_result_type(result_type)
```
krneta, line 1035: I always worry that there's data we're forgetting to copy over. For example, do we need to copy over the parameters?

src/braket/circuits/compiler_directive.py
```diff
@@ -41,6 +43,21 @@ def ascii_symbols(self) -> Tuple[str, ...]:
...
+        """Returns the "opposite" counterpart to this compiler directive.
+
+        For example, the opposite of a directive that starts a box
+        is the directive that ends the box.
+
+        TODO: Better name for this? Dual?
```
krneta, line 52: How about "counterpart"?

src/braket/circuits/gate.py
```diff
@@ -40,6 +43,27 @@ def __init__(self, qubit_count: Optional[int], ascii_symbols: Sequence[str]):
...
+        """Returns the adjoint of this gate.
+
+        Returns:
+            Gate: The adjoint of this gate.
+        """
+        new = copy.copy(self)
+        new._adjoint = not self._adjoint
+        return new
+
+    def adjoint_list(self) -> List[Gate]:
```
krneta, line 58: Isn't it weird to have both adjoint and adjoint_list? When would we know which one to use? Why not just always use adjoint_list?

src/braket/circuits/gate.py
```diff
@@ -40,6 +43,27 @@ def __init__(self, qubit_count: Optional[int], ascii_symbols: Sequence[str]):
...
+        """Returns a list of gates that comprise the adjoint of this gate.
+
+        For compatibility, it is best if all gates in ths list are of the same type as this gate.
```
krneta, line 61: nit: typo

src/braket/circuits/gate.py
```diff
@@ -40,6 +43,27 @@ def __init__(self, qubit_count: Optional[int], ascii_symbols: Sequence[str]):
...
+
+        Returns:
+            List[Gate]: The gates comprising the adjoint of this gate.
+        """
+        raise NotImplementedError(f"Gate {self.name} does not have adjoint implemented")
```
krneta, line 66: I'm confused why this isn't:
`` return [self.adjoint()]

src/braket/circuits/gate.py
```diff
@@ -49,18 +73,35 @@ def to_ir(self, target: QubitSet) -> Any:
         Returns:
             IR object of the quantum operator and target
         """
+        if self._adjoint:
+            # Use adjoint list of original gate
+            return [elem.to_ir(target) for elem in self.adjoint().adjoint_list()]
```
krneta, line 78: `.adjoin().adjoint_list()` doesn't make sense to me from the description of these functions.

src/braket/circuits/gate.py
```diff
@@ -49,18 +73,35 @@ def to_ir(self, target: QubitSet) -> Any:
...
+
+    def _to_ir(self, target: QubitSet) -> Any:
         raise NotImplementedError("to_ir has not been implemented yet.")
 
+    @property
+    def ascii_symbols(self) -> Tuple[str, ...]:
+        """Tuple[str, ...]: Returns the ascii symbols for the quantum operator."""
+        if self._adjoint:
+            return tuple(
+                f"({symbol})^†" if symbol != "C" else symbol for symbol in self._ascii_symbols
```
krneta, line 89: What's the special case here with "C", and why do we think we won't have more?

src/braket/circuits/gates.py
```diff
@@ -39,16 +40,31 @@
...
 
-class H(Gate):
+class H(_HermitianGate):
     """Hadamard gate."""
 
     def __init__(self):
         super().__init__(qubit_count=None, ascii_symbols=["H"])
 
-    def to_ir(self, target: QubitSet):
+    def _to_ir(self, target: QubitSet):
```
krneta, line 67: Can we just change all these to private? Is that a breaking change?

src/braket/circuits/gates.py
```diff
@@ -39,16 +40,31 @@
     3. Register the class with the `Gate` class via `Gate.register_gate()`.
 """
 
+
+class _FiniteOrderGate(Gate, ABC):
```
krneta, line 44: Missing docs. 

Why are these new superclasses needed?

src/braket/circuits/instruction.py
```diff
@@ -81,6 +83,24 @@ def target(self) -> QubitSet:
         """
         return self._target
 
+    def adjoint(self) -> Instruction:
+        """Returns an Instruction with the adjoint of this instruction's own operator
+
+        This operation only works on Gate operators.
```
krneta, line 89: Is this true? You seem to have a switch statement for CompilerDirective.

src/braket/circuits/instruction.py
```diff
@@ -81,6 +83,24 @@ def target(self) -> QubitSet:
...
+
+        Returns:
+            Instruction: A new instruction whose operator is the adjoint of `operator`
+
+        Raises:
+            NotImplementedError: If `operator` is not of type `Gate` or `CompilerDirective`
```
krneta, line 95: Will this be a problem for unitary gates? Will it cause a minefield?

src/braket/circuits/instruction.py
```diff
@@ -81,6 +83,24 @@ def target(self) -> QubitSet:
...
+        """
+        operator = self._operator
+        if isinstance(operator, Gate):
+            return Instruction(operator.adjoint(), self._target)
+        elif isinstance(operator, CompilerDirective):
```
krneta, line 100: I'm not sure that this makes sense generally.

test/unit_tests/braket/circuits/test_ascii_circuit_diagram.py
```diff
@@ -442,8 +417,136 @@ def test_verbatim_qubset_qubits():
...
+    _assert_correct_diagram(circ, expected)
+
+
+def test_adjoint_1q():
+    circ = Circuit().s(0).rz(0, 3.14)
+    # Column formats to length of the gate plus the ascii representation for the angle.
+    expected = (
+        "T  : |     0      |  1  |",
+        "                         ",
+        "q0 : -(Rz(3.14))^†-(S)^†-",
```
krneta, line 429: Rather than two characters it might look nicer if we just use an asterisk instead?

speller26: While an asterisk is standard for adjoints in math, in physics it just means the complex conjugate, which is not what we want. I'm fine with just having the dagger without the ^ though

src/braket/circuits/instruction.py
speller26: How so? Once if we start having more types, I suppose we can do a `singledispatch` like we do for translators.

src/braket/circuits/instruction.py
speller26: Nope, unitaries have well-defined adjoints (and adjoint lists)

src/braket/circuits/instruction.py
speller26: Whoops, should be "and CompilerDirectives"

src/braket/circuits/gates.py
speller26: I like having these because many gates fall into this category. I think it's much better than pasting
```python
def adjoint_list(self):
    return [self]
```
for all involutory gates, and 
```python
def adjoint_list(self):
    return [self, self, self]  # or more, etc
```
for other orders. As well, it helps prevent stop a developer from missing or incorrectly implementing these methods for new gates in the same category.

src/braket/circuits/gates.py
speller26: Won't be breaking because they all inherit from `Gate`, where the public `to_ir` is defined.

src/braket/circuits/gate.py
speller26: C stands for control (I would've preferred •, but it's a bit late for that). There may be more special characters down the line, but we can deal with that when it happens. I'll add a comment.

src/braket/circuits/gate.py
speller26: `adjoint_list` will often contain copies of `self`, and we don't want to recursively call `to_ir` on self because it'll never end.

src/braket/circuits/gate.py
speller26: All `adjoint` does is mark a gate as an adjoint; it doesn't have anything to do with the actual implementation of adjoints. `adjoint_list`, on the other had, is a definition of the adjoint. In principle, `adjoint_list` isn't necessary at all; the adjoint implementation could be handled by the device, but we currently have no way of communicating that.

src/braket/circuits/gate.py
speller26: See comment about difference between `adjoint` and `adjoint_list`

src/braket/circuits/compiler_directive.py
speller26: I like it!

src/braket/circuits/circuit.py
speller26: Nope, that's included in the gates. Right now, there are only two things you can add to a circuit: instructions and result types (and other circuits, which are made up of these)

src/braket/circuits/gate.py
krneta: Could we differentiate the names more to make this difference more obvious?

src/braket/circuits/angled_gate.py
```diff
@@ -96,4 +107,5 @@ def __eq__(self, other):
         return False
 
     def __repr__(self):
-        return f"{self.name}('angle': {self.angle}, 'qubit_count': {self.qubit_count})"
+        dagger_str = "^†" if self._adjoint else ""
```
krneta, line 110: You missed this ^†

src/braket/circuits/gate.py
```diff
@@ -49,18 +73,38 @@ def to_ir(self, target: QubitSet) -> Any:
...
+
+    def _to_ir(self, target: QubitSet) -> Any:
         raise NotImplementedError("to_ir has not been implemented yet.")
 
+    @property
+    def ascii_symbols(self) -> Tuple[str, ...]:
+        """Tuple[str, ...]: Returns the ascii symbols for the quantum operator."""
+        if self._adjoint:
+            return tuple(
+                # "C" stands for control, and doesn't need a dagger symbol
```
krneta, line 90: But the ascii symbols are determined by subclasses, it seems like this is information leaking to know what one of the subclass ascii symbols would be.

krneta: Why can't the subclass just override this method?

src/braket/circuits/instruction.py
krneta: It just seems like not all CompilerDirectives will have counterparts. But (I think) I get your point that we can introduce a new class that distinguishes, if we need to.

src/braket/circuits/angled_gate.py
krneta: How come it wasn't caught by unit tests?

speller26: Ah, coverage doesn't seem to cover branches within an expression. I'll fix this.

src/braket/circuits/instruction.py
speller26: Yep. In a bit more detail, all `CompilerDirective`s need to be in some `Instruction` (and compilation-relate construct that _doesn't_ go in an `Instruction` would need a new class), and `counterpart` makes sense generally for anything in an instruction; it doesn't need to be symmetric, with an open to a close like with verbatim boxes, but anything that can be reversed in some way.

src/braket/circuits/gate.py
speller26: I want to avoid having to copy the `if` statement for every subclass, as it's easy for a developer to forget to add the branch,

src/braket/circuits/gate.py
speller26: `adjoint_expansion`?

src/braket/circuits/gate.py
speller26: I think the best way to go about this is to throw an exception if `_adjoint` is `True` for any gate in `adjoint_list`, and include in the `adjoint_list` doc that the adjoint list _must_ consist only of non-adjoint gates.

src/braket/circuits/gates.py
```diff
@@ -79,11 +107,11 @@ def h(target: QubitSetInput) -> Iterable[Instruction]:
 Gate.register_gate(H)
 
 
-class I(Gate):  # noqa: E742, E261
+class I(_FiniteOrderGate):  # noqa: E742, E261
```
kshitijc, line 110: Isn't `I` hermitian ?

speller26: `I` has order 1, which implies Hermiticity, but here we do one better by simply not adding a gate at all, as opposed to a single gate. I don't have a problem with making it explicitly Hermitian.

src/braket/circuits/angled_gate.py
```diff
@@ -87,6 +88,16 @@ def bind_values(self, **kwargs):
         """
         raise NotImplementedError
 
+    def adjoint(self) -> List[Gate]:
```
licedric, line 91: This is true for all of our current gates but might not be in the future. Is there a way to catch it?

src/braket/circuits/gates.py
licedric: I suggest we remove both FiniteOrderGate and HermitianGate. It's probably more confusing for developers to use instead of just defining the adjoint operator themselves.

src/braket/circuits/angled_gate.py
speller26: This is the default behavior, and future gates could just override it.

src/braket/circuits/gates.py
speller26: I'm fine with that, as there aren't too many instances of these yet.

src/braket/circuits/gates.py
licedric: It's this kind of discussion that makes me not want to have _HermitianGate in the first place; I don't think it adds too much value and just makes development slower.

licedric (approved):

PR #314, opened 2022-03-14
feature: support multiple instances for jobs with a local simulator

ajberdy:
*Issue #, if available:*

*Description of changes:*

Support multiple instances with local simulator jobs. PR is to merge into already-approved, but unmerged `local-sim-jobs` branch. This lets the diff only show changes on top of that commit.

*Testing done:*

Unit tests. Pending e2e test once service side changes support `instanceConfg.instanceCount`

## Merge Checklist

#### General

#### Tests

### DIFF
--- a/src/braket/jobs/config.py
+++ b/src/braket/jobs/config.py
@@ -29,6 +29,7 @@ class InstanceConfig:
 
     instanceType: str = "ml.m5.large"
     volumeSizeInGb: int = 30
+    instanceCount: int = 1
 
 
 @dataclass
--- a/src/braket/jobs/quantum_job_creation.py
+++ b/src/braket/jobs/quantum_job_creation.py
@@ -139,6 +139,9 @@ def prepare_quantum_job(
     _validate_params(param_datatype_map)
     aws_session = aws_session or AwsSession()
     device_config = DeviceConfig(device)
+
+    if instance_config and instance_config.instanceCount > 1 and not device.startswith("local:"):
+        raise ValueError("Multiple instances only supported for jobs with a 'local' simulator.")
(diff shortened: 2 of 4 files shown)


src/braket/jobs/quantum_job_creation.py
```diff
@@ -139,6 +139,9 @@ def prepare_quantum_job(
     _validate_params(param_datatype_map)
     aws_session = aws_session or AwsSession()
     device_config = DeviceConfig(device)
+
+    if instance_config and instance_config.instanceCount > 1 and not device.startswith("local:"):
```
christianbmadsen, line 143: Do we need an upper limit?

christianbmadsen: I agree. Let's move to the service and simply return the API validation results.

christianbmadsen (approved):

src/braket/jobs/config.py
```diff
@@ -29,6 +29,7 @@ class InstanceConfig:
```
math411, line 29: Do we want to regionalize this config? In the event we have an issue with m5.Ls in one region, this would allow a quick change without affecting other regions.

math411: Instead of including each region, we can have a default value that can be overridden for regions. This way, we don't explicitly need to write out each region's value and can change only when it is needed to be specified outside the default.

ajberdy: Resolved offline: conclusion was that these configs are typically used for a single job, which only exists in one region

ajberdy:
e2e tested, works as expected. Merging into local-sim-jobs

PR #322, opened 2022-04-06
feat: Add support for generating OpenQASM from circuits

kshitijc:
*Issue #, if available:*

*Description of changes:*
* Adds the ability to generate OpenQASM from Braket circuits. 
* The implementation doesn't cover every supported gate/result type/noise yet. It only covers the following components:
  * Compiler Directives
  * Circuit
  * Gates (Only `Rx`)
  * Instruction
  * Noise (Only `BitFlip`)
  * Observables (Only `I` and `Hermitian`)
  * Result Types (Only `Expectation`)

*Testing done:*
* `tox -e unit-tests` succeeded.
* Validated that the generated circuits can be executed against the service.
```
rom braket.circuits import Circuit
from braket.circuits.serialization import IRType
from braket.aws import AwsDevice

c = Circuit().rx(0, 0.15).rx(7, 0.3)
r = c.to_ir(ir_type=IRType.OPENQASM)
print(r.source)
d = AwsDevice("arn:aws:braket:::device/quantum-simulator/amazon/sv1")
res = d.run(r, shots=10).result()
print(res)
```

## Merge Checklist

#### General

#### Tests

### DIFF
--- a/src/braket/circuits/circuit.py
+++ b/src/braket/circuits/circuit.py
@@ -1104,10 +1104,7 @@ class Circuit:
     def _to_openqasm(
         self, serialization_properties: OpenQASMSerializationProperties
     ) -> OpenQasmProgram:
-        ir_instructions, qubit_reference_format = self._create_openqasm_header(
-            serialization_properties
-        )
-        serialization_properties.qubit_reference_format = qubit_reference_format
+        ir_instructions = self._create_openqasm_header(serialization_properties)
         openqasm_ir_type = IRType.OPENQASM
         ir_instructions.extend(
             [
@@ -1129,14 +1126,14 @@ class Circuit:
             )
         else:
             for idx, qubit in enumerate(self.qubits):
-                qubit_target = qubit_reference_format.format(int(qubit))
+                qubit_target = serialization_properties.format_target(int(qubit))
                 ir_instructions.append(f"b[{idx}] = measure {qubit_target};")
 
         return OpenQasmProgram.construct(source="\n".join(ir_instructions))
 
     def _create_openqasm_header(
         self, serialization_properties: OpenQASMSerializationProperties
-    ) -> Tuple[List[str], str]:
+    ) -> List[str]:
         ir_instructions = ["OPENQASM 3.0;"]
         if not self.result_types:
             ir_instructions.append(f"bit[{self.qubit_count}] b;")
@@ -1144,15 +1141,12 @@ class Circuit:
         if serialization_properties.qubit_reference_type == QubitReferenceType.VIRTUAL:
             total_qubits = max(self.qubits).real + 1
             ir_instructions.append(f"qubit[{total_qubits}] q;")
-            qubit_reference_format = "q[{}]"
-        elif serialization_properties.qubit_reference_type == QubitReferenceType.PHYSICAL:
-            qubit_reference_format = "${}"
-        else:
+        elif serialization_properties.qubit_reference_type != QubitReferenceType.PHYSICAL:
             raise ValueError(
                 f"Invalid qubit_reference_type "
                 f"{serialization_properties.qubit_reference_type} supplied."
             )
-        return ir_instructions, qubit_reference_format
+        return ir_instructions
 
     def as_unitary(self) -> np.ndarray:
         """
--- a/src/braket/circuits/circuit.py
+++ b/src/braket/circuits/circuit.py
@@ -40,8 +40,10 @@ from braket.circuits.parameterizable import Parameterizable
 from braket.circuits.qubit import QubitInput
 from braket.circuits.qubit_set import QubitSet, QubitSetInput
 from braket.circuits.result_type import ObservableResultType, ResultType
+from braket.circuits.serialization import IRType, QubitReferenceType
 from braket.circuits.unitary_calculation import calculate_unitary
-from braket.ir.jaqcd import Program
+from braket.ir.jaqcd import Program as JaqcdProgram
+from braket.ir.openqasm import Program as OpenQasmProgram
 
 SubroutineReturn = TypeVar(
     "SubroutineReturn", Iterable[Instruction], Instruction, ResultType, Iterable[ResultType]
@@ -1043,24 +1045,85 @@ class Circuit:
         """
         return circuit_diagram_class.build_diagram(self)
 
-    def to_ir(self) -> Program:
+    def to_ir(
+        self,
+        ir_type: IRType = IRType.JAQCD,
+        qubit_reference_type: QubitReferenceType = QubitReferenceType.VIRTUAL,
+    ) -> Union[OpenQasmProgram, JaqcdProgram]:
         """
         Converts the circuit into the canonical intermediate representation.
         If the circuit is sent over the wire, this method is called before it is sent.
 
+        Args:
+            ir_type (IRType): The IRType to use for converting the circuit object to its
+                IR representation.
+            qubit_reference_type (QubitReferenceType): The qubit reference type to use for
+                the IR representation.
+
         Returns:
-            (Program): An AWS quantum circuit description program in JSON format.
+            (Union[OpenQasmProgram, JaqcdProgram]): A representation of the circuit in the
+            `ir_type` format.
+
+        Raises:
+            ValueError: If the supplied `ir_type` is not supported.
         """
-        ir_instructions = [instr.to_ir() for instr in self.instructions]
-        ir_results = [result_type.to_ir() for result_type in self.result_types]
+        if ir_type == IRType.JAQCD:
+            return self._to_jaqcd()
+        elif ir_type == IRType.OPENQASM:
+            return self._to_openqasm(qubit_reference_type)
+        else:
+            raise ValueError(f"Supplied ir_type {ir_type} is not supported.")
+
+    def _to_jaqcd(self) -> JaqcdProgram:
+        ir_instructions = [instr.to_ir(ir_type=IRType.JAQCD) for instr in self.instructions]
+        ir_results = [result_type.to_ir(ir_type=IRType.JAQCD) for result_type in self.result_types]
         ir_basis_rotation_instructions = [
-            instr.to_ir() for instr in self.basis_rotation_instructions
+            instr.to_ir(ir_type=IRType.JAQCD) for instr in self.basis_rotation_instructions
         ]
-        return Program.construct(
+        return JaqcdProgram.construct(
             instructions=ir_instructions,
             results=ir_results,
             basis_rotation_instructions=ir_basis_rotation_instructions,
         )
+
+    def _to_openqasm(self, qubit_reference_type: QubitReferenceType) -> OpenQasmProgram:
+        ir_instructions = ["OPENQASM 3.0;"]
+        if not self.result_types:
+            ir_instructions.append(f"bit[{self.qubit_count}] b;")
+
+        if qubit_reference_type == QubitReferenceType.VIRTUAL:
+            total_qubits = max(self.qubits).real + 1
+            ir_instructions.append(f"qubit[{total_qubits}] q;")
+            qubit_reference_format = "q[{}]"
+        elif qubit_reference_type == QubitReferenceType.PHYSICAL:
+            qubit_reference_format = "${}"
+        else:
+            raise ValueError(f"Invalid qubit_reference_type {qubit_reference_type} supplied.")
+
+        ir_instructions.extend(
+            [
+                instruction.to_ir(
+                    ir_type=IRType.OPENQASM, qubit_reference_format=qubit_reference_format
+                )
+                for instruction in self.instructions
+            ]
+        )
+
+        if self.result_types:
+            ir_instructions.extend(
+                [
+                    result_type.to_ir(
+                        ir_type=IRType.OPENQASM, qubit_reference_format=qubit_reference_format
+                    )
+                    for result_type in self.result_types
+                ]
+            )
+        else:
+            for idx, qubit in enumerate(self.qubits):
+                qubit_target = qubit_reference_format.format(int(qubit))
+                ir_instructions.append(f"b[{idx}] = measure {qubit_target};")
+
+        return OpenQasmProgram.construct(source="\n".join(ir_instructions))
 
     def as_unitary(self) -> np.ndarray:
         """
--- a/src/braket/circuits/circuit.py
+++ b/src/braket/circuits/circuit.py
@@ -40,7 +40,12 @@ from braket.circuits.parameterizable import Parameterizable
 from braket.circuits.qubit import QubitInput
 from braket.circuits.qubit_set import QubitSet, QubitSetInput
 from braket.circuits.result_type import ObservableResultType, ResultType
-from braket.circuits.serialization import IRType, QubitReferenceType
+from braket.circuits.serialization import (
+    IRType,
+    OpenQASMSerializationProperties,
+    QubitReferenceType,
+    SerializationProperties,
+)
 from braket.circuits.unitary_calculation import calculate_unitary
 from braket.ir.jaqcd import Program as JaqcdProgram
 from braket.ir.openqasm import Program as OpenQasmProgram
@@ -1048,7 +1053,7 @@ class Circuit:
     def to_ir(
         self,
         ir_type: IRType = IRType.JAQCD,
-        qubit_reference_type: QubitReferenceType = QubitReferenceType.VIRTUAL,
+        serialization_properties: SerializationProperties = None,
     ) -> Union[OpenQasmProgram, JaqcdProgram]:
         """
         Converts the circuit into the canonical intermediate representation.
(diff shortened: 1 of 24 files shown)


src/braket/circuits/gate.py
```diff
@@ -41,15 +42,52 @@ def __init__(self, qubit_count: Optional[int], ascii_symbols: Sequence[str]):
...
         super().__init__(qubit_count=qubit_count, ascii_symbols=ascii_symbols)
 
-    def to_ir(self, target: QubitSet) -> Any:
+    def to_ir(
+        self, target: QubitSet, ir_type: IRType = IRType.JAQCD, qubit_reference_format: str = "${}"
+    ) -> Any:
         """Returns IR object of quantum operator and target
 
         Args:
             target (QubitSet): target qubit(s)
```
math411, line 51:
```suggestion
            target (QubitSet): target qubit(s).
```

src/braket/circuits/noises.py
```diff
@@ -79,9 +79,13 @@ def __init__(self, probability: float):
             ascii_symbols=["BF({:.2g})".format(probability)],
         )
 
-    def to_ir(self, target: QubitSet):
+    def to_jaqcd(self, target: QubitSet):
         return ir.BitFlip.construct(target=target[0], probability=self.probability)
 
+    def to_openqasm(self, target: QubitSet, qubit_reference_format: str):
+        target_qubit = qubit_reference_format.format(int(target[0]))
```
math411, line 86: Instead of hardcoding the target at `0` can we use loop through the list using `fixed_qubit_count`?

src/braket/circuits/observables.py
```diff
@@ -63,9 +64,16 @@ def __init__(self):
         """
         super().__init__(qubit_count=1, ascii_symbols=["I"])
 
-    def to_ir(self) -> List[str]:
+    def to_jaqcd(self) -> List[str]:
         return ["i"]
 
+    def to_openqasm(self, qubit_reference_format: str, target: QubitSet = None) -> str:
+        if target:
+            qubit_target = qubit_reference_format.format(int(target[0]))
```
math411, line 72: Similarly using `qubit_count` here instead of target[0] to loop through the target(s).

math411: In  this case, just a singular target but it may be useful for multiqubit situations.

src/braket/circuits/noises.py
kshitijc: Yes, it'd be nice to avoid hard coding values. However, the generated string requires a single element instead of a list and the indexing here makes it clear what's being used.

src/braket/circuits/observables.py
kshitijc: Yes, it'd be nice to avoid hard coding values. However, the generated string requires a single element instead of a list and the indexing here makes it clear what's being used.

src/braket/circuits/circuit.py
```diff
@@ -1043,25 +1045,86 @@ def diagram(self, circuit_diagram_class=AsciiCircuitDiagram) -> str:
...
+        qubit_reference_type: QubitReferenceType = QubitReferenceType.VIRTUAL,
+    ) -> Union[OpenQasmProgram, JaqcdProgram]:
         """
         Converts the circuit into the canonical intermediate representation.
         If the circuit is sent over the wire, this method is called before it is sent.
 
+        Args:
+            ir_type (IRType): The IRType to use for converting the circuit object to its
+                IR representation.
+            qubit_reference_type (QubitReferenceType): The qubit reference type to use for
```
krneta, line 1060: Can we have more documentation that explains what this actually does/represents? I wouldn't expect customers to immediately understand the difference between virtual and physical representation of qubits in an IR.

src/braket/circuits/circuit.py
```diff
@@ -1043,25 +1045,86 @@ def diagram(self, circuit_diagram_class=AsciiCircuitDiagram) -> str:
...
-        ir_results = [result_type.to_ir() for result_type in self.result_types]
+        if ir_type == IRType.JAQCD:
+            return self._to_jaqcd()
+        elif ir_type == IRType.OPENQASM:
+            return self._to_openqasm(qubit_reference_type)
+        else:
+            raise ValueError(f"Supplied ir_type {ir_type} is not supported.")
+
+    def _to_jaqcd(self) -> JaqcdProgram:
+        ir_instructions = [instr.to_ir(ir_type=IRType.JAQCD) for instr in self.instructions]
```
krneta, line 1078: Might be a little easier to read if you define ir_type=IRTYpe.JAQCD here, and pass it in to all the other places where you pass the type (and so we don't accidentally make a mistake when adding another line).

src/braket/circuits/circuit.py
```diff
@@ -1043,25 +1045,86 @@ def diagram(self, circuit_diagram_class=AsciiCircuitDiagram) -> str:
...
         ]
-        return Program.construct(
+        return JaqcdProgram.construct(
             instructions=ir_instructions,
             results=ir_results,
             basis_rotation_instructions=ir_basis_rotation_instructions,
         )
 
+    def _to_openqasm(self, qubit_reference_type: QubitReferenceType) -> OpenQasmProgram:
+        ir_instructions = ["OPENQASM 3.0;"]
```
krneta, line 1090: Move 1090 to 1101 to a _create_openqasm_header function() ?

src/braket/circuits/circuit.py
```diff
@@ -1043,25 +1045,86 @@ def diagram(self, circuit_diagram_class=AsciiCircuitDiagram) -> str:
...
+            ir_instructions.append(f"qubit[{total_qubits}] q;")
+            qubit_reference_format = "q[{}]"
+        elif qubit_reference_type == QubitReferenceType.PHYSICAL:
+            qubit_reference_format = "${}"
+        else:
+            raise ValueError(f"Invalid qubit_reference_type {qubit_reference_type} supplied.")
+
+        ir_instructions.extend(
+            [
+                instruction.to_ir(
```
krneta, line 1105: Seems like we should be able to collect the IR for instructions and result types using the same function for both JAQCD and OpenQASM in the same function (by passing in the ir_type) and get rid of some of the duplicate code, and make sure we don't miss anything... except that we're not getting basis_rotation_instructions. Is that because this is a partial implementation, or are they not applicable for OpenQASM?

src/braket/circuits/compiler_directive.py
```diff
@@ -41,8 +43,43 @@ def ascii_symbols(self) -> Tuple[str, ...]:
         """Tuple[str, ...]: Returns the ascii symbols for the compiler directive."""
         return self._ascii_symbols
 
-    def to_ir(self, *args, **kwargs):
-        raise NotImplementedError("to_ir has not been implemented yet.")
+    def to_ir(
+        self,
+        target: QubitSet = None,
+        ir_type: IRType = IRType.JAQCD,
+        qubit_reference_format: str = "${}",
```
krneta, line 50: Should we pass the Virtual/Physical enum in case there's ever any other differences (besides how the qubits are formatted) - maybe for future IRs?

src/braket/circuits/compiler_directive.py
```diff
@@ -41,8 +43,43 @@ def ascii_symbols(self) -> Tuple[str, ...]:
...
+            ValueError: If the supplied `ir_type` is not supported.
+        """
+        if ir_type == IRType.JAQCD:
+            return self.to_jaqcd()
+        elif ir_type == IRType.OPENQASM:
+            return self.to_openqasm()
+        else:
+            raise ValueError(f"Supplied ir_type {ir_type} is not supported.")
+
+    def to_jaqcd(self) -> Any:
```
krneta, line 76: should these be private?

src/braket/circuits/compiler_directives.py
```diff
@@ -24,9 +24,12 @@ class StartVerbatimBox(CompilerDirective):
     def __init__(self):
         super().__init__(["StartVerbatim"])
 
-    def to_ir(self, *args, **kwargs):
+    def to_jaqcd(self, *args, **kwargs):
```
krneta, line 27: private?

src/braket/circuits/gate.py
```diff
@@ -41,15 +42,52 @@ def __init__(self, qubit_count: Optional[int], ascii_symbols: Sequence[str]):
...
         """
-        raise NotImplementedError("to_ir has not been implemented yet.")
+        if ir_type == IRType.JAQCD:
+            return self.to_jaqcd(target)
+        elif ir_type == IRType.OPENQASM:
+            return self.to_openqasm(target, qubit_reference_format)
+        else:
+            raise ValueError(f"Supplied ir_type {ir_type} is not supported.")
+
+    def to_jaqcd(self, target: QubitSet) -> Any:
```
krneta, line 66: private?

src/braket/circuits/gates.py
```diff
@@ -48,7 +48,7 @@ class H(Gate):
     def __init__(self):
         super().__init__(qubit_count=None, ascii_symbols=["H"])
 
-    def to_ir(self, target: QubitSet):
+    def to_jaqcd(self, target: QubitSet):
```
krneta, line 51: Is this a backward incompatible / breaking change?

src/braket/circuits/noises.py
krneta: I think Cody's question is valid. We can do it for each qubit in the target. Would that be wrong?

src/braket/circuits/compiler_directive.py
krneta: Alternatively we could pass an object that does the formatting of the qubit, rather than a string:

qubit_formatter = PhysicalQubitFormatter()

qubit_formatter.format(qubit)

?

Not sure if future-proof, or overly complex.

src/braket/circuits/observables.py
krneta: Sorry, why can't we return multiple lines in the strings?

src/braket/circuits/observables.py
```diff
@@ -311,11 +319,26 @@ def __init__(self, matrix: np.ndarray, display_name: str = "Hermitian"):
...
         super().__init__(qubit_count=qubit_count, ascii_symbols=[display_name] * qubit_count)
 
-    def to_ir(self) -> List[List[List[List[float]]]]:
+    def to_jaqcd(self) -> List[List[List[List[float]]]]:
         return [
             [[[element.real, element.imag] for element in row] for row in self._matrix.tolist()]
         ]
 
+    def to_openqasm(self, qubit_reference_format: str, target: QubitSet = None) -> str:
+        def _serialized_matrix():
```
krneta, line 328: I generally find functions-in-functions to be confusing. Why not move it out?

src/braket/circuits/compiler_directive.py
kshitijc: We could add the additional argument when needed. The object seems a little heavy handed for now. We could extend the current parameter to accept `Union[str, QubitFormatter]` when we see the need.

src/braket/circuits/circuit.py
kshitijc: Added docs to the Enum definition.

src/braket/circuits/circuit.py
kshitijc: ok, done.

src/braket/circuits/circuit.py
kshitijc: Done

src/braket/circuits/circuit.py
kshitijc: Basis rotation instructions are not applicable for OpenQASM. In the end there isn't much duplicated code between the two. Trying to unify the two code paths led to a more confusing implementation here.

src/braket/circuits/compiler_directive.py
kshitijc: This is something I wasn't sure about. It would be nice to have them private, but then having an interface define private methods that the subclasses **need** to implement felt weird. I can change to private if you think that's fine.

src/braket/circuits/compiler_directives.py
kshitijc: See reply to the previous comment

src/braket/circuits/gate.py
kshitijc: See reply to the previous comment

src/braket/circuits/observables.py
kshitijc: The openqasm syntax dictates that the observable here should act on either a single target or on `all`. Returning multiple lines would be invalid.

src/braket/circuits/noises.py
kshitijc: This operator should only ever operate on a single target qubit by definition.

src/braket/circuits/gates.py
kshitijc: This is backwards compatible since it inherits the `to_ir` method from the gate class.

src/braket/circuits/observables.py
kshitijc: Since this is only applicable for this function, and not the class or any other method in the class, I scoped it to the function. Happy to move it out if you feel strongly about it.

src/braket/circuits/compiler_directive.py
krneta: If it's a public method, and we allow string, should we validate if it's anything except for the two values that we support?

src/braket/circuits/compiler_directive.py
krneta: Yeah, I can see it either way. Technically, subclasses don't need to implement these methods (they can implement to_ir() instead), but they can choose to rely on the fact that the parent is calling these (what would be protected in C++) private functions. We definitely don't want customers calling these directly, right? That makes me lean toward making them private, but it's up to you.

src/braket/circuits/noises.py
krneta: I don't feel strongly about it, especially since we've done this (for a long time) in "to_jaqcd" as well.

src/braket/circuits/compiler_directive.py
krneta: Also, we should also make sure we document that the argument is unused if the ir_type is not OpenQASM... which is weird. It would be nicer if we could encapsulate the format with the type, since it's a feature of the format.

src/braket/circuits/observables.py
krneta: I don't feel strongly about this, but just for my information, returning "i({qubit_target[0]});i({qubit_target[1]})" for some two-target observable would be invalid?

src/braket/circuits/observables.py
krneta: I think if this becomes the norm it will make our code much more messy & hard to understand. As a code style thing, I think you would be better suited to know.

krneta:
This seems complex enough that it warrants an integ test for the stuff we have. What do you think?

src/braket/circuits/observables.py
kshitijc: Right. The way observables are applied to a result type would make such a return statement invalid in that context.

src/braket/circuits/observables.py
kshitijc: Updated

src/braket/circuits/compiler_directive.py
kshitijc: Updated to private

src/braket/circuits/circuit.py
```diff
@@ -1043,25 +1045,93 @@ def diagram(self, circuit_diagram_class=AsciiCircuitDiagram) -> str:
...
+    ) -> Union[OpenQasmProgram, JaqcdProgram]:
         """
         Converts the circuit into the canonical intermediate representation.
         If the circuit is sent over the wire, this method is called before it is sent.
 
+        Args:
+            ir_type (IRType): The IRType to use for converting the circuit object to its
+                IR representation.
+            qubit_reference_type (QubitReferenceType): The qubit reference type to use for
+                the IR representation.
```
krneta, line 1061: Is setting the qubit_reference_type as PHYSICAL exactly the same as setting it as VIRTUAL, when ir is JAQCD? If so, we should at least document that this value is only relevant for OpenQASM.

src/braket/circuits/circuit.py
```diff
@@ -1043,25 +1050,110 @@ def diagram(self, circuit_diagram_class=AsciiCircuitDiagram) -> str:
...
             basis_rotation_instructions=ir_basis_rotation_instructions,
         )
 
+    def _to_openqasm(
+        self, serialization_properties: OpenQASMSerializationProperties
+    ) -> OpenQasmProgram:
+        ir_instructions, qubit_reference_format = self._create_openqasm_header(
+            serialization_properties
+        )
+        serialization_properties.qubit_reference_format = qubit_reference_format
```
krneta, line 1110: What if this was already set by whoever created the properties? Do we get to overwrite it?

It seems like we should be asking the properties for this information, rather than setting it?

src/braket/circuits/circuit.py
```diff
@@ -1043,25 +1050,110 @@ def diagram(self, circuit_diagram_class=AsciiCircuitDiagram) -> str:
...
+                ]
+            )
+        else:
+            for idx, qubit in enumerate(self.qubits):
+                qubit_target = qubit_reference_format.format(int(qubit))
+                ir_instructions.append(f"b[{idx}] = measure {qubit_target};")
+
+        return OpenQasmProgram.construct(source="\n".join(ir_instructions))
+
+    def _create_openqasm_header(
```
krneta, line 1137: I would move this to a create_header() function in the properties, I think.

src/braket/circuits/noises.py
```diff
@@ -79,9 +80,15 @@ def __init__(self, probability: float):
...
         )
 
-    def to_ir(self, target: QubitSet):
+    def _to_jaqcd(self, target: QubitSet):
         return ir.BitFlip.construct(target=target[0], probability=self.probability)
 
+    def _to_openqasm(
+        self, target: QubitSet, serialization_properties: OpenQASMSerializationProperties
+    ):
+        target_qubit = serialization_properties.qubit_reference_format.format(int(target[0]))
```
krneta, line 89: I think I would just have a method in the properties called "format_target(target:Qubit)" or  "format_target(target:QubitSetInput)"  that does the conversion for you.

src/braket/circuits/circuit.py
kshitijc: This method has logic related to the circuit serializing itself, so I think it might be better suited here. Since this is private we can move it to the properties later if needed.

src/braket/circuits/circuit.py
kshitijc: It doesn't make sense for the user to set it and would be ignored. Removed.

src/braket/circuits/noises.py
kshitijc: ok, added.

krneta (approved):

PR #324, opened 2022-04-11
fix: add device arn error handling for badly formed ARNs

ajberdy:
*Issue #, if available:*

*Description of changes:*

While device ARN validation is handled by the service, we do some basic preprocessing that assumes the ARN will be of a certain format (ie. having at least 3 :'s). Previously, an ARN that doesn't meet this requirement would cause a crash in the BDK; now this throws an error gracefully.

*Testing done:*

Added a unit test.

## Merge Checklist

#### General

#### Tests

### DIFF
--- a/src/braket/aws/aws_quantum_job.py
+++ b/src/braket/aws/aws_quantum_job.py
@@ -529,9 +529,16 @@ class AwsQuantumJob(QuantumJob):
         return hash(self.arn)
 
     @staticmethod
+    def _get_device_region(device_arn: str) -> str:
+        try:
+            return device_arn.split(":")[3]
+        except IndexError:
+            raise ValueError(f"Device arn is not a valid format: {device_arn}")
+
+    @staticmethod
     def _initialize_session(session_value, device, logger):
         aws_session = session_value or AwsSession()
-        device_region = device.split(":")[3]
+        device_region = AwsQuantumJob._get_device_region(device)
         return (
             AwsQuantumJob._initialize_regional_device_session(aws_session, device, logger)
             if device_region
@@ -540,7 +547,7 @@ class AwsQuantumJob(QuantumJob):
 
     @staticmethod
     def _initialize_regional_device_session(aws_session, device, logger):
-        device_region = device.split(":")[3]
+        device_region = AwsQuantumJob._get_device_region(device)
         current_region = aws_session.region
         if current_region != device_region:
             aws_session = aws_session.copy_session(region=device_region)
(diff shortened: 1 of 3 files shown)


src/braket/aws/aws_quantum_job.py
```diff
@@ -528,10 +528,17 @@ def __eq__(self, other) -> bool:
     def __hash__(self) -> int:
         return hash(self.arn)
 
+    @staticmethod
+    def _get_device_region(device_arn: str) -> str:
+        try:
+            return device_arn.split(":")[3]
+        except IndexError:
+            raise ValueError(f"Device arn is not a valid format: {device_arn}")
```
math411, line 536: Would it make sense to include a global and regional arn as examples in the error output?
```suggestion
            raise ValueError(f"Device arn is not a valid format: {device_arn}"). Example of correct arns are:
            arn:aws:braket:::device/qpu/d-wave/Advantage_system1 or arn:aws:braket:us-west-2::device/qpu/d-wave/Advantage_system1
```

ajberdy: Sure, that could make sense. Adding a region to a non-regionalized ARN will not work though, so we'll have to use M1 as an example. Also need to update the test.

src/braket/aws/aws_quantum_job.py
```diff
@@ -528,10 +528,21 @@ def __eq__(self, other) -> bool:
...
 
+    @staticmethod
+    def _get_device_region(device_arn: str) -> str:
+        try:
+            return device_arn.split(":")[3]
+        except IndexError:
+            raise ValueError(
+                f"Device arn is not a valid format: {device_arn}. Examples of correct arns are: "
+                "arn:aws:braket:::device/quantum-simulator/amazon/sv1 or "
+                "arn:aws:braket:us-west-1::device/qpu/rigetti/Aspen-M-1."
```
kshitijc, line 539: I'm not sure if we should add specific arns here. Might be better to point to the docs specifying the valid arns: https://docs.aws.amazon.com/braket/latest/developerguide/braket-devices.html

kshitijc (commented):
Tagging @aws-rbs for reviewing the error message string.

kshitijc (commented):
The same logic exists in aws_device as well. Do we wish to update that as well ?

ajberdy:
> The same logic exists in aws_device as well. Do we wish to update that as well ?

Yes, good call.

kshitijc (approved):

PR #354, opened 2022-05-19
Openqasm local simulator

ajberdy:
*Issue #, if available:*

*Description of changes:*

Add infrastructure for running openqasm local simulations from the sdk

*Testing done:*

Tox unit tests 100%

## Merge Checklist

#### General

#### Tests


src/braket/tasks/gate_model_quantum_task_result.py
```diff
@@ -278,13 +279,25 @@ def _from_object_internal_computational_basis_sampling(cls, result: GateModelTas
...
-            if result.resultTypes
-            else GateModelQuantumTaskResult._calculate_result_types(
+        if result.resultTypes:
+            # Jaqcd does not return anything in the resultTypes schema field since the
+            # result types are easily parsable from the IR. However, an OpenQASM program
+            # specifies result types inline and parsing result types is more involved
+            # (ie. may involve dereferencing logical qubits at runtime), so the parsed
+            # result type specifications need to be returned, even if not calculated
+            # during simulation.
+            if not isinstance(result.resultTypes[0], ResultTypeValue):
```
speller26, line 289: Not sure I understand what the new logic branch looks like; the old code already did the trick for OpenQASM programs sent to the service no? Is it possible to unify the results format?

speller26 (commented):
Looks good, just some questions

src/braket/tasks/gate_model_quantum_task_result.py
ajberdy: Good analysis here. The service got around this limitation by calculating result types on the service side for openqasm programs, in contrast to the jaqcd decision to calculate shots result types client side. We could take the same approach here, but then we would be duplicating the logic in this file in the simulator repo.

I thought it was a cleaner solution and more in line with the design decisions made for jaqcd circuits to simply pass back information about the result type types and add a branch here. It is a slight abuse of schema typing and may be confusing since we have several different IRs for tracking result types throughout the service. Nothing we can't fix with a union in our schemas lol

speller26: That makes sense; also brings back to the table whether we should also allow device-defined behavior for shots>0.

src/braket/tasks/local_quantum_task.py
```diff
@@ -23,7 +23,10 @@ class LocalQuantumTask(QuantumTask):
     Since this class is instantiated with the results, cancel() and run_async() are unsupported.
     """
 
-    def __init__(self, result: Union[GateModelQuantumTaskResult, AnnealingQuantumTaskResult]):
+    def __init__(
```
speller26, line 26: Interesting that tox did this

speller26 (approved):

src/braket/tasks/local_quantum_task.py
ajberdy: Ah I think what actually happened here is originally I created a new result type and expanded the union to three fields, which made tox split the line; when I removed the new type tox found the split line still acceptable

PR #357, opened 2022-05-20
fix: removing validation for disable_qubit_rewiring

krneta:
*Issue #, if available:*
This fix will unblock the customer from validating the suggestion in this issue : https://github.com/aws/amazon-braket-sdk-python/issues/348 

*Description of changes:*
Removing validation for disable_qubit_rewiring and letting the service handle it.

It seems like this check was done as an optimization, since the values should be validated service-side. I'm removing this because different devices handle verbatim differently and some don't require disable_qubit_rewiring for verbatim (in fact, they don't support the flag at all). Rather than trying to handle these differences client side (especially in AwsTask, rather than AwsDevice), I think we can remove the check and let the service handle it.

***Note:*** As a result of this change, customers will no longer get a ValueError when providing the incorrect verbatim/disable_qubit_rewiring combination - instead they will get a ValidationException from the service call. 

*Testing done:*
Ran tox.

## Merge Checklist

#### General

#### Tests

### DIFF
--- a/src/braket/aws/aws_quantum_task.py
+++ b/src/braket/aws/aws_quantum_task.py
@@ -446,11 +446,6 @@ def _(
     **kwargs,
 ) -> AwsQuantumTask:
     validate_circuit_and_shots(circuit, create_task_kwargs["shots"])
-    if circuit.qubits_frozen and not disable_qubit_rewiring:
-        raise ValueError(
-            "disable_qubit_rewiring must be True to run circuit with compiler directives"
-        )
-
     # TODO: Update this to use `deviceCapabilities` from Amazon Braket's GetDevice operation
     # in order to decide what parameters to build.
     paradigm_parameters = GateModelParameters(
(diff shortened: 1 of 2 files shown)


jsrikrishna (dismissed):

src/braket/aws/aws_quantum_task.py
```diff
@@ -446,11 +444,6 @@ def _(
     **kwargs,
 ) -> AwsQuantumTask:
     validate_circuit_and_shots(circuit, create_task_kwargs["shots"])
-    if circuit.qubits_frozen and not disable_qubit_rewiring:
-        raise ValueError(
-            "disable_qubit_rewiring must be True to run circuit with compiler directives"
-        )
```
kshitijc, line 452: What error type does the service return? A change in the error type might be a breaking change here.

krneta: I wasn't able to test that because we're outside the availability window (it just tells me the device is not available).

mbeach-aws: On OQC? 
```python
ValidationException: An error occurred (ValidationException) when calling the CreateQuantumTask operation: Device arn:aws:braket:eu-west-2::device/qpu/oqc/Lucy does not support disabled qubit rewiring.
```

On Rigetti
```python
ValidationException: An error occurred (ValidationException) when calling the CreateQuantumTask operation: VerbatimBoxes cannot be used without disableQubitRewiring parameter set to "True".Please set disableQubitRewiring parameter to "True" or do not use VerbatimBoxes.
```

kshitijc (approved):

PR #393, opened 2022-06-23
Adding timestamp order

krneta:
*Issue #, if available:*

*Description of changes:*

*Testing done:*

## Merge Checklist

#### General

#### Tests

### DIFF
--- a/src/braket/ahs/time_series.py
+++ b/src/braket/ahs/time_series.py
@@ -29,6 +29,7 @@ class TimeSeries:
     def __init__(self):
         self._series = OrderedDict()
         self._sorted = True
+        self._largest_time = -1
 
     def put(
         self,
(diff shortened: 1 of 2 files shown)


src/braket/ahs/time_series.py
```diff
@@ -43,7 +45,11 @@ def put(
         Returns:
             TimeSeries: returns self (to allow for chaining).
         """
-        self._series[time] = TimeSeriesItem(time, value)
+        if time in self._series:
+            self._series[time] = TimeSeriesItem(time, value)
+        else:
+            self._series[time] = TimeSeriesItem(time, value)
+            self._sorted = False
```
peterkomar-aws, line 52: This is overly cautious. 
(In a common use case, when the user puts timepoints in ascending order, they are sorted, despite all time values are new. I'm okay with the current implementation, but if you'd like you can cover this case by keeping a highest time as an attribute, and compare against that too. This would lower the performance of put though, so I'm not sure if it is worth it.)

peterkomar-aws (approved):
Thank you!

src/braket/ahs/time_series.py
krneta: Yeah, that's a great idea. We don't need this class to be at peak performance, but I don't think it adds much to complexity or runtime of put to add the additional condition.

peterkomar-aws (approved):

PR #397, opened 2022-06-29
Feature/openqasm generation

ajberdy:
bring in openqasm generation for observables and result types


src/braket/circuits/result_types.py
```diff
@@ -45,6 +45,9 @@ def __init__(self):
     def _to_jaqcd(self) -> ir.StateVector:
         return ir.StateVector.construct()
 
+    def _to_openqasm(self, serialization_properties: OpenQASMSerializationProperties) -> str:
+        return "#pragma braket result state_vector"
```
speller26, line 49: nit: _might_ be worth extracting `pragma braket result` into a constant

speller26 (approved):

src/braket/circuits/result_types.py
ajberdy: Maybe, not sure the optimality gain is worth the readability loss. Besides, this PR is just to keep the feature branches in sync, if we make a change here, we'd have to open a pr into feature/openqasm_generation, which already has this code (without the change)

PR #401, opened 2022-07-06
fix: Lazily parse schemas for devices so getDevice calls do not rely …

math411:
…on the device.

*Issue #, if available:*

*Description of changes:*
Lazily parse schemas for device properties

*Testing done:*
`tox`

## Merge Checklist

#### General

#### Tests

### DIFF
--- a/src/braket/aws/aws_device.py
+++ b/src/braket/aws/aws_device.py
@@ -282,8 +282,7 @@ class AwsDevice(Device):
         self._status = metadata.get("deviceStatus")
         self._type = AwsDeviceType(metadata.get("deviceType"))
         self._provider_name = metadata.get("providerName")
-        qpu_properties = metadata.get("deviceCapabilities")
-        self._properties = BraketSchemaBase.parse_raw_schema(qpu_properties)
+        self._properties = metadata.get("deviceCapabilities")
         self._topology_graph = self._construct_topology_graph()
 
     @property
@@ -370,7 +369,7 @@ class AwsDevice(Device):
         Please see `braket.device_schema` in amazon-braket-schemas-python_
 
         .. _amazon-braket-schemas-python: https://github.com/aws/amazon-braket-schemas-python"""
-        return self._properties
+        return BraketSchemaBase.parse_raw_schema(self._properties)
 
     @property
     def topology_graph(self) -> DiGraph:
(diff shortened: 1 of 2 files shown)


src/braket/aws/aws_device.py
```diff
@@ -370,7 +369,7 @@ def properties(self) -> DeviceCapabilities:
         Please see `braket.device_schema` in amazon-braket-schemas-python_
 
         .. _amazon-braket-schemas-python: https://github.com/aws/amazon-braket-schemas-python"""
-        return self._properties
+        return BraketSchemaBase.parse_raw_schema(self._properties)
```
krneta, line 373: This will work for the specific problem we had last time, but not if we have other failures when getting devices in bulk (or doing other stuff in bulk). For example, what if we can't convert metadata.get("deviceType") to a known AwsDeviceType? Or what if constructing the topology graph fails? Should we fail the whole get_devices call?

- I think we should add error checking in get_devices in each place that can cause an exception - session_for_region, search_devices and AwsDevice(arn, session_for_region). The trouble here is that there's no way to surface errors to the customer (other than an exception).  That makes things tricky. Perhaps we should log a warning and return the ones we have? It's definitely not ideal, but I think we could argue it's better than the alternative.

- Can you take a look at the way that batch tasks? For example, do we want to fail getting the results if one of the results from one of the tasks can't be parsed?

- You may also want to add error checking in search_devices so that a single bad result doesn't fail them all.

krneta (changes requested):

src/braket/aws/aws_device.py
```diff
@@ -387,7 +385,7 @@ def topology_graph(self) -> DiGraph:
 
             >>> print(device.topology_graph.edges)
         """
-        return self._topology_graph
+        return self._construct_topology_graph()
```
krneta, line 388: Do we need to cache the results for either of these?

src/braket/aws/aws_device.py
math411: Per an offline discussion, going to be keeping the scope to just the aws_device changes in the PR.

src/braket/aws/aws_device.py
math411: The use of caching may conflict with `refresh_metadata`. Benchmarking has not shown a noticeable decrease in performance using non cached data.

krneta (approved):

PR #408, opened 2022-07-25
feature: provide easy mechanism to update the local container when running local job.

krneta:
*Issue #, if available:*

*Description of changes:*
Added ability for customers to use a parameter called "local_container_update" to update their local container before running a local job. This parameter is only available when running local mode and it is optional and "True" by default. We decided to make it True by default to reduce the number of instances where there would be a mismatch between running in local mode and running an actual Hybrid Job.

*Testing done:*
I have tested to make sure I can run jobs without the parameter, and with it True/False, and it correctly tries to update the container when it is true.

## Merge Checklist

#### General

#### Tests


src/braket/jobs/local/local_job_container.py
```diff
@@ -146,6 +154,13 @@ def _start_container(self, image_uri: str) -> str:
                     f"The URL {image_uri} is not available locally and can not be pulled from ECR."
                     " Please pull down the container before proceeding."
                 )
+        elif force_update:
+            try:
+                self._pull_image(image_uri)
```
math411, line 159: Do we want any sort of retry logic for the force update?

krneta: This just mirrors the existing pull image logic (with the ability to fall back to the existing image). We might want to add retry logic, but it should affect all the "pull_image" paths, not just here, and it would be a bigger (separate) task. However, I'm not sure that it's needed. The only times I've seen this fail is due to my own errors (mistyping something, or needing to update my credentials, etc.).

math411 (approved):

math411 (approved):

PR #413, opened 2022-07-28
OQ3 default

ajberdy:
*Issue #, if available:*

*Description of changes:*

Use OpenQASM by default in the BDK.

*Testing done:*

Unit/Integ

## Merge Checklist

#### General

#### Tests

### DIFF
--- a/src/braket/aws/aws_quantum_task.py
+++ b/src/braket/aws/aws_quantum_task.py
@@ -25,7 +25,11 @@ from braket.annealing.problem import Problem
 from braket.aws.aws_session import AwsSession
 from braket.circuits.circuit import Circuit
 from braket.circuits.circuit_helpers import validate_circuit_and_shots
-from braket.circuits.serialization import IRType, OpenQASMSerializationProperties, QubitReferenceType
+from braket.circuits.serialization import (
+    IRType,
+    OpenQASMSerializationProperties,
+    QubitReferenceType,
+)
 from braket.device_schema import GateModelParameters
 from braket.device_schema.dwave import (
     Dwave2000QDeviceParameters,
@@ -503,9 +507,7 @@ def _(
 
     serialization_properties = OpenQASMSerializationProperties(
         qubit_reference_type=(
-            QubitReferenceType.PHYSICAL
-            if disable_qubit_rewiring
-            else QubitReferenceType.VIRTUAL
+            QubitReferenceType.PHYSICAL if disable_qubit_rewiring else QubitReferenceType.VIRTUAL
         )
     )
 
--- a/src/braket/aws/aws_quantum_task.py
+++ b/src/braket/aws/aws_quantum_task.py
@@ -25,6 +25,7 @@ from braket.annealing.problem import Problem
 from braket.aws.aws_session import AwsSession
 from braket.circuits.circuit import Circuit
 from braket.circuits.circuit_helpers import validate_circuit_and_shots
+from braket.circuits.serialization import IRType
 from braket.device_schema import GateModelParameters
 from braket.device_schema.dwave import (
     Dwave2000QDeviceParameters,
@@ -45,8 +46,6 @@ from braket.ir.blackbird import Program as BlackbirdProgram
 from braket.ir.openqasm import Program as OpenQasmProgram
 from braket.schema_common import BraketSchemaBase
 from braket.task_result import AnnealingTaskResult, GateModelTaskResult, PhotonicModelTaskResult
-
-from braket.circuits.serialization import IRType
 from braket.tasks import (
     AnnealingQuantumTaskResult,
     GateModelQuantumTaskResult,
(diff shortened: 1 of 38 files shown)


src/braket/aws/aws_quantum_task.py
```diff
@@ -501,7 +503,10 @@ def _(
         )
 
     create_task_kwargs.update(
-        {"action": circuit.to_ir().json(), "deviceParameters": device_parameters.json()}
+        {
+            "action": circuit.to_ir(ir_type=IRType.OPENQASM).json(),
```
kshitijc, line 507: How do we determine the `SerializationProperties` for `to_ir()` ?

src/braket/devices/local_simulator.py
```diff
@@ -134,12 +136,11 @@ def _run_internal(
 
 @_run_internal.register
 def _(circuit: Circuit, simulator: BraketSimulator, shots, *args, **kwargs):
-    if DeviceActionType.JAQCD not in simulator.properties.action:
+    if DeviceActionType.OPENQASM not in simulator.properties.action:
```
kshitijc, line 139: I think we need to be careful about not introducing a breaking change here:
* pinning the local simulator to version >= (the one adding support for OpenQASM)
* the scenario where a customer is using another local simulator that they have developed. They might not want to support OpenQASM and yet still be able to use the LocalSimulator class to interact with their simulator.

src/braket/aws/aws_quantum_task.py
ajberdy: The default, virtual, is used. It is set [here](https://github.com/aws/amazon-braket-sdk-python/blob/feature/openqasm_generation/src/braket/circuits/serialization.py#L38). I believe that's the correct choice, since every device supports virtual qubits, but the local simulator does not support physical qubits.

src/braket/devices/local_simulator.py
ajberdy: That's a good point. I think this is the case no matter how we approach this. If we flip the bdk to use OpenQASM by default, then it won't be compatible with a LocalSimulator that doesn't support OpenQASM. We could always launch the OpenQASM simulator and leave Jaqcd as the default, but I believe we want to switch over.

I think pinning the version is a good solution here. A customer developing their own local simulator will likely be able to figure out what's going on and resolve quickly

ajberdy: Perhaps then we should merge the simulator repo a day earlier and release a new version so we can pin the dependency.

src/braket/aws/aws_quantum_task.py
kshitijc: I think we might want to use physical qubits in certain cases (for eg. when disableQubitRewiring is set).

src/braket/devices/local_simulator.py
kshitijc: Maybe another possibility is: 
```
if device supports OpenQASM:
  use OpenQASM
elif device supports JAQCD:
  use JAQCD
else: 
  raise error
```

ajberdy: Oh right, good idea. Maybe I'll blame the covid brain fog for missing that option 😛 jk

src/braket/aws/aws_quantum_task.py
```diff
@@ -500,8 +505,20 @@ def _(
             paradigmParameters=paradigm_parameters
         )
 
+    serialization_properties = OpenQASMSerializationProperties(
+        qubit_reference_type=(
+            QubitReferenceType.PHYSICAL if disable_qubit_rewiring else QubitReferenceType.VIRTUAL
```
kshitijc, line 510: I think we also need to check if the circuit is verbatim

src/braket/devices/local_simulator.py
```diff
@@ -134,13 +135,18 @@ def _run_internal(
...
-    results = simulator.run(program, qubits, shots, *args, **kwargs)
-    return GateModelQuantumTaskResult.from_object(results)
+    if DeviceActionType.OPENQASM in simulator.properties.action:
+        validate_circuit_and_shots(circuit, shots)
+        program = circuit.to_ir(ir_type=IRType.OPENQASM)
+        results = simulator.run(program, shots, *args, **kwargs)
+        return GateModelQuantumTaskResult.from_object(results)
+    elif DeviceActionType.JAQCD in simulator.properties.action:
+        validate_circuit_and_shots(circuit, shots)
+        program = circuit.to_ir(ir_type=IRType.OPENQASM)
```
kshitijc, line 145: * Should this be IRType.JAQCD instead ?
* Can we add a test for this scenario ?

ajberdy: Yes, good catch. Added.

src/braket/aws/aws_quantum_task.py
ajberdy: Yes, confirmed this.

kshitijc (approved):

PR #429, opened 2022-08-22
Fixing docs.

krneta:
*Issue #, if available:*

*Description of changes:*

*Testing done:*

## Merge Checklist

#### General

#### Tests

### DIFF
--- a/src/braket/aws/aws_quantum_job.py
+++ b/src/braket/aws/aws_quantum_job.py
@@ -123,7 +123,7 @@ class AwsQuantumJob(QuantumJob):
                 For convenience, this accepts other types for keys and values, but `str()`
                 is called to convert them before being passed on. Default: None.
 
-            input_data (Union[str, S3DataSourceConfig, dict]): Information about the training
+            input_data (Union[str, Dict, S3DataSourceConfig]): Information about the training
                 data. Dictionary maps channel names to local paths or S3 URIs. Contents found
                 at any local paths will be uploaded to S3 at
                 f's3://{default_bucket_name}/jobs/{job_name}/data/{channel_name}. If a local
@@ -208,7 +208,7 @@ class AwsQuantumJob(QuantumJob):
         """
         Args:
             arn (str): The ARN of the job.
-            aws_session (AwsSession, optional): The `AwsSession` for connecting to AWS services.
+            aws_session (AwsSession): The `AwsSession` for connecting to AWS services.
                 Default is `None`, in which case an `AwsSession` object will be created with the
                 region of the job.
         """
@@ -259,7 +259,7 @@ class AwsQuantumJob(QuantumJob):
         """The state of the quantum job.
 
         Args:
-            use_cached_value (bool, optional): If `True`, uses the value most recently retrieved
+            use_cached_value (bool): If `True`, uses the value most recently retrieved
                 value from the Amazon Braket `GetJob` operation. If `False`, calls the
                 `GetJob` operation to retrieve metadata, which also updates the cached
                 value. Default = `False`.
@@ -350,7 +350,7 @@ class AwsQuantumJob(QuantumJob):
         """Gets the job metadata defined in Amazon Braket.
 
         Args:
-            use_cached_value (bool, optional): If `True`, uses the value most recently retrieved
+            use_cached_value (bool): If `True`, uses the value most recently retrieved
                 from the Amazon Braket `GetJob` operation, if it exists; if does not exist,
                 `GetJob` is called to retrieve the metadata. If `False`, always calls
                 `GetJob`, which also updates the cached value. Default: `False`.
@@ -382,7 +382,7 @@ class AwsQuantumJob(QuantumJob):
                 when there is a conflict. Default: MetricStatistic.MAX.
 
         Returns:
-            Dict[str, List[Union[str, float, int]]] : The metrics data.
+            Dict[str, List[Any]] : The metrics data.
         """
         fetcher = CwlInsightsMetricsFetcher(self._aws_session)
         metadata = self.metadata(True)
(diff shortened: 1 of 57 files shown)


src/braket/circuits/moments.py
```diff
@@ -272,17 +283,20 @@ def items(self) -> ItemsView[MomentsKey, Instruction]:
...
 
-    def get(self, key: MomentsKey, default=None) -> Instruction:
+    def get(self, key: MomentsKey, default: Any = None) -> Instruction:
         """
         Get the instruction in self by key.
 
         Args:
             key (MomentsKey): Key of the instruction to fetch.
-            default (Any, optional): Value to return if `key` is not in `moments`. Default = `None`.
+            default (Any): Value to return if `key` is not in `moments`. Default = `None`.
```
kshitijc, line 299:
```suggestion
            default (Optional[Any]): Value to return if `key` is not in `moments`. Default = `None`.
```

src/braket/circuits/noise.py
```diff
@@ -145,18 +147,26 @@ def __repr__(self):
 
     @classmethod
     def from_dict(cls, noise: dict) -> Noise:
+        """
+        Converts a dictionary representing an object of this class into an instance of this class.
+
+        Args:
+            noise (dict): A dictionary representation of an object of this class.
```
kshitijc, line 154: Is `dict` a valid type (I'm confused since we do from typing import Dict)

src/braket/circuits/quantum_operator_helpers.py
```diff
@@ -23,7 +23,7 @@ def verify_quantum_operator_matrix_dimensions(matrix: np.array) -> None:
     raising `ValueError` otherwise.
 
     Args:
-        matrix (np.ndarray): matrix to verify
+        matrix (array): matrix to verify
```
kshitijc, line 26: What is the array type ? np.ndarray tells me that it is a numpy array but this makes it ambiguous?

kshitijc (commented):
Thanks for these changes! There appear to be inconsistencies in the doc strings for Optional args. Could we please fix all those instances?

src/braket/aws/aws_quantum_job.py
```diff
@@ -208,7 +208,7 @@ def __init__(self, arn: str, aws_session: AwsSession = None):
         """
         Args:
             arn (str): The ARN of the job.
-            aws_session (AwsSession, optional): The `AwsSession` for connecting to AWS services.
+            aws_session (AwsSession): The `AwsSession` for connecting to AWS services.
```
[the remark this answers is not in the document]
ajberdy: IIRC we use `, optional` in a lot of our code ([example](https://github.com/aws/amazon-braket-sdk-python/blob/main/src/braket/aws/aws_quantum_task.py#L178))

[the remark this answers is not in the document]
krneta: We do this inconsistently. In some places we do it for arguments that are optional (param: int = 0). In some places we only do it when it's an optional object (Optional[Obj] = None). We should either do it consistently, or not at all. I've opted to remove them, but we can change our mind and add them everywhere.

src/braket/aws/aws_session.py
```diff
@@ -202,17 +232,14 @@ def create_job(self, **boto3_kwargs) -> str:
         """
         Create a quantum job.
 
-        Args:
-            **boto3_kwargs: Keyword arguments for the Amazon Braket `CreateJob` operation.
```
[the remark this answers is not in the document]
krneta: We're inconsistent in how/when we specify args and kwargs. We also don't provide any information in the documentation other than the obvious. I'm ambivalent about removing them, but by it's nature args/kwargs don't have specific meanings (most of the time). So I opted to remove them all. 

The other option would be to always require documentation whenever we have args/kwargs in the signature.

src/braket/circuits/circuit.py
```diff
@@ -1353,9 +1358,9 @@ def subroutine(register=False):
         Instruction('operator': 'H', 'target': QubitSet(Qubit(1),))
     """
 
-    def subroutine_function_wrapper(func: Callable[..., SubroutineReturn]) -> SubroutineReturn:
+    def _subroutine_function_wrapper(func: Callable[..., SubroutineReturn]) -> SubroutineReturn:
```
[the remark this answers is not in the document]
krneta: I currently have our linter set so all public functions should be documented (with some exceptions for simple return statements). By making this a private function we signal that it doesn't strictly need documentation.

src/braket/circuits/noise.py
krneta: The only difference I've seen is that dict doesn't specify type hints for what's in the dict, and Dict does. So it seemed like our standard was that Dict[Any, Any] == dict.

src/braket/circuits/quantum_operator_helpers.py
krneta: Although it's often used, "np" is actually not a module.  I think we should document it as numpy.array or array. I chose the shorter option, since we often don't specify the modules for objects that we use, and I don't think that numpy should be an exception (what do you think?)

ajberdy: I think np is pretty ubiquitous and also in scope, so I personally think np.ndarray is more clear for docs. Numpy just uses "ndarray" and "array_like" in their documentation, which could be another option too

src/braket/circuits/noise.py
ajberdy: fyi you can do `dict[type, type]` from python 3.10

src/braket/circuits/quantum_operator_helpers.py
krneta: We actually use "array" here in the type hint, or I would have left the doc as "ndarray" as well. If it they really are ndarrays, we should change both.

ajberdy: Oh good catch, ya I believe that should be ndarray

src/braket/aws/aws_session.py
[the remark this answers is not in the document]
krneta: OK, I added these back in, since I'm ambivalent about it. We can decide on this later. 

We still have a handful of errors to look at after this PR is done.

kshitijc (approved):

PR #450, opened 2022-09-22
feature: support inputs in the device interface

ajberdy:
*Issue #, if available:*

*Description of changes:*

Support specifying inputs from `device.run` for the local simulator. A circuit with free parameters, when passed along with correct input, will run successfully on the local simulator. The AwsDevice class does not at this moment have an option for inputs added, so the behavior is unchanged.

*Testing done:*

Unit tests and manual e2e verification.

## Merge Checklist

#### General

#### Tests

### DIFF
--- a/src/braket/devices/device.py
+++ b/src/braket/devices/device.py
@@ -35,8 +35,8 @@ class Device(ABC):
     def run(
         self,
         task_specification: Union[Circuit, Problem],
-        shots: Optional[int],
-        inputs: Optional[Dict[str, float]],
+        shots: Optional[int] = 0,
+        inputs: Optional[Dict[str, float]] = None,
         *args,
         **kwargs
     ) -> QuantumTask:
--- a/src/braket/devices/device.py
+++ b/src/braket/devices/device.py
@@ -12,7 +12,7 @@
 # language governing permissions and limitations under the License.
 
 from abc import ABC, abstractmethod
-from typing import Optional, Union
+from typing import Dict, Optional, Union
 
 from braket.annealing.problem import Problem
 from braket.circuits import Circuit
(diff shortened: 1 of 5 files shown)


src/braket/devices/device.py
```diff
@@ -33,7 +33,12 @@ def __init__(self, name: str, status: str):
 
     @abstractmethod
     def run(
-        self, task_specification: Union[Circuit, Problem], shots: Optional[int], *args, **kwargs
+        self,
+        task_specification: Union[Circuit, Problem],
+        shots: Optional[int],
+        inputs: Optional[Dict[str, float]],
```
speller26, line 39: Do we want to default here too?
```suggestion
        inputs: Optional[Dict[str, float]] = None,
```

speller26 (approved):

src/braket/devices/device.py
ajberdy: Ya then we should probably do the same for shots

src/braket/devices/device.py
```diff
@@ -33,7 +33,12 @@ def __init__(self, name: str, status: str):
 
     @abstractmethod
     def run(
-        self, task_specification: Union[Circuit, Problem], shots: Optional[int], *args, **kwargs
+        self,
+        task_specification: Union[Circuit, Problem],
+        shots: Optional[int] = 0,
```
speller26, line 38: This I'm not as convinced by, since QPUs are also based on this. No strong opinions though.

speller26 (approved):

src/braket/devices/device.py
ajberdy: Then with that logic, task specification could be e.g. an annealing problem which won't take inputs, so maybe we shouldn't have defaults for either?

speller26 (approved):

PR #479, opened 2022-11-23
loosen oqpy requirement

natestemen:
*Issue #, if available:* NA

*Description of changes:* `oqpy` [released version 0.1.2](https://github.com/openqasm/oqpy/releases/tag/v0.1.2) which is python 3.11 compatible. We'd like to support python 3.11 over on Mitiq as well, but are blocked by the pinned version of oqpy here. I've gone ahead and removed the version specifier completely since many other dependencies are also unspecified, but let me know if having some range would be better.

*Testing done:* None

#### General

#### Tests

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -29,7 +29,7 @@ setup(
     install_requires=[
         "amazon-braket-schemas>=1.12.0",
         "amazon-braket-default-simulator>=1.10.0",
-        "oqpy",
+        "oqpy~=0.1.1",
         "backoff",
         "boltons",
         "boto3>=1.22.3",
--- a/setup.py
+++ b/setup.py
@@ -29,7 +29,7 @@ setup(
     install_requires=[
         "amazon-braket-schemas>=1.12.0",
         "amazon-braket-default-simulator>=1.10.0",
-        "oqpy==0.1.1",
+        "oqpy",
         "backoff",
         "boltons",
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -29,7 +29,7 @@
     install_requires=[
         "amazon-braket-schemas>=1.12.0",
         "amazon-braket-default-simulator>=1.10.0",
-        "oqpy==0.1.1",
+        "oqpy",
```
kshitijc, line 32: Could we please change this to the compatible release clause instead? (`oqpy ~= 0.1.1`) This would ensure that we don't pull in an incompatible release without validating it first.

natestemen: Makes sense! Thanks for the reminder about this.

natestemen:
Not really sure why flake8 seems to be complaining about this. Any ideas?

math411:
@natestemen, changing https://github.com/aws/amazon-braket-sdk-python/pull/479/files#diff-60f61ab7a8d1910d86d9fda2261620314edcae5894d5aaa236b821c7256badd7R48 to:
```
"flake8<=5.0.4",
```
should fix this. Seems that the latest release is causing issues here

kshitijc (approved):

kshitijc:
Thanks for this change @natestemen! 👍

PR #480, opened 2022-11-28
fix: abort batch task submission on interrupt

shpface:
*Issue #, if available:*

*Description of changes:*

*Testing done:*

## Merge Checklist

#### General

#### Tests


src/braket/aws/aws_quantum_task_batch.py
```diff
@@ -130,24 +130,30 @@ def _execute(
...
+                        poll_interval_seconds=poll_interval_seconds,
+                        *args,
+                        **kwargs,
+                    )
+                    for task in task_specifications
+                ]
+            tasks = [future.result() for future in task_futures]
+        except:  # noqa: E722
+            # If we are unable to return the tasks, attempt to cancel any pending tasks
+            for future in task_futures:
```
math411, line 154: Do we want to log that we are canceling all remaining tasks or some sort of logging?

shpface: I think the initial comment was a bit unclear. I think if tasks were being cancelled, that would be worth noting. I have edited it to clarify that tasks are not being cancelled on abortion. The only action is to stop creating new tasks. Since the exception is then reraised to be handled elsewhere, I do not think logging would add as much value here.

src/braket/aws/aws_quantum_task_batch.py
```diff
@@ -130,24 +130,31 @@ def _execute(
...
+                        shots,
+                        poll_timeout_seconds=poll_timeout_seconds,
+                        poll_interval_seconds=poll_interval_seconds,
+                        *args,
+                        **kwargs,
+                    )
+                    for task in task_specifications
+                ]
+            tasks = [future.result() for future in task_futures]
+        except:  # noqa: E722
```
krneta, line 152: Can we be more specific? I'm worried that a single exception that could be retried will fail all the tasks.

shpface: We could be more specific, but I don't think we actually want to be. Firstly, no tasks are failed; only the background threads are stopped from submitting new tasks. The existing behavior when there's an exception (potentially corresponding to a retryable error) is that the `AwsQuantumTaskBatch` fails to be created and the exception propagates to whatever handles it. In the meantime, the background threads continue submitting tasks from the `task_specifications`. At this point, there is no longer any way for the main thread to stop these background threads or even to query which tasks are still in their queues to submit. To me it seems harder to successfully handle the error without access to these, but if you think there is a good use case for this, we could just catch `KeyboardInterrupt`.

krneta: Good point. I guess my concern was that if I submit 100 tasks and something causes an error at task 2, there may still be value in having all the other 98 tasks be started/completed (if they are in an availability window, for example). However, you're right that it would be cumbersome for customers to go through them since they'd need to do that manually, and making that easier isn't simple.

shpface: Thank you, I think that is a good point. When one of the worker threads hits an error, it will log the error to the `Future`. The thread pool will continue executing the rest of the tasks in the submission queue, and the exception will only be rethrown when the main thread checks for the value of `future.result()`, so such an error does not prevent the submission of the rest of the tasks. This change will only cancel the submission queue when the exception is raised on the main thread while waiting for the workers to finish, such as would happen when the user tries to interrupt with ctrl-C. Any other exception would likely come from an error when attempting to join the threads, which seems unlikely enough that I am restricting the handling to `KeyboardInterrupt` and we can reevaluate if more is needed later.

krneta (dismissed):

krneta (approved):

PR #486, opened 2022-12-08
fix: Hamiltonian ascii simplification

ajberdy:
*Issue #, if available:*

*Description of changes:*

Default ASCII representation of Hamiltonians (Sum observables) to "Hamiltonian".

Opt to underestimate qubit count for Sum observables rather than overestimate

*Testing done:*

unit testing, manual performance testing

## Merge Checklist

#### General

#### Tests

### DIFF
--- a/src/braket/circuits/observables.py
+++ b/src/braket/circuits/observables.py
@@ -431,7 +431,7 @@ Observable.register_observable(TensorProduct)
 class Sum(Observable):
     """Sum of observables"""
 
-    def __init__(self, observables: List[Observable]):
+    def __init__(self, observables: List[Observable], display_name: str = "Hamiltonian"):
         """
         Args:
             observables (List[Observable]): List of observables for Sum
@@ -451,8 +451,7 @@ class Sum(Observable):
                 flattened_observables.append(obs)
 
         self._summands = tuple(flattened_observables)
-        qubit_count = sum(obs.qubit_count for obs in flattened_observables)
-        display_name = "+".join([obs.ascii_symbols[0] for obs in flattened_observables])
+        qubit_count = max(flattened_observables, key=lambda obs: obs.qubit_count).qubit_count
         super().__init__(qubit_count=qubit_count, ascii_symbols=[display_name] * qubit_count)
 
     @property
--- a/test/unit_tests/braket/circuits/test_observable.py
+++ b/test/unit_tests/braket/circuits/test_observable.py
@@ -212,5 +212,5 @@ def test_sum_observable_with_subtraction():
     obs2 = -4 * Observable.Y()
     result = obs1 - obs2
     assert isinstance(result, Observable.Sum)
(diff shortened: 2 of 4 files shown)


src/braket/circuits/observables.py
```diff
@@ -431,7 +431,7 @@ def _compute_eigenvalues(observables: Tuple[Observable], num_qubits: int) -> np.
 class Sum(Observable):
     """Sum of observables"""
 
-    def __init__(self, observables: List[Observable]):
+    def __init__(self, observables: List[Observable], display_name: str = "Hamiltonian"):
```
christianbmadsen, line 434: Should'nt the display name be "Observable", "O" or "..."? Is it guaranteed to be a Hamiltonian?

ajberdy: Yes. This constructor is only called in our code when someone uses the `+` operator between two terms, so it is almost certainly a multi-term Hamiltonian. The edge case exception being if someone manually constructs a Sum observable with a list of length one, which would make it a single-term Hamiltonian (unlikely and still technically correct).

test/unit_tests/braket/circuits/test_observable.py
```diff
@@ -212,5 +212,5 @@ def test_sum_observable_with_subtraction():
     obs2 = -4 * Observable.Y()
     result = obs1 - obs2
     assert isinstance(result, Observable.Sum)
-    assert result.qubit_count == 2
-    assert result.ascii_symbols == ("6X+4Y", "6X+4Y")
+    assert result.qubit_count == 1
+    assert result.ascii_symbols == ("6X+4Y",)
```
kshyatt-aws, line 216: Can we add a test (unit or integ) using an actual large Hamiltonian, so that if a similar bug recurs, the tests will fail/timeout?

christianbmadsen (approved):

test/unit_tests/braket/circuits/test_observable.py
ajberdy: Hesitant to add purposefully long-running tests to our integ tests since that increases friction and costs. This would be a great candidate for continuous benchmarking and I believe we should add that as a follow-up

src/braket/circuits/observables.py
rmshaffer: I think that whether it is a "Hamiltonian" or not is a problem-specific detail that is not known here? Often it will indeed be a Hamiltonian in the case where the observable is measuring energy. But in general the observable could be measuring anything.

ajberdy: I believe it's a Hamiltonian in the sense that it is literally the sum of observables being measured. That may directly translate to an energy Hamiltonian of a simulated system or some other problem-specific construct, but the actual quantity being measured is a Hamiltonian from the physical system created in the QPU. At least, that's my current understanding.

PR #508, opened 2023-02-24
feat: Introduce AHS-related utils from examples repo

ykharkov:
This is a probe PR.
Do not accept this PR yet. Please check the code style of unittest files.

*Description of changes:*
Added new methods to AHS classes.

DrivingField:
 -  DrivingField.from_lists(times: Iterable[Number], amplitude_values: Iterable[Number], detuning_values: Iterable[Number], phase_values: Iterable[Number]) -> DrivingField
 - DrivingField().concatenate(other: DrivingField) -> DrivingField
 - DrivingField.concatenate(driving_fields: List[DrivingField]) -> DrivingField

TimeSeries:
 - TimeSeries.constant_like(other: TimeSeries, value: Number) -> TimeSeries
 - TimeSeries().concatenate(other: TimeSeries) -> TimeSeries
 
ShiftingField:
 - ShiftingField().concatenate(other: ShiftingField) -> ShiftingField
 - ShiftingField.concatenate(shifting_fields: List[ShiftingField]) -> ShiftingField

Task results postprocessing:
 - AnalogHamiltonianSimulationQuantumTaskResult().marginal_probabilities() -> Dict[str, np.ndarray]
 - AnalogHamiltonianSimulationQuantumTaskResult().counts() -> Dict[str, int]

*Testing done:*

## Merge Checklist

#### General

#### Tests


src/braket/tasks/analog_hamiltonian_simulation_quantum_task_result.py
```diff
@@ -98,6 +99,49 @@ def _get_measurements(cls, result: AnalogHamiltonianSimulationTaskResult) -> Lis
...
+            r: Rydberg state atom
+            g: ground state atom
+        """
+
+        state_counts = Counter()
+        states = ["e", "r", "g"]
+        for shot in result.measurements:
+            pre = shot.pre_sequence
+            post = shot.post_sequence
+            state_idx = np.array(pre) * (1 + np.array(post))
```
krneta, line 121: Why do we need to convert them to numpy arrays? Maybe I'm misunderstanding the algorithm, but wouldn't it be easier to use a defaultdict?

src/braket/timings/time_series.py
```diff
@@ -87,6 +88,72 @@ def _ensure_sorted(self) -> None:
             self._series = OrderedDict(sorted(self._series.items()))
             self._sorted = True
 
+    @staticmethod
+    def constant_like(other_time_series: TimeSeries, constant: float = 0.0) -> TimeSeries:
```
krneta, line 92: Maybe since the values from the other time series aren't used, I think it would be better to pass in a list of times instead. 

Since we're at it, maybe we should have a function that takes a list of times and values and a default value? That way you can use it in concatenate too? It seems like that would be more general. Do you think it's a good idea?

src/braket/timings/time_series.py
```diff
@@ -87,6 +88,72 @@ def _ensure_sorted(self) -> None:
...
+        new_time_series = TimeSeries()
+        new_times = self.times() + other.times()
+        new_values = self.values() + other.values()
+        for t, v in zip(new_times, new_values):
+            new_time_series.put(t, v)
+
+        return new_time_series
+
+    @staticmethod
+    def rabi_pulse(
```
krneta, line 124: I'm not sure if this belongs in this class.

src/braket/timings/time_series.py
```diff
@@ -87,6 +88,72 @@ def _ensure_sorted(self) -> None:
...
+        rabi_pulse_area: float, omega_max: float, omega_slew_rate_max: float
+    ) -> Tuple[List[float], List[float]]:
+        """Get a time series for Rabi frequency with specified Rabi phase, maximum amplitude
+        and maximum slew rate
+            Args:
+                rabi_pulse_area (float): Total area under the Rabi frequency time series
+                omega_max (float): The maximum amplitude
+                omega_slew_rate_max (float): The maximum slew rate
+            Returns:
```
krneta, line 133: Doesn't it return a TimeSeries? There seems to be a mismatch between the declared return type hint and the actual return type. I'm not sure why it wasn't caught by the IDE.

src/braket/ahs/driving_field.py
```diff
@@ -83,6 +83,67 @@ def detuning(self) -> Field:
         r"""Field: global detuning (:math:`\Delta(t)`). Time is in s, and value is in rad/s."""
         return self._detuning
 
+    @staticmethod
```
krneta, line 86: Nit: Maybe we should keep all the static methods together in the file?

test/unit_tests/braket/ahs/test_driving_field.py
```diff
@@ -107,6 +107,111 @@ def test_discretize():
...
+    assert dr_field.phase.time_series.values() == phases
+
+    assert dr_field.amplitude.time_series.times() == times
+    assert dr_field.detuning.time_series.times() == times
+    assert dr_field.phase.time_series.times() == times
+
+
+def test_concatenate():
+    dr_field_1 = DrivingField.from_lists(
+        times=[0, 0.1, 0.2],
```
krneta, line 128: I would suggest you structure this test the way you did the one below, so you're testing the whole lists and not just the first/last elements.

test/unit_tests/braket/ahs/test_driving_field.py
```diff
@@ -107,6 +107,111 @@ def test_discretize():
...
+    times_1 = dr_field_1.amplitude.time_series.times()
+    times_2 = dr_field_2.amplitude.time_series.times()
+    times_3 = dr_field_3.amplitude.time_series.times()
+
+    assert len(new_times) == len(times_1) + len(times_2) + len(times_3)
+    assert new_dr.amplitude.time_series.values() == amplitudes_1 + amplitudes_2 + amplitudes_3
+    assert new_dr.detuning.time_series.values() == detunings_1 + detunings_2 + detunings_3
+    assert new_dr.phase.time_series.values() == phases_1 + phases_2 + phases_3
+
+
```
krneta, line 214: Don't forget to test your assert cases (eg. where the lists aren't all the same length).

src/braket/timings/time_series.py
ykharkov: I agree that it looks a bit akward in TimeSeries class. Moved to DrivingField since rubi_pulse function defines Omega(t), seem to be more logical to put it there

src/braket/timings/time_series.py
ykharkov: Typo, fixed

src/braket/timings/time_series.py
ykharkov: Yeah, I modified the interface. Now creating TimeSeries from list of time points and a constant value, this feels more natural

src/braket/tasks/analog_hamiltonian_simulation_quantum_task_result.py
ykharkov: Modified code using List comprehension instead of numpy array, perhaps now the code is more readable

src/braket/ahs/driving_field.py
```diff
@@ -83,6 +83,67 @@ def detuning(self) -> Field:
...
+        of Hamiltonian parameters (Rabi frequency, detuning, phase).
+        The values of the parameters at each time points are global for all atoms.
+
+        Args:
+            times (List[float]): The time points of the driving field
+            amplitudes (List[float]): The values of the amplitude
+            detunings (List[float]): The values of the detuning
+            phases (List[float]): The values of the phase
+        """
+        assert len(times) == len(amplitudes)
```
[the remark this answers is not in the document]
ykharkov: Thanks, good points. Added Error message.

src/braket/ahs/driving_field.py
```diff
@@ -83,6 +83,67 @@ def detuning(self) -> Field:
...
+            DrivingField: The concatenated driving field
+        """
+        return DrivingField(
+            amplitude=self.amplitude.time_series.concatenate(other.amplitude.time_series),
+            detuning=self.detuning.time_series.concatenate(other.detuning.time_series),
+            phase=self.phase.time_series.concatenate(other.phase.time_series),
+        )
+
+    @staticmethod
+    def concatenate_list(driving_fields: List[DrivingField]) -> DrivingField:
```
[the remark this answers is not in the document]
ykharkov: The base class is Hamiltonian. I don't think it's possible to move concatenate_list for DrivingField and ShiftingField to Hamiltonian class.

test/unit_tests/braket/ahs/test_driving_field.py
ykharkov: Agree, updated

src/braket/ahs/driving_field.py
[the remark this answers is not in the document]
krneta: We could add a new class between them. But I would understand if we don't want to do that yet.

src/braket/ahs/driving_field.py
```diff
@@ -83,6 +84,68 @@ def detuning(self) -> Field:
...
+        """Concatenate a list of driving fields to a single driving field
+        Args:
+            driving_fields (List[DrivingField]):
+            The list of driving field time series to be concatenated
+        Returns:
+            DrivingField: The concatenated driving field
+        """
+        drive = driving_fields[0]
```
krneta, line 144: We should protect against driving_fields being an empty array, or we should call out in the documentation that we'll throw an error if it is empty.

src/braket/tasks/analog_hamiltonian_simulation_quantum_task_result.py
```diff
@@ -98,6 +99,63 @@ def _get_measurements(cls, result: AnalogHamiltonianSimulationTaskResult) -> Lis
...
+            Shot result '1' corresponds to the ground state and does not contribute to the density.
+        """
+
+        postSeqs = []
+        for shot in result.measurements:
+            if shot.status != AnalogHamiltonianSimulationShotStatus.SUCCESS:
+                raise RuntimeWarning("Shot status: {}. Skipping.".format(shot.status))
+            else:
+                postSeqs.append(shot.post_sequence)
+        print("postSeqs", postSeqs)
```
krneta, line 154: remove print

src/braket/tasks/analog_hamiltonian_simulation_quantum_task_result.py
```diff
@@ -98,6 +99,63 @@ def _get_measurements(cls, result: AnalogHamiltonianSimulationTaskResult) -> Lis
...
+            r: Rydberg state atom
+            g: ground state atom
+        """
+
+        state_counts = Counter()
+        states = ["e", "r", "g"]
+        for shot in result.measurements:
+            if shot.status != AnalogHamiltonianSimulationShotStatus.SUCCESS:
+                state = None
+                raise RuntimeWarning("Shot status: {}. Skipping.".format(shot.status))
```
krneta, line 122: I'm not sure if a RuntimeWarning is the right thing here. Also you can use f"{varname}"

src/braket/tasks/analog_hamiltonian_simulation_quantum_task_result.py
```diff
@@ -98,6 +99,63 @@ def _get_measurements(cls, result: AnalogHamiltonianSimulationTaskResult) -> Lis
...
+            ndarray (float): The average densities from the result
+        Notes:
+            Shot result '0' corresponds to the Rydberg state (r) and contributes to the density.
+            Shot result '1' corresponds to the ground state and does not contribute to the density.
+        """
+
+        postSeqs = []
+        for shot in result.measurements:
+            if shot.status != AnalogHamiltonianSimulationShotStatus.SUCCESS:
+                raise RuntimeWarning("Shot status: {}. Skipping.".format(shot.status))
```
krneta, line 151: I'm not sure if a RuntimeWarning is the right thing here. Also you can use f"{varname}"

src/braket/timings/time_series.py
```diff
@@ -87,6 +87,38 @@ def _ensure_sorted(self) -> None:
...
+    def concatenate(self, other: TimeSeries) -> TimeSeries:
+        """Concatenate two time series to a single time series
+        Args:
+            other (TimeSeries): The second time series to be concatenated
+        Returns:
+            TimeSeries: The concatenated time series.
+            The time points in the second time series are shifted
+            by the total duration of the first time series.
+        """
+        assert other.times()[0] > self.times()[-1]
```
krneta, line 112: Assert message.

src/braket/timings/time_series.py
```diff
@@ -107,6 +139,28 @@ def discretize(self, time_resolution: Decimal, value_resolution: Decimal) -> Tim
...
+            TimeSeries: A new periodic time series.
+        """
+
+        new_time_series = TimeSeries()
+
+        num_points_new = len(values) * num_repeat
+        new_times = [i * dt for i in range(num_points_new)]
+        new_values = values * num_repeat
+        for t, v in zip(new_times, new_values):
+            new_time_series.put(t, v)
```
krneta, line 160: How about:
```
num_values = len(values)
for index in range(num_values * num_repeat)
  value = values[index % num_values]
  new_time_series.put(index * dt, value)
```

src/braket/tasks/analog_hamiltonian_simulation_quantum_task_result.py
```diff
@@ -98,6 +99,49 @@ def _get_measurements(cls, result: AnalogHamiltonianSimulationTaskResult) -> Lis
...
+            Dict[str, int]: number of times each state configuration is measured
+        Notes: We use the following convention to denote the state of an atom (site):
+            e: empty site
+            r: Rydberg state atom
+            g: ground state atom
+        """
+
+        state_counts = Counter()
+        states = ["e", "r", "g"]
+        for shot in result.measurements:
```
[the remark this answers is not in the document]
ykharkov: added status check, ignoring shots with partial/failed measurements

src/braket/timings/time_series.py
ykharkov: fixed

src/braket/ahs/driving_field.py
```diff
@@ -83,6 +84,72 @@ def detuning(self) -> Field:
...
+        """Concatenate a list of driving fields to a single driving field
+        Args:
+            driving_fields (List[DrivingField]):
+            The list of driving field time series to be concatenated
+        Returns:
+            DrivingField: The concatenated driving field
+        """
+
+        if len(driving_fields) == 0:
+            raise ValueError("The list must not be empty.")
```
krneta, line 146: Rather than "the list" please say "driving_fields parameter", and please add documentation in the args of the function that lets the caller know an empty list is not a valid parameter, and raises documentation that explains what we raise.

src/braket/ahs/shifting_field.py
```diff
@@ -59,6 +61,56 @@ def magnitude(self) -> Field:
...
+    @staticmethod
+    def concatenate_list(shift_fields: List[ShiftingField]) -> ShiftingField:
+        """Concatenate a list of shifting fields to a single driving field
+        Args:
+            shift_fields (List[ShiftingField]): The list of shifting fields to be concatenated
+        Returns:
+            ShiftingField: The concatenated shifting field
+        """
+        if len(shift_fields) == 0:
+            raise ValueError("The list must not be empty.")
```
krneta, line 107: As above

src/braket/ahs/driving_field.py
```diff
@@ -110,3 +173,36 @@ def discretize(self, properties: DiscretizationProperties) -> DrivingField:
         return DrivingField(
             amplitude=discretized_amplitude, phase=discretized_phase, detuning=discretized_detuning
         )
+
+    @staticmethod
+    def rabi_pulse(
```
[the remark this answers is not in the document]
ykharkov: The rabi_pulse is only useful to define DrivingField, this is why its here

src/braket/ahs/shifting_field.py
```diff
@@ -59,6 +61,58 @@ def magnitude(self) -> Field:
...
+        Args:
+            shift_fields (List[ShiftingField]): The list of shifting fields to be concatenated
+        Returns:
+            ShiftingField: The concatenated shifting field.
+            For the empty input list returns empty ShiftingField object.
+        """
+        if len(shift_fields) == 0:
+            return ShiftingField(magnitude=TimeSeries())
+
+        shift = shift_fields[0]
```
krneta, line 91: Please create a copy here. We shouldn't return a non-copy when len(shift_fields) == 1 and a copy otherwise.

src/braket/ahs/driving_field.py
```diff
@@ -110,3 +124,90 @@ def discretize(self, properties: DiscretizationProperties) -> DrivingField:
...
+            The list of driving field time series to be concatenated
+        Returns:
+            DrivingField: The concatenated driving field
+            For the empty input list returns empty DrivingField object.
+        """
+
+        if len(driving_fields) == 0:
+            return DrivingField(amplitude=TimeSeries(), detuning=TimeSeries(), phase=TimeSeries())
+
+        drive = driving_fields[0]
```
krneta, line 177: Please create a copy here.

src/braket/tasks/analog_hamiltonian_simulation_quantum_task_result.py
ykharkov: using logging now instead, thanks

src/braket/tasks/analog_hamiltonian_simulation_quantum_task_result.py
ykharkov: using logging.warning now instead

src/braket/ahs/driving_field.py
ykharkov: returning empty object instead

src/braket/ahs/shifting_field.py
ykharkov: returning empty object instead

src/braket/ahs/shifting_field.py
ykharkov: fixed

src/braket/ahs/driving_field.py
ykharkov: fixed

src/braket/ahs/shifting_field.py
```diff
@@ -59,6 +61,57 @@ def magnitude(self) -> Field:
...
         return self._magnitude
 
+    def concatenate(self, other: ShiftingField) -> ShiftingField:
+        """Concatenate two driving fields to a single driving field.
+        Assumes that the spatial modulation pattern is the same for the both driving fields.
+            Args:
+                other (ShiftingField): The second shifting field to be concatenated
+            Returns:
+                ShiftingField: The concatenated shifting field
+        """
```
peterkomar-aws, line 71:
```suggestion
            Raises:
                ValueError: if the patterns of the two shifting fields are not identical.
        """
```

src/braket/ahs/shifting_field.py
```diff
@@ -59,6 +61,57 @@ def magnitude(self) -> Field:
...
+        Args:
+            shift_fields (List[ShiftingField]): The list of shifting fields to be concatenated
+        Returns:
+            ShiftingField: The concatenated shifting field.
+            For the empty input list returns empty ShiftingField object.
+        """
+        if len(shift_fields) == 0:
+            return ShiftingField(magnitude=TimeSeries())
+
+        shift = ShiftingField(magnitude=shift_fields[0].magnitude)
```
peterkomar-aws, line 90: Again, if `ShiftingField(magnitude=TimeSeries()).concatenate(sf)` works, you can simplify this part of the code.

peterkomar-aws (changes requested):
Thank you. Great first version. Let's discuss.

src/braket/timings/time_series.py
```diff
@@ -87,6 +87,100 @@ def _ensure_sorted(self) -> None:
...
+                * "mean" - take the average of the boundary value points of the first
+                and the second time series.
+                * "left" - use the last value from the left time series as the boundary point.
+                * "right" - use the first value from the right time series as the boundary point.
+        Returns:
+            TimeSeries: The merged time series.
+        """
+
+        new_time_series = TimeSeries()
+        other_times = [t + gap_t + self.times()[-1] for t in other.times()]
```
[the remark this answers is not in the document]
ykharkov: agree, in the case when `gap_t=0` and `gap_t>0` the expected behaviour and the length of the output TimeSeries should be different. This looks a bit too complicated, so I suggest to remove the padding option altogether (set it to zero).
If needed, a User can easily `stitch` a constant-valued time series to realize the padding behaviour.

src/braket/timings/time_series.py
```diff
@@ -107,6 +201,27 @@ def discretize(self, time_resolution: Decimal, value_resolution: Decimal) -> Tim
             )
         return discretized_ts
 
+    @staticmethod
+    def periodic_signal(values: List[float], dt: float, num_repeat: int = 1):
```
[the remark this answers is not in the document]
ykharkov: changed to a List input

src/braket/ahs/shifting_field.py
ykharkov: thanks, I made changes

src/braket/tasks/analog_hamiltonian_simulation_quantum_task_result.py
```diff
@@ -98,6 +99,54 @@ def _get_measurements(cls, result: AnalogHamiltonianSimulationTaskResult) -> Lis
...
+
+        postSeqs = []
+        for shot in self.measurements:
+            if shot.status != AnalogHamiltonianSimulationShotStatus.SUCCESS:
+                pass
+            else:
+                postSeqs.append(shot.post_sequence)
+        if postSeqs == []:
+            raise ValueError("The list of measurements is empty. Average desnity is not defined.")
+        avg_density = np.sum(1 - np.array(postSeqs), axis=0) / len(postSeqs)
```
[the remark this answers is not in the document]
ykharkov: rewrote `get_avg_density` function, I think it looks more elegant now

src/braket/ahs/driving_field.py
```diff
@@ -110,3 +123,61 @@ def discretize(self, properties: DiscretizationProperties) -> DrivingField:
...
+        """
+
+        if len(driving_fields) == 0:
+            return DrivingField(amplitude=TimeSeries(), detuning=TimeSeries(), phase=TimeSeries())
+
+        amplitude = driving_fields[0].amplitude
+        detuning = driving_fields[0].detuning
+        phase = driving_fields[0].phase
+
+        drive = DrivingField(amplitude=amplitude, detuning=detuning, phase=phase)
```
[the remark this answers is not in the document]
ykharkov: made changes

src/braket/tasks/analog_hamiltonian_simulation_quantum_task_result.py
```diff
@@ -98,6 +99,54 @@ def _get_measurements(cls, result: AnalogHamiltonianSimulationTaskResult) -> Lis
...
+                state = "".join(map(lambda s_idx: states[s_idx], state_idx))
+                state_counts.update((state,))
+
+        return dict(state_counts)
+
+    def get_avg_density(self) -> np.ndarray:
+        """Get the average Rydberg state densities from the result
+        Returns:
+            ndarray (float): The average densities from the result
+        Notes:
```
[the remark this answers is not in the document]
ykharkov: removed

peterkomar-aws (commented):
Looks good. One comment on the concatenate - stitch axis:
* Currently, `concatenate` is used by the concatenate methods of DrivingField and ShiftingField.
* Do you think having a `stitch` method for DrivingField and ShiftingField classes would be helpful too?
* Once we are thinking about it, can we think about if it's enough to have one of them, either `stitch` or `concatenate`?

ykharkov:
Slight change of scope after discussion with @peterkomar-aws . In the last 2 commits I removed `concatenate` and `concatenate_list` from ShiftingField and DrivingField classes, but keep `concatenate` in base TimeSeries class.

Currently, the essential functionality pertaining to merging DrivingField and ShiftingField is realized in `.stitch` method, which causes some redundancy. `Stitch` method is somewhat less trivial, compared to `concatenate` and is very useful for users. In order to keep the SDK functionality frugal we decided to keep only `.stitch` method.

Having said that, `concatenate` and `concatenate_list` for Fields could be easily realized by the User from `TimeSeries.concatenate` method, so we don't loose much.

peterkomar-aws (dismissed):

src/braket/ahs/driving_field.py
```diff
@@ -83,6 +83,27 @@ def detuning(self) -> Field:
...
+        """Stitches two driving fields based on TimeSeries.stitch method.
+        Shifts time points in the second DrivingField to align with the first DrivingField.
+        Args:
+            other (DrivingField): The second shifting field to be stitched
+            boundary (str): {"mean", "left", "right"}. Boundary point handler.
+            Possible options are
+                * "mean" - take the average of the boundary value points of the first
+                and the second time series.
+                * "left" - use the last value from the left time series as the boundary point.
+                * "right" - use the first value from the right time series as the boundary point.
```
speller26, line 96: Argument docs need to indent one more level after the first line for the docs to display correctly
```suggestion
                Possible options are
                    * "mean" - take the average of the boundary value points of the first
                    and the second time series.
                    * "left" - use the last value from the left time series as the boundary point.
                    * "right" - use the first value from the right time series as the boundary point.
```
Also, might be worth considering using an `Enum` instead of a raw string for this, since there are only three allowed values.

src/braket/ahs/shifting_field.py
```diff
@@ -59,6 +61,46 @@ def magnitude(self) -> Field:
...
+        """Stitches two shifting fields based on TimeSeries.stitch method.
+        Shifts time points in the second ShiftingField to align with the first ShiftingField.
+        Args:
+            other (ShiftingField): The second shifting field to be stitched
+            boundary (str): {"mean", "left", "right"}. Boundary point handler.
+            Possible options are
+                * "mean" - take the average of the boundary value points of the first
+                and the second time series.
+                * "left" - use the last value from the left time series as the boundary point.
+                * "right" - use the first value from the right time series as the boundary point.
```
speller26, line 94: See comment in `DrivingField`
```suggestion
                Possible options are
                    * "mean" - take the average of the boundary value points of the first
                    and the second time series.
                    * "left" - use the last value from the left time series as the boundary point.
                    * "right" - use the first value from the right time series as the boundary point.
```

src/braket/timings/time_series.py
```diff
@@ -87,6 +87,111 @@ def _ensure_sorted(self) -> None:
...
+    def stitch(self, other: TimeSeries, boundary: str = "mean") -> TimeSeries:
+        """Stitch two time series to a single time series and shifts the time points accordingly.
+        Args:
+            other (TimeSeries): The second time series to be concatenated
+            boundary (str): {"mean", "left", "right"}. Boundary point handler.
+            Possible options are
+                * "mean" - take the average of the boundary value points of the first
+                and the second time series.
+                * "left" - use the last value from the left time series as the boundary point.
+                * "right" - use the first value from the right time series as the boundary point.
```
speller26, line 160: See suggestion in `DrivingField`
```suggestion
                Possible options are
                    * "mean" - take the average of the boundary value points of the first
                    and the second time series.
                    * "left" - use the last value from the left time series as the boundary point.
                    * "right" - use the first value from the right time series as the boundary point.
```

src/braket/tasks/analog_hamiltonian_simulation_quantum_task_result.py
```diff
@@ -98,6 +99,54 @@ def _get_measurements(cls, result: AnalogHamiltonianSimulationTaskResult) -> Lis
...
+                continue
+            else:
+                pre = shot.pre_sequence
+                post = shot.post_sequence
+                # converting presequence and postsequence measurements to state_idx
+                state_idx = [
+                    0 if pre_i == 0 else 1 if post_i == 0 else 2 for pre_i, post_i in zip(pre, post)
+                ]
+                state = "".join(map(lambda s_idx: states[s_idx], state_idx))
+                state_counts.update((state,))
```
speller26, line 126: Instead of using `continue`, this reads cleaner as
```suggestion
            if shot.status == AnalogHamiltonianSimulationShotStatus.SUCCESS:
                pre = shot.pre_sequence
                post = shot.post_sequence
                # converting presequence and postsequence measurements to state_idx
                state_idx = [
                    0 if pre_i == 0 else 1 if post_i == 0 else 2 for pre_i, post_i in zip(pre, post)
                ]
                state = "".join(map(lambda s_idx: states[s_idx], state_idx))
                state_counts.update((state,))
```

src/braket/ahs/driving_field.py
```diff
@@ -83,6 +83,27 @@ def detuning(self) -> Field:
         r"""Field: global detuning (:math:`\Delta(t)`). Time is in s, and value is in rad/s."""
         return self._detuning
 
+    def stitch(self, other: DrivingField, boundary: str = "mean") -> DrivingField:
+        """Stitches two driving fields based on TimeSeries.stitch method.
+        Shifts time points in the second DrivingField to align with the first DrivingField.
+        Args:
```
speller26, line 89: Lack of spacing might be interfering with docs rendering correctly; this applies to the other methods and files too
```suggestion
        Shifts time points in the second DrivingField to align with the first DrivingField.

        Args:
```

yourball: ok, fixed

src/braket/tasks/analog_hamiltonian_simulation_quantum_task_result.py
yourball: thanks, fixed

src/braket/timings/time_series.py
yourball: thanks, fixed

src/braket/ahs/driving_field.py
```diff
@@ -83,6 +83,30 @@ def detuning(self) -> Field:
         r"""Field: global detuning (:math:`\Delta(t)`). Time is in s, and value is in rad/s."""
         return self._detuning
 
+    def stitch(self, other: DrivingField, boundary: str = "mean") -> DrivingField:
```
krneta, line 86: Should we make boundary an Enum instead?

src/braket/tasks/analog_hamiltonian_simulation_quantum_task_result.py
```diff
@@ -98,6 +99,55 @@ def _get_measurements(cls, result: AnalogHamiltonianSimulationTaskResult) -> Lis
             measurements.append(ShotResult(status, pre_sequence, post_sequence))
         return measurements
 
+    def get_counts(self) -> Dict[str, int]:
+        """Aggregate state counts from AHS shot results
+
+        Returns:
+            Dict[str, int]: number of times each state configuration is measured.
```
krneta, line 106: should we say somewhere in this doc that we'll only get counts for successful shots?

src/braket/tasks/analog_hamiltonian_simulation_quantum_task_result.py
```diff
@@ -98,6 +99,55 @@ def _get_measurements(cls, result: AnalogHamiltonianSimulationTaskResult) -> Lis
...
+    def get_avg_density(self) -> np.ndarray:
+        """Get the average Rydberg state densities from the result
+
+        Returns:
+            ndarray (float): The average densities from the result
+        """
+
+        counts = self.get_counts()
+
+        Nr, Ng = [], []
```
krneta, line 139: Can you make the variable names for descriptive?

src/braket/timings/time_series.py
```diff
@@ -87,6 +87,120 @@ def _ensure_sorted(self) -> None:
...
+        Returns:
+            TimeSeries: The concatenated time series.
+
+        Notes:
+            Keeps the time points in both time series unchanged.
+            Assumes that the time points in the first TimeSeries
+            are at earler times then the time points in the second TimeSeries.
+        """
+
+        if len(other.times()) * len(self.times()) == 0:
```
krneta, line 142: Have you tested when one time series is empty, but the other is not? 

Regardless, can you rewrite this so the "pass" isn't necessary?

src/braket/timings/time_series.py
```diff
@@ -87,6 +87,120 @@ def _ensure_sorted(self) -> None:
...
+        for t, v in zip(new_times, new_values):
+            new_time_series.put(t, v)
+
+        return new_time_series
+
+    def stitch(self, other: TimeSeries, boundary: str = "mean") -> TimeSeries:
+        """Stitch two time series to a single time series and shifts the time points accordingly.
+
+        Args:
+            other (TimeSeries): The second time series to be concatenated
```
krneta, line 162: Are they concatenated?

src/braket/timings/time_series.py
```diff
@@ -87,6 +87,120 @@ def _ensure_sorted(self) -> None:
...
+        new_time_series = TimeSeries()
+        new_times = self.times() + other.times()
+        new_values = self.values() + other.values()
+        for t, v in zip(new_times, new_values):
+            new_time_series.put(t, v)
+
+        return new_time_series
+
+    def stitch(self, other: TimeSeries, boundary: str = "mean") -> TimeSeries:
+        """Stitch two time series to a single time series and shifts the time points accordingly.
```
krneta, line 159: I don't think I understand what "accordingly" means here. What are we actually doing to the time points?

src/braket/timings/time_series.py
```diff
@@ -87,6 +87,120 @@ def _ensure_sorted(self) -> None:
...
+                    and the second time series.
+                    * "left" - use the last value from the left time series as the boundary point.
+                    * "right" - use the first value from the right time series as the boundary
+                    point.
+        Returns:
+            TimeSeries: The stitched time series.
+        """
+
+        if len(self.times()) == 0:
+            return TimeSeries.from_lists(times=other.times(), values=other.values())
```
krneta, line 175: Does this still make sense for "mean" and "left"?

test/unit_tests/braket/ahs/test_shifting_field.py
```diff
@@ -63,6 +63,64 @@ def test_iadd_to_other(default_shifting_field):
     assert other.terms == expected
 
 
+def test_from_lists():
+    times = [0, 0.1, 0.2, 0.3]
+    glob_amplitude = [0.5, 0.8, 0.9, 1.0]
+    pattern = [0.3, 0.7, 0.6, -0.5, 0, 1.6]
```
krneta, line 69: You may be able to simplify these tests by making these fixtures.

test/unit_tests/braket/timings/test_time_series.py
```diff
@@ -54,6 +54,123 @@ def test_get_sorted(default_values, default_time_series):
...
 
 
+def test_constant_like():
+    times = list(range(10))
+    constant_ts = TimeSeries.constant_like(times, constant=3.14)
+    assert times == constant_ts.times()
+    assert constant_ts.values() == [3.14] * len(times)
+
+
+def test_periodic_signal():
```
krneta, line 64: Please test with different "boundary" types. I suggest you use "@pytest.mark.parametrize"

test/unit_tests/braket/timings/test_time_series.py
```diff
@@ -54,6 +54,123 @@ def test_get_sorted(default_values, default_time_series):
...
+
+
+@pytest.mark.xfail(raises=ValueError)
+def test_from_lists_not_equal_size():
+    times = list(range(4))
+    values = [0.5, 1, 1]
+    TimeSeries.from_lists(times=times, values=values)
+
+
+def test_stitch():
```
krneta, line 129: As above, for these stitch methods, please test with different boundary types.

test/unit_tests/braket/timings/test_time_series.py
```diff
@@ -54,6 +54,123 @@ def test_get_sorted(default_values, default_time_series):
...
+    assert new_ts_mean.times() == excepted_times
+    assert new_ts_left.times() == excepted_times
+    assert new_ts_right.times() == excepted_times
+
+    assert new_ts_mean.values() == [0.5, 1, 1, -0.25, -1, -1, 0]
+    assert new_ts_left.values() == [0.5, 1, 1, 0, -1, -1, 0]
+    assert new_ts_right.values() == [0.5, 1, 1, -0.5, -1, -1, 0]
+
+
+def test_stitch_empty_ts():
```
krneta, line 152: Please test with different lists being zero length (and both being zero length).

For this, and the above tests, I suggest using "@pytest.mark.parametrize"

src/braket/timings/time_series.py
ykharkov: Yes, it makes sense. When the self.time_series is empty we simply return the second time series regardless of the boundary condition.  When we are stitching an empty and a non-empty time series, we are returning the non-empty one, since it is the only logical possibility.

test/unit_tests/braket/timings/test_time_series.py
ykharkov: Thanks, included additional tests with either one or both TimeSeries being empty.
I still prefer the inputs for the tests to be defined in each test function, makes it easier to follow the code.

src/braket/timings/time_series.py
ykharkov: I added more comprehensive explanation and added an example to docstring. This should help to explain how .stitch is different from .concatenate.

src/braket/timings/time_series.py
```diff
@@ -87,6 +87,120 @@ def _ensure_sorted(self) -> None:
...
+            other (TimeSeries): The second time series to be concatenated
+            boundary (str): {"mean", "left", "right"}. Boundary point handler.
+                Possible options are
+                    * "mean" - take the average of the boundary value points of the first
+                    and the second time series.
+                    * "left" - use the last value from the left time series as the boundary point.
+                    * "right" - use the first value from the right time series as the boundary
+                    point.
+        Returns:
+            TimeSeries: The stitched time series.
```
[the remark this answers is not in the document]
ykharkov: Concatenate is a bare bones method that simply concatenates lists for the times and values, respectively. 

On the other hand, stitch is more versatile, it shifts the time points in the second time series, so that the end of the first time series matches the start of the second series (stitched at the boundary point). See added "Example" in the docstring.
The main benefit of .stitch is that the user doesn't need to care about aligning the time points in both time series.

src/braket/timings/time_series.py
ykharkov: typo, fixed

src/braket/timings/time_series.py
```diff
@@ -87,6 +87,120 @@ def _ensure_sorted(self) -> None:
...
+
+        new_time_series = TimeSeries()
+        new_times = self.times() + other.times()
+        new_values = self.values() + other.values()
+        for t, v in zip(new_times, new_values):
+            new_time_series.put(t, v)
+
+        return new_time_series
+
+    def stitch(self, other: TimeSeries, boundary: str = "mean") -> TimeSeries:
```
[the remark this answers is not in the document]
ykharkov: Changed to Enum as requested

src/braket/timings/time_series.py
ykharkov: thanks, rewrote the code without "pass".

Also added more test functionality when wither one or both TimeSeries are empty

src/braket/tasks/analog_hamiltonian_simulation_quantum_task_result.py
ykharkov: modified, hopefully now its more clear (N_ryd - Rydberg states, N_ground - ground states)

src/braket/tasks/analog_hamiltonian_simulation_quantum_task_result.py
ykharkov: Ok, thanks, updated the docstring

src/braket/ahs/driving_field.py
ykharkov: updated, see below (TimeSeries)

peterkomar-aws (approved):

src/braket/timings/time_series.py
```diff
@@ -87,6 +94,164 @@ def _ensure_sorted(self) -> None:
...
+            return TimeSeries.from_lists(times=self.times(), values=self.values())
+
+        new_time_series = TimeSeries()
+        left_t, right_t = self.times()[-1], other.times()[0]
+        other_times = [t - right_t + left_t for t in other.times()]
+        new_times = self.times() + other_times[1:]
+
+        left, right = self.values()[-1], other.values()[0]
+        if boundary == StitchBoundaryCondition.MEAN:
+            bndry_val = 0.5 * sum([left, right])
```
krneta, line 237: nit: I'm not sure why you didn't just do (left + right)?

krneta (approved):

christianbmadsen (commented):
I assume we will update the relevant amazon-braket-examples content accordingly. Can you create an Issue in https://github.com/aws/amazon-braket-examples/issues/ to track that change once this has been merged? Thanks!

ykharkov:
> I assume we will update the relevant amazon-braket-examples content accordingly. Can you create an Issue in https://github.com/aws/amazon-braket-examples/issues/ to track that change once this has been merged? Thanks!

Opened an issue in examples repo https://github.com/aws/amazon-braket-examples/issues/270

christianbmadsen (approved):

Issue #523, opened 2023-03-30
Gate modifiers on local simulator through the BDK Circuit class


ajberdy:
**Describe the feature you'd like**
The local simulator supports using OpenQASM gate modifiers through its OpenQASM interface. I would like to be able to use gate modifiers from the BDK Circuit class when using the local simulator.

**How would this feature be used? Please describe.**
Users could modify gates in a circuit as it is constructed. For example:
```python
circ = Circuit()
circ.ry(target=2, angle=11/7, control=[0, 1])
cirx.x(target=0, power=1/3)
```

**Describe alternatives you've considered**
This is currently attainable by using an OpenQASM string to define the circuit.

**Additional context**
I will begin working on the implementation for this feature on the branch [`feature/gate-modifiers`](https://github.com/aws/amazon-braket-sdk-python/tree/feature/gate-modifiers)

christianbmadsen:
@ajberdy can this be closed based on #524 and #226 that are merged?

PR #524, opened 2023-03-30
feat: control modifier

ajberdy:
*Issue #, if available:*

#523

*Description of changes:*

Initial commit to add support for control modifiers

*Testing done:*

Unit testing and manual testing

## Merge Checklist

#### General

#### Tests

### DIFF
--- a/src/braket/circuits/gate.py
+++ b/src/braket/circuits/gate.py
@@ -46,6 +46,7 @@ class Gate(QuantumOperator):
             ValueError: `qubit_count` is less than 1, `ascii_symbols` are `None`, or
                 `ascii_symbols` length != `qubit_count`
         """
+        # todo: implement ascii symbols for control modifier
         super().__init__(qubit_count=qubit_count, ascii_symbols=ascii_symbols)
 
     @property
--- a/src/braket/circuits/gate.py
+++ b/src/braket/circuits/gate.py
@@ -48,6 +48,10 @@ class Gate(QuantumOperator):
         """
         super().__init__(qubit_count=qubit_count, ascii_symbols=ascii_symbols)
 
+    @property
+    def _qasm_name(self):
+        raise NotImplementedError()
+
     def adjoint(self) -> List[Gate]:
         """Returns a list of gates that implement the adjoint of this gate.
 
@@ -63,6 +67,8 @@ class Gate(QuantumOperator):
         target: QubitSet,
         ir_type: IRType = IRType.JAQCD,
         serialization_properties: SerializationProperties = None,
+        *,
+        control: Optional[QubitSet] = None,
     ) -> Any:
         """Returns IR object of quantum operator and target
 
@@ -73,14 +79,18 @@ class Gate(QuantumOperator):
             serialization_properties (SerializationProperties): The serialization properties to use
                 while serializing the object to the IR representation. The serialization properties
                 supplied must correspond to the supplied `ir_type`. Defaults to None.
+            control (Optional[QubitSet]): control qubit(s). Only supported for OpenQASM.
         Returns:
             Any: IR object of the quantum operator and target
 
         Raises:
             ValueError: If the supplied `ir_type` is not supported, or if the supplied serialization
             properties don't correspond to the `ir_type`.
+            ValueError: If gate modifiers are supplied with `ir_type` Jaqcd.
         """
         if ir_type == IRType.JAQCD:
+            if control:
+                raise ValueError("Gate modifiers are not supported with Jaqcd.")
             return self._to_jaqcd(target)
         elif ir_type == IRType.OPENQASM:
             if serialization_properties and not isinstance(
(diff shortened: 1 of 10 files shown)


src/braket/circuits/instruction.py
```diff
@@ -168,14 +206,29 @@ def copy(
         if target_mapping and target is not None:
             raise TypeError("Only 'target_mapping' or 'target' can be supplied, but not both.")
         elif target is not None:
-            return Instruction(self._operator, target)
+            new_target = target
         else:
-            return Instruction(self._operator, self._target.map(target_mapping or {}))
+            new_target = self._target.map(target_mapping or {})
+        if control_mapping and control is not None:
```
speller26, line 212: _super_ nit: extra space for readability
```suggestion
            new_target = self._target.map(target_mapping or {})

        if control_mapping and control is not None:
```

src/braket/circuits/instruction.py
```diff
@@ -168,14 +206,29 @@ def copy(
         if target_mapping and target is not None:
             raise TypeError("Only 'target_mapping' or 'target' can be supplied, but not both.")
         elif target is not None:
-            return Instruction(self._operator, target)
+            new_target = target
         else:
-            return Instruction(self._operator, self._target.map(target_mapping or {}))
+            new_target = self._target.map(target_mapping or {})
```
speller26, line 211: Completely optional, but since this is now an assignment instead of a return statement, how about just assigning the variable without branching?
```suggestion
        new_target = target if target is not None else self._target.map(target_mapping or {})
```
In general, it's just easier for me to read code like this rather than making sure an assignment happens in all branches

src/braket/circuits/instruction.py
```diff
@@ -168,14 +206,29 @@ def copy(
...
+        if control_mapping and control is not None:
+            raise TypeError("Only 'control_mapping' or 'control' can be supplied, but not both.")
+        elif control is not None:
+            new_control = control
+        else:
+            new_control = self._control.map(control_mapping or {})
```
speller26, line 217: Same nit:
```suggestion
        new_control = control if control is not None else self._control.map(control_mapping or {})
```

src/braket/circuits/gates.py
```diff
@@ -1954,21 +1979,27 @@ def bind_values(self, **kwargs) -> GPi:
     @staticmethod
     @circuit.subroutine(register=True)
     def gpi(
-        target: QubitInput, angle: Union[FreeParameterExpression, float]
+        target: QubitSetInput,
```
speller26, line 1982: Why not `QubitInput` if this is a 1q gate?

src/braket/circuits/gates.py
```diff
@@ -857,7 +879,8 @@ def cnot(control: QubitInput, target: QubitInput) -> Instruction:
         Examples:
             >>> circ = Circuit().cnot(0, 1)
         """
-        return Instruction(CNot(), target=[control, target])
+        control_qubits = QubitSet(control)
+        return Instruction(CNot(), target=[control_qubits.pop(), target], control=control_qubits)
```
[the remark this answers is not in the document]
ajberdy: Lol fair enough.. alternatively I could make control a keyword-only arg for Instruction

src/braket/circuits/gates.py
ajberdy: Our single-qubit gates are 50/50 on whether they broadcast over multiple qubits when provided with more than one. If you look at the first few gates, you'll find they all do. According to the implementation, so does GPi. Some gates differ from this behavior and I think we should make it consistent; not convinced it has to happen this PR though

speller26: Oh, good catch!

src/braket/circuits/gate.py
```diff
@@ -46,8 +46,13 @@ def __init__(self, qubit_count: Optional[int], ascii_symbols: Sequence[str]):
             ValueError: `qubit_count` is less than 1, `ascii_symbols` are `None`, or
                 `ascii_symbols` length != `qubit_count`
         """
+        # todo: implement ascii symbols for control modifier
         super().__init__(qubit_count=qubit_count, ascii_symbols=ascii_symbols)
 
+    @property
+    def _qasm_name(self):
```
speller26, line 53: Might actually be worth making this a public property, so users can see _what_ a gate gets serialized as; it's also easier to miss and get confused over when adding new gates.

Ideally, this would actually _be_ the `name` property.

src/braket/circuits/gates.py
```diff
@@ -1831,7 +1862,13 @@ def fixed_qubit_count() -> int:
 
     @staticmethod
     @circuit.subroutine(register=True)
-    def ccnot(control1: QubitInput, control2: QubitInput, target: QubitInput) -> Instruction:
+    def ccnot(
+        control1: QubitInput,
+        control2: QubitInput,
+        target: QubitInput,
+        *,
+        control: Optional[QubitSetInput] = None,
```
speller26, line 1870: Not sure if I like this this, but I also don't have a better solution off the top of my head.

speller26 (approved):
LGTM, just remember to finish the two TODOs

src/braket/circuits/gates.py
ajberdy: Yup same

src/braket/circuits/gate.py
ajberdy: Makes sense to me, I'll go ahead and do that in a follow-up pr

PR #539, opened 2023-04-28
test: add flaky tag

ajberdy:
*Issue #, if available:*

*Description of changes:*

Add example flaky tag to the most recent integ test to flake

*Testing done:*

Ran integ test (with extra probabilistic failure modes) and watched the reruns succeed, ran integ tests successfully with the latest revision

## Merge Checklist

#### General

#### Tests


test/integ_tests/test_simulator_quantum_task.py
```diff
@@ -148,6 +148,7 @@ def test_result_types_tensor_hermitian_hermitian(
     )
 
 
+@pytest.mark.flaky(reruns=3)
```
krneta, line 151: Do we have details about what makes it flaky? Can we change/simplify the test at all to fix it that way instead?

ajberdy: There's an assert comparing expected to actual variance of n shots [here](https://github.com/aws/amazon-braket-sdk-python/blob/0cf2f59fa24ab82031ece126a54fca5752cfbc4e/test/integ_tests/gate_model_device_testing_utils.py#L280). This isn't that big of an issue and I've only ever seen it fail once, but mostly I'm adding this to serve as an example of how we handle appropriately flaky tests. I agree of course that if a test fails too often, we should adjust it.

krneta: > I'm adding this to serve as an example of how we handle appropriately flaky tests

That's exactly what I'm worried about. I'm worried that people will start doing this instead of investigating and trying to fix the issues. So I think the least we could do is add some comments about why the test is flaky, what we tried to do to fix it, and why it can't be fixed easily so we're relying on the tag.

ajberdy: I think we find similar issues with seeding the tests, since people might just keep picking seeds until it passes instead of refining the test. Ideally, this tag would be used in conjunction with calibrating the tests to a 99+% success rate

shpface:
Flaky with 3 reruns seems way too permissive. Which of these actually failed? Looking at a couple, it seems like they should fail by random chance ~never (1 in 10^70). I'd rather see more shots, or an update to the `get_tol` function to make the tests not flaky.

ajberdy:
In this case, since the failure mode is probabilistic sampling (vs unreliable external system etc), adding retries is not fundamentally different from adjusting the test parameters. Opting for increasing shots over adding retries or lowering tolerance.

krneta (approved):

PR #575, opened 2023-06-15
feat: enum for device arns

ajberdy:
*Issue #, if available:*

*Description of changes:*

Autocomplete-compatible enum for device ARNs -- no more navigating to the AWS console just to copy and paste an ARN!

*Testing done:*

Manual and integ testing.

## Merge Checklist

#### General

#### Tests

### DIFF
--- a/src/braket/aws/aws_device.py
+++ b/src/braket/aws/aws_device.py
@@ -47,50 +47,6 @@ class AwsDeviceType(str, Enum):
     QPU = "QPU"
 
 
-class BraketDevices:
-    class _Amazon(str, Enum):
-        SV1 = "arn:aws:braket:::device/quantum-simulator/amazon/sv1"
-        TN1 = "arn:aws:braket:::device/quantum-simulator/amazon/tn1"
-        DM1 = "arn:aws:braket:::device/quantum-simulator/amazon/dm1"
-
-    class _DWave(str, Enum):
-        _Advantage1 = "arn:aws:braket:::device/qpu/d-wave/Advantage_system1"
-        _Advantage3 = "arn:aws:braket:::device/qpu/d-wave/Advantage_system3"
-        _Advantage4 = "arn:aws:braket:::device/qpu/d-wave/Advantage_system4"
-        _Advantage6 = "arn:aws:braket:us-west-2::device/qpu/d-wave/Advantage_system6"
-        _DW2000Q6 = "arn:aws:braket:::device/qpu/d-wave/DW_2000Q_6"
-
-    class _IonQ(str, Enum):
-        Harmony = "arn:aws:braket:us-east-1::device/qpu/ionq/Harmony"
-        Aria1 = "arn:aws:braket:us-east-1::device/qpu/ionq/Aria-1"
-
-    class _OQC(str, Enum):
-        Lucy = "arn:aws:braket:eu-west-2::device/qpu/oqc/Lucy"
-
-    class _Quera(str, Enum):
-        Aquila = "arn:aws:braket:us-east-1::device/qpu/quera/Aquila"
-
-    class _Rigetti(str, Enum):
-        _Aspen8 = "arn:aws:braket:::device/qpu/rigetti/Aspen-8"
-        _Aspen9 = "arn:aws:braket:::device/qpu/rigetti/Aspen-9"
-        _Aspen10 = "arn:aws:braket:::device/qpu/rigetti/Aspen-10"
-        _Aspen11 = "arn:aws:braket:::device/qpu/rigetti/Aspen-11"
-        _AspenM1 = "arn:aws:braket:us-west-1::device/qpu/rigetti/Aspen-M-1"
-        _AspenM2 = "arn:aws:braket:us-west-1::device/qpu/rigetti/Aspen-M-2"
-        AspenM3 = "arn:aws:braket:us-west-1::device/qpu/rigetti/Aspen-M-3"
-
-    class _Xanadu(str, Enum):
-        _Borealis = "arn:aws:braket:us-east-1::device/qpu/xanadu/Borealis"
-
-    Amazon = _Amazon
-    # DWave = _DWave
-    IonQ = _IonQ
-    OQC = _OQC
-    Quera = _Quera
-    Rigetti = _Rigetti
-    # Xanadu = _Xanadu
-
-
 class AwsDevice(Device):
     """
     Amazon Braket implementation of a device.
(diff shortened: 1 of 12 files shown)


src/braket/aws/aws_device.py
```diff
@@ -47,6 +47,50 @@ class AwsDeviceType(str, Enum):
     QPU = "QPU"
 
 
+class BraketDevices:
```
ajberdy, line 50: Considering putting this inside `AwsDevice` as `Devices` so users can just have a single import and write something like `AwsDevice(AwsDevice.Devices.Amazon.SV1`. Thoughts?

mdpratt, line 50: I think this is a nice QOL change, but this does mean that we'll need to manually maintain this enum. Is that scalable long term?

mdpratt: I think there is going to be some confusion regarding naming. When I see `AwsDevice(BraketDevices.OQC.Lucy)` is it an AWS device? A Braket Device? An  OQC device?

I think the enum should just be `Devices` to help make that clearer: `Devices.<Provider>.<Name>`.

ajberdy: Yes, it seems pretty light to maintain. I'll make sure to update the device onboarding template to include this update

ajberdy: Sure, down for `BraketDevices`->`Devices`.

My alternate suggestion would just be moving the enum to the `AwsDevice` class, but still wouldn't require instantiating the object to access. So same as currently proposed, `AwsDevice.Devices.Amazon.SV1` would just give the string ARN for SV1, pulled from the enum and without any service calls.

I don't see the need to break the `AwsDevice(arn)` pattern of instantiating devices here

ajberdy: Another benefit of putting it in the `AwsDevice` class is that it resolves some of the scope ambiguity. Technically the local simulator is a Braket "Device", so `AwsDevice.Devices.<etc>` may be a clearer indication of which devices are considered.

As a counterpoint, maybe we should leave it outside of AwsDevice and add `local:braket/default` as the ARN for the local simulator (as prescribed by the jobs interface), but I'm less convinced about the utility of that

src/braket/aws/aws_device.py
```diff
@@ -47,6 +47,50 @@ class AwsDeviceType(str, Enum):
...
+        _Aspen8 = "arn:aws:braket:::device/qpu/rigetti/Aspen-8"
+        _Aspen9 = "arn:aws:braket:::device/qpu/rigetti/Aspen-9"
+        _Aspen10 = "arn:aws:braket:::device/qpu/rigetti/Aspen-10"
+        _Aspen11 = "arn:aws:braket:::device/qpu/rigetti/Aspen-11"
+        _AspenM1 = "arn:aws:braket:us-west-1::device/qpu/rigetti/Aspen-M-1"
+        _AspenM2 = "arn:aws:braket:us-west-1::device/qpu/rigetti/Aspen-M-2"
+        AspenM3 = "arn:aws:braket:us-west-1::device/qpu/rigetti/Aspen-M-3"
+
+    class _Xanadu(str, Enum):
+        _Borealis = "arn:aws:braket:us-east-1::device/qpu/xanadu/Borealis"
```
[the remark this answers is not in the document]
ajberdy: We still list retired devices from our API, so presumable there is some context where a customer may want to get the ARN. Retired QPUs and Devices are prefixed with a `_`, so they will not show up in the IDE autocomplete

src/braket/aws/aws_device.py
shpface: Can we create this device list dynamically? What do you think of something like `Devices = Enum('Devices',[(device.name, device.arn) for device in AwsDevice.get_devices()])`?

src/braket/aws/aws_device.py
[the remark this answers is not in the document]
guomanmin: Thanks for the explanation! Resolving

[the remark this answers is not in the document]
guomanmin: In hindsight, pretty obvious pattern there, sorry for the channel vision :P

src/braket/aws/aws_device.py
ajberdy: I bet we could, but idk how I feel about API calls at import time, plus I don't think that would enable IDE autocomplete.

The other complexity is that not all "device names" are valid python variable names (multiple words, dashes, etc.), so we'd need to have some kind of mapping anyways (could be done by regex, but this seems like something we want manual quality control over)

src/braket/aws/aws_device.py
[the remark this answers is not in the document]
ajberdy: Easy to miss, thanks for reviewing!

src/braket/aws/aws_device.py
zmohammad01: +1 to Mikes point - `Devices.<Provider>.<Name>` 

Devices without the AWS or Braket as the enum allows us additional flexibility as we scale up in the future - slack me and I can elaborate.

ajberdy: Sounds good to me!

speller26: What about `DeviceIdentifiers`? I like that `Devices` is short and sweet, but the class just contains the ARNs with no additional functionality.

speller26: I think it's fine to have this as long as the critical path doesn't rely on its contents, the way it used to; `AwsDevice` used to assign boto3 session regions by hardcoding them in a dict, which made it a pain to add new devices.

src/braket/devices/devices.py
```diff
@@ -0,0 +1,45 @@
+from enum import Enum
```
speller26, line 1: Needs license header:
```suggestion
# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
#
# Licensed under the Apache License, Version 2.0 (the "License"). You
# may not use this file except in compliance with the License. A copy of
# the License is located at
#
#     http://aws.amazon.com/apache2.0/
#
# or in the "license" file accompanying this file. This file is
# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
# ANY KIND, either express or implied. See the License for the specific
# language governing permissions and limitations under the License.

from enum import Enum
```

src/braket/aws/aws_device.py
ajberdy: I'd probably vote for `Devices` since the "identifiers" part is now somewhat abstracted from the customers. It's true that there's no functionality outside of the ARNs though, but maybe later we can add a class method to a common base class like `.create()` or something, which could have the same interface as the `AwsDevice` constructor

speller26 (approved):

PR #595, opened 2023-07-05
change: Increase types module coverage to 100% and fix related bugs

laurencap:
*Issue #, if available:*

*Description of changes:*

Increase code coverage to 100% for types modules of AutoQASM

 - Remove unused code
 - Add lots of tests: various parameter type annotations and various return type annotations
where various means: bit, bool, int, float, array, program.Program and some other random input to check implicit handling
 - Discovered some bugs:
    - bugfix for returned values that are never instantiated
    - bugfix for return variables initialized with length 0 instead of matching subroutine call return size
    - opened issue for bug: return constant value yields no instructions or incomplete instructions
    - opened issue for bug: passing arguments directly to function call doesn't give the generated variable a name
    - opened issue for bug: Array lengths are hardcoded in subroutine parameter types

*Testing done:*

New unit tests!

## Merge Checklist

#### General

#### Tests

### DIFF
--- a/src/braket/experimental/autoqasm/types/py_to_oqpy.py
+++ b/src/braket/experimental/autoqasm/types/py_to_oqpy.py
@@ -13,14 +13,11 @@
 
 """Type conversions between Python and the internal oqpy representation for types."""
 
-import ast as python_ast
 import typing
-from collections.abc import Iterable
 from functools import singledispatch
 from typing import Any, Union
 
 import numpy as np
-import openqasm3.ast as qasm_ast
 import oqpy
 
 from braket.experimental.autoqasm import program
@@ -45,14 +42,22 @@ def map_type(python_type: type) -> type:
     if issubclass(origin_type, (float, np.floating)):
         return oqpy.FloatVar
     if issubclass(origin_type, list) and issubclass(type_args[0], (int, np.integer)):
-        return oqpy.ArrayVar[oqpy.IntVar, 10]  # TODO don't use fixed length 10
+        # TODO: Update array length to match the input rather than hardcoding
+        #       OQPY and QASM require arrays have a set length. python doesn't require this,
+        #       so the length of the array is indeterminate.
+        #       At this point we only have access to the _parameter_ (type hint), not the
+        #       _argument_ (concrete value), which is the only place length information is stored
+        # Here's where the info is stored for oqpy variables:
+        # ctx = program.get_program_conversion_context()
+        # dims = ctx.get_oqpy_program().declared_vars[name_of_var].dimensions
+        return oqpy.ArrayVar[oqpy.IntVar, 10]
 
     # TODO add all supported types
     return python_type
 
 
 @singledispatch
-def wrap_value(node: Any, **kwargs: dict) -> Any:
+def wrap_value(node: Any) -> Any:
     """Wraps an object in an oqpy variable.
 
     Args:
@@ -73,7 +78,7 @@ def wrap_value(node: Any, **kwargs: dict) -> Any:
 
 
 @wrap_value.register(bool)
-def _(node: bool, **kwargs: dict):
+def _(node: bool):
     return oqpy.BoolVar(
         node, name=program.get_program_conversion_context().next_var_name(oqpy.BoolVar)
     )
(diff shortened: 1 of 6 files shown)


src/braket/experimental/autoqasm/types/py_to_oqpy.py
```diff
@@ -73,107 +78,38 @@ def wrap_value(node: Any, **kwargs: dict) -> Any:
...
-        python_ast.Lt: "<",
-        python_ast.LtE: "<=",
-        python_ast.Gt: ">",
-        python_ast.GtE: ">=",
-    }
-    return oqpy.base.OQPyBinaryExpression(
-        getattr(qasm_ast.BinaryOperator, op_map[type(op)]),
-        wrap_value(node.left),
-        wrap_value(node.comparators[0]),
-    )
```
speller26, line 179: Why the removals?

src/braket/experimental/autoqasm/types/types.py
```diff
@@ -28,10 +28,14 @@ def is_qasm_type(val: Any) -> bool:
...
-        (oqpy.Range, oqpy._ClassicalVar, oqpy.base.OQPyExpression),
-    )
+    try:
+        if issubclass(val, (oqpy.Range, oqpy._ClassicalVar, oqpy.base.OQPyExpression)):
+            return True
+    except TypeError:
+        # `val` is not a class
+        pass
+
+    return isinstance(val, (oqpy.Range, oqpy._ClassicalVar, oqpy.base.OQPyExpression))
```
speller26, line 38: Why not
```suggestion
        return issubclass(val, (oqpy.Range, oqpy._ClassicalVar, oqpy.base.OQPyExpression)):
    except TypeError:
        # `val` is not a class
        return False
```

speller26 (commented):
I think this PR does more than just raise coverage; can you include the source change details and change the PR/commit name accordingly?

speller26:
> ## [Codecov](https://app.codecov.io/gh/aws/amazon-braket-sdk-python/pull/595?src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=aws) Report
> > ❗ No coverage uploaded for pull request base (`feature/autoqasm@4114432`). [Click here to learn what that means](https://docs.codecov.io/docs/error-reference?utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=aws#section-missing-base-commit).
> > The diff coverage is `n/a`.
> 
> ```diff
> @@                 Coverage Diff                 @@
> ##             feature/autoqasm     #595   +/-   ##
> ===================================================
>   Coverage                    ?   99.21%           
> ===================================================
>   Files                       ?      146           
>   Lines                       ?     8489           
>   Branches                    ?     1836           
> ===================================================
>   Hits                        ?     8422           
>   Misses                      ?       49           
>   Partials                    ?       18           
> ```

I wonder why there's no more diff coverage information...

src/braket/experimental/autoqasm/types/py_to_oqpy.py
laurencap: Unused code, and no known way to reach these code paths. I'd consider this unsupported.

src/braket/experimental/autoqasm/types/types.py
laurencap: because `issubclass` and `isinstance` do not have the same behavior. It should be `return issubclass(...) or isinstance(...)` but `issubclass` will error if `val` is a type rather than a class (which I found surprising)

laurencap: In other words, there are cases where `isinstance(...)` is `True`, and therefore we must handle it like a QASM object, even though `issubclass` either raises an error or returns `False`

speller26: Whoops, missed that one was `isinstance` and the other was `issubclass`.

speller26: Any reason not to do this
```suggestion
        if issubclass(val, (oqpy.Range, oqpy._ClassicalVar, oqpy.base.OQPyExpression)):
            return True
    except TypeError:
        # `val` is not a class
        return isinstance(val, (oqpy.Range, oqpy._ClassicalVar, oqpy.base.OQPyExpression))
```
Just easier to reason with in my opinion, but I don't feel too strongly.

speller26 (approved):
One suggestion, feel free to take or leave

rmshaffer (approved):

src/braket/experimental/autoqasm/types/types.py
laurencap: Yes, we don't want to only check `isinstance` if a TypeError is raised. Instead, we want to ignore it. The simpler logic is `return issubclass(...) or instance(...)`, and your suggestion would check `return issubclass(...) or (issubclass raises error and isinstance(...))`

PR #612, opened 2023-07-11
Change: update how num_qubits get supplied by autoqasm functions

laurencap:
*Issue #, if available:* https://github.com/orgs/amazon-braket/projects/2/views/1?pane=issue&itemId=32622428

*Description of changes:*

Previous:
```
@aq.function
def my_func():
    for i in aq.range(5):
        cnot(i, i + 1)

prog = my_func(num_qubits=6)
```

New:
```
@aq.function(num_qubits=6)
def my_func():
    for i in aq.range(5):
        cnot(i, i + 1)

prog = my_func()
```

*Testing done:*
Updated unit tests

## Merge Checklist

#### General

#### Tests

### DIFF
--- a/src/braket/experimental/autoqasm/README.md
+++ b/src/braket/experimental/autoqasm/README.md
@@ -59,7 +59,7 @@ AutoQASM enables users to use more complicated program constructs with a compact
 structure. We can demonstrate this with an active reset program:
 ```
 # A program that actively resets a qubit back to the ground state
-@aq.function(num_qubits=1)
+@aq.function
 def reset(qubit: int, num_repeats: int) -> None:
     for repeats in aq.range(num_repeats):
         if measure(qubit):
--- a/src/braket/experimental/autoqasm/README.md
+++ b/src/braket/experimental/autoqasm/README.md
@@ -59,7 +59,7 @@ AutoQASM enables users to use more complicated program constructs with a compact
 structure. We can demonstrate this with an active reset program:
 ```
 # A program that actively resets a qubit back to the ground state
-@aq.function
+@aq.function(num_qubits=1)
 def reset(qubit: int, num_repeats: int) -> None:
     for repeats in aq.range(num_repeats):
         if measure(qubit):
@@ -67,10 +67,8 @@ def reset(qubit: int, num_repeats: int) -> None:
 ```
 
 Now that the program takes inputs, you must pass those inputs when you call your function.
-In this case, the qubits are indexed by variable, rather than by integer literals, so we
-must additionally pass the `num_qubits` keyword argument to AutoQASM.
 ```
-my_reset_program = reset(qubit=0, num_repeats=3, num_qubits=1)
+my_reset_program = reset(qubit=0, num_repeats=3)
 ```
 
 AutoQASM can support nested subroutines and complex control flow. You can use the Python runtime
--- a/src/braket/experimental/autoqasm/api/api.py
+++ b/src/braket/experimental/autoqasm/api/api.py
@@ -26,7 +26,6 @@ import braket.experimental.autoqasm.program as aq_program
 import braket.experimental.autoqasm.transpiler as aq_transpiler
 import braket.experimental.autoqasm.types as aq_types
 from braket.experimental.autoqasm import errors
-from braket.experimental.autoqasm.autograph import ag_logging
 from braket.experimental.autoqasm.autograph.core import ag_ctx, converter
 from braket.experimental.autoqasm.autograph.impl.api_core import (
     autograph_artifact,
@@ -35,7 +34,7 @@ from braket.experimental.autoqasm.autograph.impl.api_core import (
 from braket.experimental.autoqasm.autograph.tf_utils import tf_decorator
 
 
-def function(f: Callable) -> Callable[[Any], aq_program.Program]:
+def function(*args, num_qubits: int = None) -> Callable[[Any], aq_program.Program]:
     """Decorator that converts a function into a callable that returns
     a Program object containing the quantum program.
 
@@ -43,8 +42,43 @@ def function(f: Callable) -> Callable[[Any], aq_program.Program]:
     function is called, and a new Program object is returned.
 
     Args:
+        num_qubits (int): Configuration to set the total number of qubits to declare in the program.
+
+    Returns:
+        Callable[[Any], Program]: A callable which returns the converted
+        quantum program when called.
+    """
+    # First, we just process the arguments to the decorator function
+    user_config = aq_program.UserConfig(num_qubits=num_qubits)
+
+    if len(args):
+        # In this case, num_qubits wasn't supplied.
+        # Matches the following syntax:
+        #    @aq.function
+        #    def my_func(...):
+        # Equivalently, `function(my_func, num_qubits=None)`
+        return _function_without_params(args[0], user_config=user_config)
+    else:
+        # In this case, num_qubits was supplied, and we don't know `f` yet.
+        # Matches the following syntax:
+        #    @aq.function(num_qubits=x)
+        #    def my_func(...):
+        # Equivalently: `function(num_qubits=x)(my_func)`
+        def _function_with_params(f):
+            return _function_without_params(f, user_config=user_config)
+
+        return _function_with_params
+
+
+def _function_without_params(
+    f: Callable, user_config: aq_program.UserConfig
+) -> Callable[[Any], aq_program.Program]:
+    """Wrapping and conversion logic around the user function `f`.
+
+    Args:
         f (Callable): The target function to be converted which represents
             the entry point of the quantum program.
+        user_config (UserConfig): User-specified settings that influence program building
 
     Returns:
         Callable[[Any], Program]: A callable which returns the converted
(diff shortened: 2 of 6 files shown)


src/braket/experimental/autoqasm/README.md
```diff
@@ -59,18 +59,16 @@ AutoQASM enables users to use more complicated program constructs with a compact
 structure. We can demonstrate this with an active reset program:
 ```
 # A program that actively resets a qubit back to the ground state
-@aq.function
+@aq.function(num_qubits=1)
```
rmshaffer, line 62: This program does not require `num_qubits` to be specified at all. The `qubit=0` argument is evaluated by the Python interpreter at trace time, and so autoqasm correctly infers that only qubit 0 is used.

It may still be fine to keep this here for instructional reasons, just wanted to mention it.

src/braket/experimental/autoqasm/errors.py
```diff
@@ -24,12 +22,28 @@ class UnknownQubitCountError(AutoQasmError):
...
+
+
+class InconsistentNumQubits(AutoQasmError):
+    """Num qubits supplied to main function does not match subroutine."""
+
+    def __init__(self):
+        self.message = """\
+The number of qubits specified by one of your subroutines does not match the \
+arguments supplied to your main function. Remove the num_qubits decorator \
+argument to all aq.functions except the one you wish to call."""
```
rmshaffer, line 46: above you use `autoqasm.function`, and here you call it `aq.function`. I'm not sure which we should prefer (I lean toward `aq.function` since this is what all of our examples will use) but we should at least be consistent.

test/unit_tests/braket/experimental/autoqasm/test_api.py
```diff
@@ -112,14 +112,14 @@ def do_h(q: int):
...
-@aq.function
+@aq.function(num_qubits=6)
 def recursive_h(q: int):
     do_h(q)
     if q > 0:
         recursive_h(q - 1)
 
 
-@aq.function
+@aq.function(num_qubits=6)
```
rmshaffer, line 122: Ugh. This is almost a counterexample to this change, because the way this program is written, the `q` argument has to be correlated with `num_qubits`, and now we have split that to two locations in the code. This is a contrived test case so I wouldn't overindex on it, just something I hadn't thought of before.

test/unit_tests/braket/experimental/autoqasm/test_api.py
```diff
@@ -294,7 +294,7 @@ def test_bell_measurement_invalid_declared_size() -> None:
     assert expected_error_message in str(exc_info.value)
 
 
-@aq.function
+@aq.function(num_qubits=5)
```
rmshaffer, line 297: It would be pretty cool if `num_qubits` could be accessible inside the function (so that, in this case, `n_qubits` doesn't have to be declared again). But outside the scope of this PR, probably.

src/braket/experimental/autoqasm/errors.py
```diff
@@ -24,12 +22,28 @@ class UnknownQubitCountError(AutoQasmError):
...
+    def my_autoqasm_function():
+        ...
 """
 
     def __str__(self):
         return self.message
+
+
+class InconsistentNumQubits(AutoQasmError):
+    """Num qubits supplied to main function does not match subroutine."""
```
rmshaffer, line 40: nit: "main function" and "subroutine" are not really terms we have exposed to the user. But I think it's understandable regardless - probably this is fine.

src/braket/experimental/autoqasm/api/api.py
```diff
@@ -134,10 +158,29 @@ def _wrapper(*args, **kwargs) -> aq_program.Program:
     return _decorator
 
 
+def _validate_subroutine_args(user_config: aq_program.UserConfig) -> None:
+    """Validate decorator arguments to subroutines. Subroutines do not currently support
+    decorator parameters; however, we allow them if they match the arguments provided to
+    main.
```
rmshaffer, line 164: I'm not sure if I like allowing them in specific cases - it works fine for num_qubits, but will the same pattern work for other potential arguments? Is it easier to explain/document if we just disallow them completely? Or, if we just log/print a warning that we are ignoring the arguments on subroutines?

src/braket/experimental/autoqasm/api/api.py
```diff
@@ -35,16 +34,51 @@
 from braket.experimental.autoqasm.autograph.tf_utils import tf_decorator
 
 
-def function(f: Callable) -> Callable[[Any], aq_program.Program]:
+def function(*args, num_qubits: int = None) -> Callable[[Any], aq_program.Program]:
```
rmshaffer, line 37: should the type hint for `num_qubits` be `Optional[int]`?

rmshaffer (approved):
This is a big usability improvement! I left a few comments which you may or may not want to address before merging, but looks good to me overall.

src/braket/experimental/autoqasm/README.md
laurencap: Ah yes I can remove it. Actually Tim already replaced this whole section so it'll be short lived either way

src/braket/experimental/autoqasm/errors.py
laurencap: True. I also prefer `aq.function`, but technically there is no `aq`. The error message with `aq` would not be meaningful to someone using `import autoqasm` instead of `import autoqasm as aq`. I would say documentation should use the full name, and tutorials / examples should use `aq` (and anything that has the opportunity to include import statements)

laurencap: I'll update for consistency

test/unit_tests/braket/experimental/autoqasm/test_api.py
laurencap: Yeah, I also noticed that :/ I still think this is an improvement, but we'll have to be on the lookout for real world examples that run into friction with this style

test/unit_tests/braket/experimental/autoqasm/test_api.py
laurencap: good idea, should be very easy to implement. could be something like
```
@aq.function(num_qubits=5)
def my_func():
    aq.get_num_qubits()
```

src/braket/experimental/autoqasm/errors.py
laurencap: I know, writing this made me feel like we do need to make the distinction. because I couldn't figure out how to explain the difference between the one you call and the one that one calls in a clear way :/ Open to ideas

src/braket/experimental/autoqasm/api/api.py
laurencap: I think these questions can't be answered until we have other arguments besides num_qubits, or until we distinguish `main` and `subroutine` functions. But I think we can allow/disallow on a parameter by parameter basis.

This docstring is maybe too specific to the current state of affairs for what it should do in general, so I'll remove the second sentence or convert to an inline comment

src/braket/experimental/autoqasm/errors.py
krneta: Yeah. It's difficult because customers may feel like they need to specify the number of qubits each time they use the @aq.function decorator, which means that they need to keep them all in sync (or have errors thrown) and when they want to change their algorithm to use more qubits, they would need to revisit every function they created.

test/unit_tests/braket/experimental/autoqasm/test_api.py
krneta: Is it a typical use case to have wrapper functions like this? It seems like either could be the "main" if you split up the concepts of main function and subroutine. Customers would be back to needing to make sure that they sync between all of them.

src/braket/experimental/autoqasm/errors.py
```diff
@@ -24,12 +22,28 @@ class UnknownQubitCountError(AutoQasmError):
...
+
+    def __init__(self):
+        self.message = """\
+The number of qubits specified by one of your subroutines does not match the \
+arguments supplied to your main function. Remove the num_qubits decorator \
+argument to all AutoQASM decorated functions except the one you wish to call."""
```
krneta, line 46: "wish to call" can be confusing. They likely want to call them all (in some way or another) even if some of them are subroutines. 

Just needing this error points to that we'll likely making this in a way that's confusing.

laurencap: Updated the error message, thank you.

Yeah, it's a tough problem. So far, this is the least error prone, and best documented method that we've thought of that is currently available to us. At some point, we may expose two different kinds of decorators, or we may be able to perform transformations on the generated program, or we may support more output formats. If any of those changes are made, then we will have additional options for how to handle this.

test/unit_tests/braket/experimental/autoqasm/test_api.py
laurencap: What do you mean by sync between all of them? Splitting up the main/subroutine concepts would mean that these tests would have to be updated, because they would not be correct as written. Specifically:

```
@aq.subroutine  # num_qubits not accepted, cannot generate recursive_h directly
def recursive_h(q: int):
    do_h(q)
    if q > 0:
        recursive_h(q - 1)

@aq.main(num_qubits=6)
def recursive_h_wrapper():
    q = aq.get_num_qubits()  # showing for demonstration, Ryan made this suggestion
    recursive_h(q)

program = recursive_h_wrapper()
```

laurencap: I add `(num_qubits=6)` to both of them here because the tests call each of them, and that minimizes the changes to the tests. An end user should not have to call both of them -- these examples are pretty contrived as Ryan said.

src/braket/experimental/autoqasm/errors.py
laurencap: `num_qubits` should only be specified at the top level AutoQASM function. They should not be specified by nested subroutines, so there's no need to sync them up. However, we have no such distinguishing main/subroutine feature at the moment, so I can't be that precise yet.

Correct:
```
@aq.function
def nested():
    ...

@aq.function(num_qubits=5)
def main():
    nested()
```

Incorrect:
```
@aq.function(num_qubits=10)
def nested():
    ...

@aq.function(num_qubits=5)
def main():
    nested()
```

Not preferred, but not a hard failure:
```
@aq.function(num_qubits=5)
def nested():
    ...

@aq.function(num_qubits=5)
def main():
    nested()
```

Do you think the last case should be a hard failure?

laurencap: It would block the functionality you describe. I'm not convinced it's an important use case. It feels analogous to a package (lets say C or Rust) that has some defined executables, and a bunch of library code. The library code can be called from the executables. It would be odd and awkward to try to use one executable from another executable.

I did leave it for flexibility (allowing the nested code to be called directly) and it like the right approach until we decide to split `aq.main` and `aq.subroutine`

PR #624, opened 2023-07-18
feat: Run local simulator tasks asynchronously

Sai-prakash15:
*Issue #, if available:*

*Description of changes:*

*Testing done:*

## Merge Checklist

#### General

#### Tests


Sai-prakash15 (commented):
Yet to add new unit tests.

test/unit_tests/braket/devices/test_local_simulator.py
```diff
@@ -506,19 +506,19 @@ def test_init_unregistered_backend():
 @pytest.mark.xfail(raises=NotImplementedError)
 def test_run_unsupported_type():
     sim = LocalSimulator(DummyCircuitSimulator())
-    sim.run("I'm unsupported")
+    sim.run("I'm unsupported").result()
```
ajberdy, line 509: these tests are testing that the `run` method raises a NotImplementedError. By adding a `result` call, the test becomes more ambiguous and would pass if either raise the error (unwanted). If the run method is still raising the error, the result call shouldn't even be reached here

test/unit_tests/braket/tasks/test_local_quantum_task.py
```diff
@@ -29,33 +29,38 @@
...
 def test_result():
+    TASK = LocalQuantumTask(RESULT)
+    result = TASK.result()
+    assert result == RESULT
     assert RESULT.task_metadata.id == TASK.id
-    assert TASK.result() == RESULT
 
 
-@pytest.mark.xfail(raises=NotImplementedError)
+@pytest.mark.xfail(raises=AttributeError)
```
ajberdy, line 52: why? is this expected or just for now

test/unit_tests/braket/devices/test_local_simulator.py
ajberdy: Also, if you're updating tests that use the xfail decorator, I would recommend reformatting them to use `with pytest.raises(<ErrorType>, match=expected_str):`

src/braket/tasks/local_quantum_task.py
```diff
@@ -30,38 +44,306 @@ class LocalQuantumTask(QuantumTask):
...
+        except CancelledError:
+            """Cancelling the task"""
+        except Exception as e:
+            raise e
+
+    def _get_task(self) -> asyncio.Task:
+        try:
+            asyncio.get_event_loop()
+        except Exception:
+            print("No event loop found; creating new event loop")
```
ajberdy, line 228: Should this really be printed?

test/unit_tests/braket/devices/test_local_simulator.py
Sai-prakash15: > these tests are testing that the `run` method raises a NotImplementedError. By adding a `result` call, the test becomes more ambiguous and would pass if either raise the error (unwanted). If the run method is still raising the error, the result call shouldn't even be reached here

Now run method returns a LocalQuantumTask. And only calling the result will make sure that the asynchronous task is completed and the error is raised. Since the task is running asynchronously in different thread we cannot catch it explicitly in the main thread unless we call the result or check the status

test/unit_tests/braket/tasks/test_local_quantum_task.py
```diff
@@ -29,33 +29,45 @@
...
 def test_cancel():
-    TASK.cancel()
+    num_qubits = 5
+    qc = Circuit()
+    for qubit in range(num_qubits):
+        qc.h(qubit)
+
+    for i in range(num_qubits - 1):
+        qc.cnot(i, i + 1)
+    task = LocalQuantumTask().create(qc, "default", 1000)
```
Sai-prakash15, line 60: Suggestion for better circuit so it doesn't finish before `task.cancel()` is invoked ?

test/unit_tests/braket/devices/test_local_simulator.py
ajberdy: Hm I'm still not sure, since in both cases the run method returns a LocalQuantumTask. The creation of the task doesn't have any blocking factors (queues/availability windows), so it seems like it could make sense to create the task synchronously (and raise any errors during task creation here), without needing to poll for results. I believe tasks created against the API fail quickly too; how come this is different?

Sai-prakash15: When .run() it returns a LocalQuantumTask and the running of the task happens on a different thread. If we want to catch any kind of errors synchronously in the main thread we will have to call .run().result()

src/braket/tasks/local_quantum_task_batch.py
```diff
@@ -29,16 +36,112 @@ class LocalQuantumTaskBatch(QuantumTaskBatch):
...
         self._results = results
 
+    # flake8: noqa: C901
+    @staticmethod
+    def create(
+        task_specifications: Union[
+            Union[Circuit, Problem, Program, AnalogHamiltonianSimulation],
+            List[Union[Circuit, Problem, Program, AnalogHamiltonianSimulation]],
+        ],
+        delegate,  # noqa
```
speller26, line 58: Why is the `# noqa` needed?

src/braket/devices/local_simulator.py
```diff
@@ -132,51 +124,14 @@ def run_batch(
...
-                        f"Cannot execute circuit with unbound parameters: "
-                        f"{unbounded_parameters}"
-                    )
-
-        with Pool(min(max_parallel, len(tasks_and_inputs))) as pool:
-            param_list = [(task, shots, inp, *args, *kwargs) for task, inp in tasks_and_inputs]
-            results = pool.starmap(self._run_internal_wrap, param_list)
-
-        return LocalQuantumTaskBatch(results)
+        return LocalQuantumTaskBatch(local_quantum_tasks_res)
```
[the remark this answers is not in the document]
Sai-prakash15: `LocalQuantumTaskBatch.create() ` returns a `[LocalQuantumTask]` unlike `LocalQuantumTask` which returns a `LocalQuantumTask` and `LocalQuantumTask` has a reference of the `asyncio.Task`.

Do I have to do it similarly here? (creating a _tasks as a private variable on LocalQuantumTaskBatch)

src/braket/tasks/local_quantum_task_batch.py
Sai-prakash15: Type hint `LocalSimulator` is creating a circular import

src/braket/tasks/local_quantum_task_batch.py
```diff
@@ -48,42 +49,71 @@ def __init__(
...
+            circuit_spec = Circuit(...)
+            simulator = LocalSimulator()
+            task_list = LocalQuantumTask.create(task_specifications=circuit_spec,
+                                                delegate=simulator, shots=100)
+
+            # Create multiple quantum tasks using a list of Circuit and Program specifications
+            circuit_spec_1 = Circuit(...)
+            circuit_spec_2 = Circuit(...)
+            program_spec = Program(...)
+            task_list = LocalQuantumTask.create(task_specifications=[circuit_spec_1,
```
ajberdy, line 100: LocalQuantumTaskBatch?

src/braket/circuits/braket_program_context.py
```diff
@@ -134,7 +134,15 @@ def add_result(self, result: Results) -> None:
...
-    def handle_parameter_value(self, value):
+    def handle_parameter_value(self, value: Any) -> Any:
+        """Convert parameter value to required format. Default conversion is noop.
+
+        Args:
+            value (Any): parameter value
+
+        Returns:
+            Any: Required format
+        """
```
ajberdy, line 145: This doc string / type annotation has been updated on main already

src/braket/tasks/local_quantum_task.py
```diff
@@ -30,38 +44,332 @@ class LocalQuantumTask(QuantumTask):
...
+        shots: Optional[int] = None,
+        inputs: Optional[Dict[str, float]] = None,
+    ):
+        simulator = self._delegate
+        if DeviceActionType.OPENQASM in simulator.properties.action:
+            validate_circuit_and_shots(circuit, shots)
+            program = circuit.to_ir(ir_type=IRType.OPENQASM)
+            program.inputs.update(inputs or {})
+            self._args = [program, shots] + list(self._args)
+            return
```
ajberdy, line 240: Why are we doing this instead of just returning a value from the function?

Sai-prakash15: That is so true. I don't know what I was thinking when I did that, haha

src/braket/tasks/local_quantum_task.py
```diff
@@ -30,38 +44,306 @@ class LocalQuantumTask(QuantumTask):
...
+            )
+            self._thread.start()
+
+        return self._task
+
+    async def _async_run_internal(self):
+        return self._run_internal(self._task_specification, *self._args, **self._kwargs)
+
+    @singledispatchmethod
+    def _convert_to_sim_format(
```
ajberdy, line 217: this looks like it could be a static/non-class method

src/braket/tasks/local_quantum_task.py
```diff
@@ -30,38 +44,306 @@ class LocalQuantumTask(QuantumTask):
...
+    def _run_internal(
+        self,
+        task_specification: Union[
+            Circuit, Problem, Program, AnalogHamiltonianSimulation, AHSProgram
+        ],
+        shots: Optional[int] = None,
+        *args,
+        **kwargs,
+    ) -> Union[GateModelQuantumTaskResult, AnnealingQuantumTaskResult]:
+        raise NotImplementedError(f"Unsupported task type {type(task_specification)}")
```
ajberdy, line 319: How are you testing this? Is there any real code path that passes _convert_to_sim_format but fails here?

src/braket/tasks/local_quantum_task.py
```diff
@@ -30,38 +44,306 @@ class LocalQuantumTask(QuantumTask):
...
+    @_run_internal.register(Program)
+    def _(
+        self,
+        circuit: Union[Circuit, Program],
+        *args,
+        **kwargs,
+    ):
+        simulator = self._delegate
+        results = simulator.run(*args, **kwargs)
+        return GateModelQuantumTaskResult.from_object(results)
```
ajberdy, line 331: do you not need to pass circuit into simulator.run?

ajberdy: Since these implementations for _run_internal are basically identical besides the results wrapping, I'm thinking it might even be cleaner to just have one method for running and then split the logic for wrapping the result object.

```
@singledispatch
def wrap_results(results):
    raise NotImplementedError

@wrap_results.register
def _(results: GateModelTaskResult) -> GateModelQuantumTaskResult:
    return GateModelQuantumTaskResult.from_object(results)

@wrap_results.register
def _(results: AnalogHamiltonianSimulationTaskResult) -> AnalogHamiltonianSimulationQuantumTaskResult:
    return AnalogHamiltonianSimulationQuantumTaskResult.from_object(results)
```

src/braket/tasks/local_quantum_task_batch.py
```diff
@@ -29,21 +37,156 @@ class LocalQuantumTaskBatch(QuantumTaskBatch):
...
+
+        if not max_parallel:
+            max_parallel = cpu_count()
+
+        single_task = isinstance(
+            task_specifications,
+            (Circuit, Program, Problem, AnalogHamiltonianSimulation),
+        )
+
+        single_input = isinstance(inputs, dict)
```
ajberdy, line 128: this is a lot of duplicated code. Maybe it would make sense to factor out to somewhere common like `quantum_job_creation` did for local and aws jobs?

test/unit_tests/braket/devices/test_local_simulator.py
```diff
@@ -503,22 +507,28 @@ def test_init_unregistered_backend():
     LocalSimulator("foo")
 
 
-@pytest.mark.xfail(raises=NotImplementedError)
 def test_run_unsupported_type():
     sim = LocalSimulator(DummyCircuitSimulator())
-    sim.run("I'm unsupported")
+    with pytest.raises(NotImplementedError, match="Unsupported task type <class 'str'>"):
+        sim.run("I'm unsupported")
```
ajberdy, line 513: 🙌

src/braket/tasks/local_quantum_task.py
```diff
@@ -30,38 +49,280 @@ class LocalQuantumTask(QuantumTask):
...
+@wrap_results.register
+def _(
+    results: AnalogHamiltonianSimulationTaskResult,
+) -> AnalogHamiltonianSimulationQuantumTaskResult:
+    return AnalogHamiltonianSimulationQuantumTaskResult.from_object(results)
+
+
+@wrap_results.register
+def _(results: AnnealingTaskResult) -> AnnealingQuantumTaskResult:
+    return AnnealingQuantumTaskResult.from_object(results)
```
Sai-prakash15, line 328: Should I move this to a new file or a different file ?

ajberdy: If you end up creating a file in braket.tasks with shared code between aws/local tasks, this would probably go there, but I don't mind moving it later. We should it make it a private function `_wrap_results` though, so we can move it/alter it without it being a breaking change

src/braket/tasks/local_quantum_task_batch.py
```diff
@@ -29,21 +37,156 @@ class LocalQuantumTaskBatch(QuantumTaskBatch):
...
+
+        if not single_task and not single_input:
+            if len(task_specifications) != len(inputs):
+                raise ValueError(
+                    "Multiple inputs and task specifications must " "be equal in number."
```
ajberdy, line 133: linter put these on the same line but didn't combine `"..." "..."`

ajberdy: Just checked this is also in aws_quantum_task_batch, if you could update that also (/should be updated together when code is factored out)

src/braket/tasks/local_quantum_task.py
Sai-prakash15: Gotcha.

src/braket/tasks/local_quantum_task_batch.py
speller26: Wouldn't the type be `BraketSimulator` though?

speller26 (approved):
Let's keep this in the feature branch for the time being.

src/braket/tasks/local_quantum_task_batch.py
Sai-prakash15: Yes! My bad. Would it be BraketLocalSimulator or BraketSimulator ?

PR #635, opened 2023-07-21
test: update devices enum test

ajberdy:
*Issue #, if available:*

*Description of changes:*

update devices enum test to automatically generate the expected names

*Testing done:*

verified integ test passes

## Merge Checklist

#### General

#### Tests

### DIFF
--- a/test/integ_tests/test_device_creation.py
+++ b/test/integ_tests/test_device_creation.py
@@ -91,12 +91,7 @@ def test_device_enum():
     def get_device_name(device: AwsDevice):
         arn_device_name = device.arn.split("/")[-1]
 
-        device_name = (
-            arn_device_name
-            .replace("_", "")
-            .replace("-", "")
-            .replace("system", "")
-        )
+        device_name = arn_device_name.replace("_", "").replace("-", "").replace("system", "")
 
         if device.provider_name == "Amazon Braket":
             device_name = device_name.upper()
--- a/test/integ_tests/test_device_creation.py
+++ b/test/integ_tests/test_device_creation.py
@@ -74,43 +74,45 @@ def test_get_devices_all():
         assert arn in result_arns
 
 
+def _get_provider_name(device: AwsDevice):
+    arn_provider = device.arn.split("/")[2]
+
+    # capitalize as in provider name
+    provider_primary_name = device.provider_name.split()[0]
+    if arn_provider == provider_primary_name.lower():
+        capitalized = provider_primary_name
+    else:
+        capitalized = arn_provider.upper()
+
+    # remove dashes
+    return capitalized.replace("-", "")
+
+
+def _get_device_name(device: AwsDevice):
+    arn_device_name = device.arn.split("/")[-1]
+
+    device_name = arn_device_name.replace("_", "").replace("-", "").replace("system", "")
+
+    if device.provider_name == "Amazon Braket":
+        device_name = device_name.upper()
+    return device_name
+
+
 def test_device_enum():
-    def get_provider_name(device: AwsDevice):
-        arn_provider = device.arn.split("/")[2]
-
-        # capitalize as in provider name
-        provider_primary_name = device.provider_name.split()[0]
-        if arn_provider == provider_primary_name.lower():
-            capitalized = provider_primary_name
-        else:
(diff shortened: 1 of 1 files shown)


test/integ_tests/test_device_creation.py
```diff
@@ -75,42 +75,55 @@ def test_get_devices_all():
...
+
+        # remove dashes
+        return capitalized.replace("-", "")
+
+    def get_device_name(device: AwsDevice):
+        arn_device_name = device.arn.split("/")[-1]
+
+        device_name = (
+            arn_device_name
+            .replace("_", "")
```
krneta: Should this be "\W"?

test/integ_tests/test_device_creation.py
```diff
@@ -75,42 +75,55 @@ def test_get_devices_all():
...
+            .replace("system", "")
         )
+
+        if device.provider_name == "Amazon Braket":
+            device_name = device_name.upper()
+        return device_name
+
+    aws_devices = AwsDevice.get_devices()
+
+    # determine retired providers
```
krneta, line 104: Comments like this usually hint that I should just move it to a function. I do that less for test functions, but here I think it would be useful.

test/integ_tests/test_device_creation.py
```diff
@@ -75,42 +75,55 @@ def test_get_devices_all():
...
+            device_name = f"_{device_name}"
+
+        assert getattr(getattr(Devices, provider_name), device_name) == device.arn
+
+    # validate all devices in enum
+    providers = [getattr(Devices, attr) for attr in dir(Devices) if not attr.startswith("__")]
+    for provider in providers:
+        devices = [getattr(provider, attr) for attr in dir(provider) if not attr.startswith("__")]
+        for arn in devices:
+            AwsDevice(arn)
```
krneta, line 126: It might be good to have a better assert comment here.

test/integ_tests/test_device_creation.py
ajberdy: I don't think so, I believe that would actually miss "_", which is considered a word character

krneta (approved):

test/integ_tests/test_device_creation.py
```diff
@@ -74,43 +75,61 @@ def test_get_devices_all():
         assert arn in result_arns
 
 
+def _get_provider_name(device: AwsDevice) -> str:
+    arn_provider = device.arn.split("/")[2]
+
+    # capitalize as in provider name
+    provider_primary_name = device.provider_name.split()[0]
```
krneta, line 82: I'm not crazy about this heuristic, why can't we always just capitalize?

test/integ_tests/test_device_creation.py
```diff
@@ -74,43 +75,61 @@ def test_get_devices_all():
...
+        capitalized = arn_provider.upper()
+
+    # remove dashes
+    return capitalized.replace("-", "")
+
+
+def _get_device_name(device: AwsDevice) -> str:
+    arn_device_name = device.arn.split("/")[-1]
+
+    device_name = arn_device_name.replace("_", "").replace("-", "").replace("system", "")
```
krneta, line 95: I'm afraid that in the future we'll have a device with "system" in it and we won't want to remove it and we'll be confused. But I don't think it's the end of the world to change this then.

test/integ_tests/test_device_creation.py
ajberdy: Python uppercase variables are generally global constants, so I'm hesitant to use them for classes and/or constants that could one day become classes. For me, the tiebreaker that seals my vote is that changing the enum would be a breaking change and I don't feel strongly enough that this warrants that

ajberdy: We could make an argument for the device names themselves being uppercase as enum values ([docs](https://docs.python.org/3/library/enum.html)), but that wouldn't solve this issue in specific

test/integ_tests/test_device_creation.py
ajberdy: I agree, I've updated this PR to be more explicitly about this transformation (Advantage_system -> Advantage)

krneta (approved):

PR #650, opened 2023-07-27
fix: autoqasm programs with assignments inside conditionals and branches

rmshaffer:
*Issue #, if available:*
- [Branches/conditional expressions with variable assignment produce incorrect results
](https://github.com/orgs/amazon-braket/projects/2/views/1?pane=issue&itemId=32619400)

*Description of changes:*
- For conditional expressions, the `if_exp` operator needs to return the appropriate variable object. It also needs to validate that both the `if` and `else` expressions are of the same type.
- For branches (if-else blocks), manually calling `declare()` for assignment statements caused the variable to be defined inside the if-else block. Instead, we need to allow `oqpy` to automatically declare the variable, which will put it at the correct scope. Note that for subroutines, we now must call `autodeclare()` after executing the subroutine function so that `oqpy` will properly place any necessary variable declarations inside the subroutine.

*Testing done:*
- tox passes.
- Tests added for newly supported scenarios. Diff code coverage is 100%.
- Existing tests updated to pass. Note that many variable declarations have moved around a bit due to the changes in `assignments.py`. But the resulting programs are equivalent.

## Merge Checklist

#### General

#### Tests

### DIFF
--- a/src/braket/experimental/autoqasm/operators/conditional_expressions.py
+++ b/src/braket/experimental/autoqasm/operators/conditional_expressions.py
@@ -17,7 +17,6 @@
 from typing import Any, Callable, Optional
 
 import oqpy.base
-from openpulse import ast
 
 from braket.experimental.autoqasm import program as aq_program
 from braket.experimental.autoqasm import types as aq_types
@@ -44,25 +43,6 @@ def if_exp(
         return _py_if_exp(cond, if_true, if_false)
 
 
-def _aq_type_from_ast_type(ast_type: ast.ClassicalType) -> type:
-    if isinstance(ast_type, ast.IntType):
-        return aq_types.IntVar
-    if isinstance(ast_type, ast.FloatType):
-        return aq_types.FloatVar
-    if isinstance(ast_type, ast.BoolType):
-        return aq_types.BoolVar
-    if isinstance(ast_type, ast.BitType):
-        return aq_types.BitVar
-    if isinstance(ast_type, ast.ArrayType):
-        return aq_types.ArrayVar
-
-
-def _aq_type(expr_or_var: Any) -> type:
-    if isinstance(expr_or_var, oqpy.base.OQPyExpression):
-        return _aq_type_from_ast_type(expr_or_var.type)
-    return type(expr_or_var)
-
-
 def _oqpy_if_exp(
     cond: Any,
     if_true: Callable[[None], Any],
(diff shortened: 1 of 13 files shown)


src/braket/experimental/autoqasm/operators/slices.py
```diff
@@ -90,9 +90,10 @@ def _oqpy_set_item(target: Any, i: Any, x: Any) -> Any:
         raise NotImplementedError("Slice assignment is not supported.")
 
     oqpy_program = program.get_program_conversion_context().get_oqpy_program()
-    if x.name in oqpy_program.declared_vars.keys():
+    if x.name in oqpy_program.declared_vars.keys() or x.init_expression is None:
```
yitchen-tim, line 93: condition for `declared_vars` would almost never trigger because now most variables are in `undeclared_vars` until the very end

src/braket/experimental/autoqasm/operators/conditional_expressions.py
```diff
@@ -47,15 +48,28 @@ def _oqpy_if_exp(
...
+) -> Optional[oqpy.base.Var]:
     """Overload of if_exp that stages an oqpy conditional."""
-    oqpy_program = program.get_program_conversion_context().get_oqpy_program()
+    oqpy_program = aq_program.get_program_conversion_context().get_oqpy_program()
     if isinstance(cond, oqpy.base.Var) and cond.name not in oqpy_program.declared_vars.keys():
         oqpy_program.declare(cond)
+
+    result_var = None
     with oqpy.If(oqpy_program, cond):
-        if_true()
```
yitchen-tim, line 56: Question for my education: Why does this solve the issue? And why we don't need to do the same in if-else in control_flow.py

src/braket/experimental/autoqasm/operators/slices.py
yitchen-tim: Should we make `_is_variable_used` function in assignment.py a general utility? (Or even better, add it in oqpy)

src/braket/experimental/autoqasm/operators/conditional_expressions.py
rmshaffer: For conditional expressions, the entire expression (`aq.IntVar(1) if foo else aq.IntVar(2)`) is supposed to be resolved to a single value. However, OpenQASM does not support inline conditional expressions. So we have to convert this into a traditional `if`/`else` block in OpenQASM. But because the result of the conditional expression may be assigned to something, we need to do an assignment in both branches as well.

For the `if`/`else` in control_flow.py, the user has already written it as an if-else block with the appropriate assignment statements in each block. So we don't have to do this type of translation there.

src/braket/experimental/autoqasm/operators/slices.py
rmshaffer: Yes, good idea. Refactored it into `ProgramConversionContext` and called it here instead of checking `declared_vars` directly.

yitchen-tim (approved):

PR #661, opened 2023-08-07
Usability: Warning when user calls "return" from their autoqasm entry point function

laurencap:
*Issue #, if available:*

*Description of changes:*

Warn if the called function returns a value

*Testing done:*

New unit tests

## Merge Checklist

#### General

#### Tests

### DIFF
--- a/src/braket/experimental/autoqasm/converters/return_statements.py
+++ b/src/braket/experimental/autoqasm/converters/return_statements.py
@@ -25,8 +25,8 @@ from braket.experimental.autoqasm.autograph.core import ag_ctx, converter
 class ReturnValidator(converter.Base):
     def visit_Return(self, node: ast.stmt) -> ast.stmt:
         aq_context = program.get_program_conversion_context()
-        if aq_context.is_in_main and node.value is not None:
-            warnings.warn("`output` is currently unsupported; `return` statement has no effect.")
+        if not aq_context.subroutines_processing and node.value is not None:
+            warnings.warn("Return value from top level function is ignored.")
         return node
 
 
--- a/src/braket/experimental/autoqasm/converters/return_statements.py
+++ b/src/braket/experimental/autoqasm/converters/return_statements.py
@@ -1,0 +1,40 @@
+# Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You
+# may not use this file except in compliance with the License. A copy of
+# the License is located at
+#
+#     http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+
+
+"""Converters for return statement nodes."""
+
+import ast
+import warnings
+
+from braket.experimental.autoqasm import program
+from braket.experimental.autoqasm.autograph.converters import return_statements
+from braket.experimental.autoqasm.autograph.core import ag_ctx, converter
+
+
+class ReturnValidator(converter.Base):
+    def visit_Return(self, node: ast.stmt) -> ast.stmt:
+        aq_context = program.get_program_conversion_context()
+        if aq_context.is_in_main and node.value is not None:
+            warnings.warn("`output` is currently unsupported; `return` statement has no effect.")
+        return node
+
+
+def transform(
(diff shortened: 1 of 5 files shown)


yitchen-tim:
Looks great!

yitchen-tim (approved):

src/braket/experimental/autoqasm/converters/return_statements.py
```diff
@@ -0,0 +1,40 @@
...
+from braket.experimental.autoqasm import program
+from braket.experimental.autoqasm.autograph.converters import return_statements
+from braket.experimental.autoqasm.autograph.core import ag_ctx, converter
+
+
+class ReturnValidator(converter.Base):
+    def visit_Return(self, node: ast.stmt) -> ast.stmt:
+        aq_context = program.get_program_conversion_context()
+        if aq_context.is_in_main and node.value is not None:
+            warnings.warn("`output` is currently unsupported; `return` statement has no effect.")
```
rmshaffer, line 29: Technically the `return` statement still has an effect (it ends execution of the program), it's only the returned object that is ignored. Also, will the user know what `output` means in this context? (that's an OpenQASM concept)

src/braket/experimental/autoqasm/program/program.py
```diff
@@ -88,6 +88,7 @@ def __init__(self, user_config: Optional[UserConfig] = None):
         self.return_variable = None
         self._qubits_seen = set()
         self._var_idx = 0
+        self._in_main = False
```
rmshaffer, line 91: Just curious, could you check that `self.subroutines_processing` is empty, instead of introducing a new variable for this? Or does `subroutines_processing` behave differently than what you need?

src/braket/experimental/autoqasm/program/program.py
```diff
@@ -107,6 +108,15 @@ def qubits(self) -> List[int]:
...
 
+    @property
+    def is_in_main(self) -> bool:
+        """Whether the main block is currently being processed."""
+        return self._in_main
+
+    def in_main(self, in_main: bool) -> None:
+        """Set context for processing the `main` block."""
+        self._in_main = in_main
+
```
rmshaffer, line 119: Do you get a benefit from making this a property rather than just using/accessing it as a simple variable `self.is_main`?

src/braket/experimental/autoqasm/converters/return_statements.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+def transform(
+    node: ast.stmt, ctx: ag_ctx.ControlStatusCtx, default_to_null_return: bool = True
+) -> ast.stmt:
+    """Handle AutoQASM-specific return statement functionality before
+    passing control to AutoGraph.
+    """
+    ReturnValidator(ctx).visit(node)
+    return return_statements.transform(node, ctx)
```
rmshaffer, line 40: This is a cool way to override/extend the built-in converters!

rmshaffer (changes requested):
Makes sense overall, thanks for doing this! Left a few comments/questions.

src/braket/experimental/autoqasm/program/program.py
laurencap: As a property, no, except for documentation, but the setter syntax becomes a function call. I was on the fence about this myself, I preferred `context.in_main(True)` but only stylistically. I'm happy to update it as a simple variable

src/braket/experimental/autoqasm/converters/return_statements.py
laurencap: Good points -- Struggling to think of a way to explain to the user that they might not want to return a value without connecting to openqasm. Maybe "Returning a value from the top level function call has undefined behavior"? Or, we could close this and allow it to be ambiguous until we implement `output`

rmshaffer: Yeah this is a weird one to explain - especially if we plan to support it in the future. Maybe just something to communicate that "returning a value from the top-level function doesn't do anything right now".

src/braket/experimental/autoqasm/program/program.py
rmshaffer: Documentation is a good point. I'm indifferent, just curious.

laurencap: Turns out it's not necessary anyway, because your rec about `subroutines_processing` worked great!

src/braket/experimental/autoqasm/program/program.py
laurencap: Thank you!! This worked great

rmshaffer (approved):

PR #662, opened 2023-08-07
Raise error if the user puts a break in their autoqasm loop

laurencap:
*Issue #, if available:*

*Description of changes:*
Previously, break in a loop would silently have no effect. Now, raise an error.

*Testing done:*
Unit tests

## Merge Checklist

#### General

#### Tests

### DIFF
--- a/src/braket/experimental/autoqasm/converters/break_statements.py
+++ b/src/braket/experimental/autoqasm/converters/break_statements.py
@@ -1,0 +1,32 @@
+# Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You
+# may not use this file except in compliance with the License. A copy of
+# the License is located at
+#
+#     http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+
+
+"""Converters for return statement nodes."""
+
+import ast
+
+from braket.experimental.autoqasm import errors
+from braket.experimental.autoqasm.autograph.core import ag_ctx, converter
+
+
+class BreakValidator(converter.Base):
+    def visit_Break(self, node: ast.stmt) -> ast.stmt:
+        """Break statements are currently unsupported."""
+        raise errors.UnsupportedFeature("Break statements are currently unsupported.")
+
+
+def transform(
+    node: ast.stmt, ctx: ag_ctx.ControlStatusCtx, default_to_null_return: bool = True
+) -> ast.stmt:
+    return BreakValidator(ctx).visit(node)
--- a/src/braket/experimental/autoqasm/converters/break_statements.py
+++ b/src/braket/experimental/autoqasm/converters/break_statements.py
@@ -12,7 +12,7 @@
 # language governing permissions and limitations under the License.
 
 
-"""Converters for return statement nodes."""
+"""Converters for break statement nodes."""
 
 import ast
 
--- a/src/braket/experimental/autoqasm/converters/break_statements.py
+++ b/src/braket/experimental/autoqasm/converters/break_statements.py
@@ -17,6 +17,7 @@
 import ast
 
 from braket.experimental.autoqasm import errors
+from braket.experimental.autoqasm.autograph.converters import break_statements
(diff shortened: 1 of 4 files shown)


test/unit_tests/braket/experimental/autoqasm/test_converters.py
```diff
@@ -59,3 +59,32 @@ def fn() -> None:
...
+        main()
+
+
+def test_break_while_loop():
+    @aq.function
+    def uses_while():
+        while aq.gates.measure(0):
+            aq.gates.x(0)
+
+    uses_while()
```
yitchen-tim, line 81: Why do we need to test `while` here?

test/unit_tests/braket/experimental/autoqasm/test_converters.py
```diff
@@ -59,3 +59,32 @@ def fn() -> None:
...
+
+    @aq.function
+    def uses_while_w_break():
+        while aq.gates.measure(0):
+            aq.gates.x(0)
+            break
+
+    with pytest.raises(aq.errors.UnsupportedFeature):
+        uses_while_w_break()
```
rmshaffer, line 90: I am not sure that `break` statements in `while` loops are actually unsupported. I think it might work. (autograph converts the `break` condition in `for` loops to an `extra_test` function that we need to handle, but there is no such thing in `while` loops - I think it transpiles the `break` condition into part of the loop condition.)

However, from a UX perspective it makes complete sense IMO to disallow both until we actually support both and have fully tested them, etc.

src/braket/experimental/autoqasm/converters/break_statements.py
```diff
@@ -0,0 +1,32 @@
...
+class BreakValidator(converter.Base):
+    def visit_Break(self, node: ast.stmt) -> ast.stmt:
+        """Break statements are currently unsupported."""
+        raise errors.UnsupportedFeature("Break statements are currently unsupported.")
+
+
+def transform(
+    node: ast.stmt, ctx: ag_ctx.ControlStatusCtx, default_to_null_return: bool = True
+) -> ast.stmt:
+    return BreakValidator(ctx).visit(node)
```
rmshaffer, line 32: Do you want to call the autograph converter here as well? e.g. `return break_statements.transform(node, ctx)`?

I guess that would be unreachable code, but maybe you could add it anyway with `# pragma: no cover` just so we don't forget it's needed when we support break statements again? Or at least add a comment that it should be added back here?

src/braket/experimental/autoqasm/converters/break_statements.py
```diff
@@ -0,0 +1,32 @@
...
+#
+#     http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+
+
+"""Converters for return statement nodes."""
```
rmshaffer, line 15:
```suggestion
"""Converters for break statement nodes."""
```

rmshaffer (approved):

test/unit_tests/braket/experimental/autoqasm/test_converters.py
laurencap: You're right, we don't. I'll remove it, since it's tested elsewhere.

I had originally thought `while` wasn't supported, because it was broken for a bit while I was debugging, then, realizing it was supported, I was suspicious that my implementation was messing with the while functionality, so I added this, but it should be cleaned up now.

test/unit_tests/braket/experimental/autoqasm/test_converters.py
laurencap: Right now on `feature/autoqasm`, a while loop with a `break` statement fails because we don't yet support `and`

laurencap: More concretely, I can reproduce the same error with these inputs:
```
def main():
    while measure(0):
        aq.gates.x(0)
        break
```
and
```
def main():
    while measure(0) and measure(1):
        aq.gates.x(0)
```

Also, there's some unexpected behavior that `measure(0)` is only executed once, and the same output is referenced every time the condition is evaluated, so I'll open a bug report for that

laurencap: https://github.com/orgs/amazon-braket/projects/2/views/6?pane=issue&itemId=35372617

PR #667, opened 2023-08-09
Return type hint ignored when possible in favor of actual return type

laurencap:
*Issue #, if available:*

*Description of changes:*
 - rename `UnsupportedFeature` -> `UnsupportedFeatureError`
 - Use actual return variable to set return type, except for recursive functions (task created: https://github.com/orgs/amazon-braket/projects/2/views/1?pane=issue&itemId=35482594)
 - better error message if user supplies `list` instead of `list[type]` for parameter type hint
 - follow up task: https://github.com/orgs/amazon-braket/projects/2/views/1?pane=issue&itemId=35482594

*Testing done:*

unit tests

### TODO
 - [ ] Fix coverage

## Merge Checklist

#### General

#### Tests

### DIFF
--- a/src/braket/experimental/autoqasm/api.py
+++ b/src/braket/experimental/autoqasm/api.py
@@ -290,10 +290,10 @@ def _convert_program_as_subroutine(
         subroutine_function_call = oqpy_sub(oqpy_program, *args, **kwargs)
 
     # Add the subroutine invocation to the program
-    return_instance = _make_return_instance(f, program_conversion_context)
+    ret_type = subroutine_function_call.subroutine_decl.return_type
+    return_instance = _make_return_instance_from_oqpy_return_type(ret_type)
     return_variable = None
     if isinstance(return_instance, list):
-        ret_type = subroutine_function_call.subroutine_decl.return_type
         return_variable = aq_types.ArrayVar(
             return_instance,
             dimensions=[d.value for d in ret_type.dimensions],
@@ -320,9 +320,18 @@ def _convert_program_as_subroutine(
         root_oqpy_program._add_subroutine(subroutine_name, subroutine_function_call.subroutine_decl)
 
 
-def _make_return_instance(
-    f: Callable, program_conversion_context: aq_program.ProgramConversionContext
-) -> Any:
+def _make_return_instance_from_oqpy_return_type(return_type) -> Any:
+    if not return_type:
+        return None
+
+    return_type = aq_types.conversions.var_type_from_ast_type(return_type)
+    if return_type == aq_types.ArrayVar:
+        return []
+    return return_type()
+
+
+def _make_return_instance_from_f_annotation(f: Callable) -> Any:
+    # TODO: Recursive functions should work even if the user's type hint is wrong
     annotations = f.__annotations__
     return_type = annotations["return"] if "return" in annotations else None
 
@@ -386,7 +395,7 @@ def _clone_function(f_source: Callable) -> Callable:
 
 
 def _dummy_function(f_source: Callable) -> Callable:
-    return_instance = _make_return_instance(f_source, aq_program.get_program_conversion_context())
+    return_instance = _make_return_instance_from_f_annotation(f_source)
 
     def f_dummy(*args, **kwargs) -> Any:
         return return_instance  # pragma: no cover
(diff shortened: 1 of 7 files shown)


test/unit_tests/braket/experimental/autoqasm/test_types.py
rmshaffer, line 1: Should _missing_ return type hints also work correctly with these changes? Is it worth adding test case(s) for that as well? I wonder if we can support it well enough that we could just tell users that return type hints are not required at all.

(I guess it may not work yet for recursive functions, but for the non-recursive case it seems like it should work?)

rmshaffer (approved):

src/braket/experimental/autoqasm/errors.py
```diff
@@ -21,7 +21,11 @@ class AutoQasmError(Exception):
     """Base class for all AutoQASM exceptions."""
 
 
-class UnsupportedFeature(AutoQasmError):
+class UnsupportedFeatureError(AutoQasmError):
+    """AutoQASM unsupported feature."""
+
+
+class ParameterTypeError(AutoQasmError):
     """AutoQASM unsupported feature."""
```
laurencap, line 29: TODO

test/unit_tests/braket/experimental/autoqasm/test_types.py
laurencap: I'm looking into it a bit more. The relevant code is line 287 of api.py

laurencap: We don't do this kind of parsing or validation in autoqasm right now. Instead, I have to see if I can re-mangle the oqpy structures after they've been built (since the required info isn't available at the time it's built). If we had a Program class and could run some passes on it, that would probably make this more straightforward :)

laurencap: Proving quite difficult. Unfortunately, I can't explain it succinctly (I don't understand it well enough), but I could go through it over a call. Let's move forward with this PR as is and keep the recursive type hints in a separate task.

Also -- added the tests you mentioned Ryan, thank you.

src/braket/experimental/autoqasm/api.py
```diff
@@ -320,9 +320,18 @@ def _convert_program_as_subroutine(
...
+    return_type = aq_types.conversions.var_type_from_ast_type(return_type)
+    if return_type == aq_types.ArrayVar:
+        return []
+    return return_type()
+
+
+def _make_return_instance_from_f_annotation(f: Callable) -> Any:
+    # TODO: Recursive functions should work even if the user's type hint is wrong
     annotations = f.__annotations__
     return_type = annotations["return"] if "return" in annotations else None
```
laurencap, line 336: Need to add coverage for this func

rmshaffer: There is something weird about the code coverage detection for this codepath. We had to add `#pragma: no cover` inside `f_dummy` below, even though that code is actually hit during tests (e.g., if you put `raise NotImplementedError` inside `f_dummy`, it causes tests to fail). So I'd be curious to know if this code is actually not hit during tests, or if it's just a bug in the coverage detection.

test/unit_tests/braket/experimental/autoqasm/test_types.py
rmshaffer: Agree, the recursive case can be handled separately (especially since it's much less likely to be used). I guess a big part of the problem is that (as the code is today) we don't actually invoke the recursive function calls while tracing the program, so we never get to see what value is returned.

test/unit_tests/braket/experimental/autoqasm/test_types.py
```diff
@@ -535,3 +558,113 @@ def main():
...
+    return retval_;
+}
+bool __bool_1__;
+__bool_1__ = ret_test();"""
+
+    assert main().to_ir() == expected
+
+
+def test_ignore_ret_typehint_list():
+    """Test type discovery of boolean return values."""
```
rmshaffer, line 587:
```suggestion
    """Test type discovery of list return values."""
```

src/braket/experimental/autoqasm/api.py
laurencap: It used to get called for any subroutine, now it's only called for recursive subroutines. I had to add a new test with a recursive subroutine that returned a `List`

PR #673, opened 2023-08-11
[autoqasm] Extend gate set

yitchen-tim:
*Issue #, if available:*
https://github.com/orgs/amazon-braket/projects/2/views/6?pane=issue&itemId=32619795

*Description of changes:*
- Rename the `gates` module as `instructions`, which include gates, reset and measure
- Extend the gate set

*Testing done:*
Unit tests

## Merge Checklist

#### General

#### Tests


yitchen-tim:
Somehow an AutoQASM unrelated part fails the coverage test occasionally. Redoing the check would fix it.

src/braket/experimental/autoqasm/instructions/quantum_gates.py
```diff
@@ -0,0 +1,89 @@
...
+        oqpy_program.gate(qubits, name, *angles)
+
+    gate_doc = description + "\n\n Args:"
+    for i in range(n_control):
+        gate_doc += f"\n\tcontrol_{i} (QubitIdentifierType): Control qubit {i}."
+    for i in range(n_target):
+        gate_doc += f"\n\ttarget_{i} (QubitIdentifierType): Target qubit {i}."
+    for i in range(n_angle):
+        gate_doc += f"\n\tangle_{i} (float): Rotation angle {i} in radians."
+    quantum_gate.__doc__ = gate_doc
```
rmshaffer, line 49: Including the docstring is a great step! But I think this still wouldn't allow a linter or IDE to provide type checking for the params for each gate. This could be sort of hacked by setting `quantum_gate.__signature__` but I'm not sure how well that will work.

Just curious, did you consider at all actually using/extending the existing BDK gate definitions in `braket.circuits.gates`? Is there a reason that AutoQASM needs a separate set of gate definitions, or a different pattern for defining/calling them?

src/braket/experimental/autoqasm/instructions/__init__.py
```diff
@@ -11,7 +11,19 @@
 # ANY KIND, either express or implied. See the License for the specific
 # language governing permissions and limitations under the License.
 
-"""Quantum operations for use in AutoQASM programs."""
+"""Instructions that apply to qubits, including quantum gates, reset and measure.
 
-from .gates import QubitIdentifierType, cnot, cphaseshift, h, reset, rx, rz, x, y, z  # noqa: F401
+Example of using a `h` gate and a `cnot` gate to create a Bell circuit:
```
laurencap, line 16: Awesome!! Great ops excellence to keep the docs bar raising as we go along 🙌

src/braket/experimental/autoqasm/instructions/instructions.py
```diff
@@ -0,0 +1,31 @@
...
+#     http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+
+
+"""Instructions that apply to qubits.
+"""
```
laurencap, line 16: Possible to elaborate on how they differ from gates? Perhaps
```suggestion
"""Non-unitary instructions that apply to qubits."""
```

src/braket/experimental/autoqasm/instructions/quantum_gates.py
yitchen-tim: I considered extending the SDK Gate class. Currently, there is not much we can reuse from SDK Gate class because AutoQASM serializes using oqpy while SDK uses hard-coded strings. We can possibly add a `_to_oqpy` method to each gate in SDK, but that would me adding AutoQASM program context dependencies into core part of SDK. Not sure we are ready to do it yet. In addition, SDK gates are classes, while in AutoQASM we want functions. We would need to add a wrapper on SDK gates too.

yitchen-tim: Good point about IDE typing checking. I tried a little bit and I was able to add type hints too using `__signature__`. But I think we might as well define the gate functions explicitly as that is more readable and maintainable.

yitchen-tim: Also, I just found that the way I do it does not automatically generate docs in html pages

yitchen-tim: Based on some searching, IDE type checking is based on either function signature or type annotations. I include both in the gate functions. Now, all gate functions have their own docs, function signatures and type hints.

rmshaffer: I do agree that defining each gate function explicitly is more readable, and more likely to be compatible with all IDE features. I know this is quite a bit more tedious to write but perhaps it is the most robust solution for now.

yitchen-tim: What works for me is the jupyter environment. In jupyter, tab completion works for these gates. When I do `instructions.` followed by tab, it shows a list of gates.
image

I can also do `cnot?` which gives me docstring.

yitchen-tim: It turned out that, in VS code and PyCharm, this type of dynamic function definition is not supported (couple issues I found: [1](https://github.com/microsoft/pylance-release/issues/2394) and [2](https://github.com/microsoft/pylance-release/issues/2052)). The reason is likely that the extensions in these editors only do static analysis, not dynamic analysis. So I might need to switch to the explicit definition of gates.

yitchen-tim: Switched to the explicit function definitions

src/braket/experimental/autoqasm/instructions/instructions.py
yitchen-tim: updated!

src/braket/experimental/autoqasm/instructions/quantum_gates.py
```diff
@@ -0,0 +1,89 @@
...
+#
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+
+
+"""Quantum gates, which are applied to qubits. Some gates take parameters as arguments in addition
+to the qubits.
+"""
```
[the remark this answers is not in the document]
yitchen-tim: updated!

src/braket/experimental/autoqasm/instructions/gates.py
```diff
@@ -0,0 +1,456 @@
...
+
+
+"""Quantum gates, unitary instructions, that apply to qubits.
+"""
+from braket.experimental.autoqasm import program
+
+from .qubits import _qubit
+
+
+def ccnot(control_0: int, control_1: int, target: int) -> None:
```
rmshaffer, line 22: I believe all of the qubit args should have type hint `QubitIdentifierType` rather than `int`.

src/braket/experimental/autoqasm/instructions/gates.py
```diff
@@ -0,0 +1,456 @@
...
+    """CCNOT gate or Toffoli gate.
+
+    Args:
+        control_0 (QubitIdentifierType): Control qubit 0.
+        control_1 (QubitIdentifierType): Control qubit 1.
+        target (QubitIdentifierType): Target qubit.
+    """
+    qubits = [_qubit(control_0), _qubit(control_1), _qubit(target)]
+    oqpy_program = program.get_program_conversion_context().get_oqpy_program()
+    oqpy_program.gate(qubits, "ccnot")
```
rmshaffer, line 32: Is it worth reducing duplication a bit (and to isolate oqpy-related logic) by having a common implementation for all of these gates? For example, maybe this goes into `instructions.py`:
```
def _qubit_instruction(name: str, qubits: List[QubitIdentifierType], *args: Any):
    oqpy_program = program.get_program_conversion_context().get_oqpy_program()
    oqpy_program.gate([_qubit(q) for q in qubits], name, *args)
```
and then each gate (or other instruction such as `reset`) can be implemented as ~1 line:
```
def cphaseshift00(control: QubitIdentifierType, target: QubitIdentifierType, angle: float) -> None:
    """Controlled phase shift gate for phasing the \\|00> state.
    Args:
        control (QubitIdentifierType): Control qubit.
        target (QubitIdentifierType): Target qubit.
        angle (float): Rotation angle in radians.
    """
    _qubit_instruction("cphaseshift00", [control, target], angle)
```

rmshaffer (changes requested):
This approach makes sense! Just a couple more comments.

PR #675, opened 2023-08-14
update: restricting parameter names to not collide with ones we use for OpenQASM generation.

krneta:
*Issue #, if available:*
https://github.com/aws/amazon-braket-sdk-python/issues/603

*Description of changes:*
When generating OpenQASM we decided to use 'b' and 'q' for 'bits' and 'qubits' variables, but this can collide with parameters chosen by customers. I've renamed these to ```__bits__``` and  ```__qubits__``` and have changed our requirements for FreeParameter names so that we exclude names that start with double underlines. I've also updated the verification function for our FreeParameter names to be closer to the identifiers definition from [OpenQASM](https://openqasm.com/language/types.html) to further reduce changes there will be transformation errors (some of which are mentioned in the ticket above).

This check is not exact, so it's possible that we'll allow for FreeParameter names that are not translatable to OpenQASM, since there's a trade-off between performance and upkeep (and keeping track of every reserved word in OpenQASM).

*Testing done:*

- Ran tests and unit tests and did searches. I hope there's nowhere that I've missed a 'b' or 'q' used in this way because they're quite hard to search for.

## Merge Checklist

#### General

#### Tests

### DIFF
--- a/src/braket/parametric/free_parameter.py
+++ b/src/braket/parametric/free_parameter.py
@@ -47,12 +47,12 @@ class FreeParameter(FreeParameterExpression):
             name (str): Name of the :class:'FreeParameter'. Must begin with a letter [A-Za-z],
                 an underscore or an element from the Unicode character categories Lu/Ll/Lt/Lm/Lo/Nl.
                 Must not begin with two underscores '__'. May contain numbers [0-9] after the
-                first characeter.
+                first character.
 
         Examples:
             >>> param1 = FreeParameter("theta")
             >>> param1 = FreeParameter("\u03B8")
-            >>> param1 = FreeParameter("a1")
+            >>> param1 = FreeParameter("a1b2")
         """
         FreeParameter._validate_name(name)
         self._name = Symbol(name)
--- a/src/braket/parametric/free_parameter.py
+++ b/src/braket/parametric/free_parameter.py
@@ -15,7 +15,6 @@ from __future__ import annotations
 
 from numbers import Number
 from typing import Dict, Union
-from unicodedata import category
 
 from sympy import Symbol
 
@@ -44,15 +43,12 @@ class FreeParameter(FreeParameterExpression):
         Initializes a new :class:'FreeParameter' object.
 
         Args:
-            name (str): Name of the :class:'FreeParameter'. Must begin with a letter [A-Za-z],
-                an underscore or an element from the Unicode character categories Lu/Ll/Lt/Lm/Lo/Nl.
-                Must not begin with two underscores '__'. May contain numbers [0-9] after the
-                first character.
+            name (str): Name of the :class:'FreeParameter'. Can be a unicode value.
+                Must not start with '__'.
 
         Examples:
             >>> param1 = FreeParameter("theta")
             >>> param1 = FreeParameter("\u03B8")
-            >>> param1 = FreeParameter("a1b2Ï")
         """
         FreeParameter._validate_name(name)
         self._name = Symbol(name)
@@ -111,13 +107,8 @@ class FreeParameter(FreeParameterExpression):
             raise ValueError("FreeParameter names must be non empty")
         if not isinstance(name, str):
             raise TypeError("FreeParameter name must be a string")
-        if name.startswith("_"):
-            if name.startswith("__"):
-                raise ValueError("FreeParameter names must not start with two underscores '__'")
-        else:
-            uni_category = category(name[0])
-            if not uni_category or (not uni_category.startswith("L") and uni_category != "Nl"):
-                raise ValueError("FreeParameter names must start with a letter or underscore")
+        if name.startswith("__"):
+            raise ValueError("FreeParameter names must not start with two underscores '__'")
 
     @classmethod
     def from_dict(cls, parameter: dict) -> FreeParameter:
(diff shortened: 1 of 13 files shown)


src/braket/parametric/free_parameter.py
```diff
@@ -99,6 +105,22 @@ def to_dict(self) -> dict:
             "name": self.name,
         }
 
+    @staticmethod
+    def _validate_name(name: str) -> None:
```
math411, line 109: If we only validate for the symbol creation, can we use a `setter` here and combine L58?

src/braket/parametric/free_parameter.py
```diff
@@ -99,6 +105,22 @@ def to_dict(self) -> dict:
...
+        if not name:
+            raise ValueError("Symbol names must be non empty")
```
math411, line 111:
```suggestion
            raise ValueError("FreeParameter names must be non empty")
```
also applies to L113, L116

test/unit_tests/braket/parametric/test_free_parameter.py
```diff
@@ -61,3 +61,39 @@ def test_sub_successful(free_parameter):
...
+    "name",
+    (
+        "",
+        "1",
+        "__a",
+        "!",
+        "\u33B8",
+    ),
+)
+@pytest.mark.xfail(raises=ValueError, strict=True)
```
math411, line 97: Use ```with pytest.raises(ValueError)``` instead

src/braket/parametric/free_parameter.py
```diff
@@ -99,6 +105,22 @@ def to_dict(self) -> dict:
...
+        if not isinstance(name, str):
+            raise TypeError("Symbol name must be a string")
+        if name.startswith("_"):
```
math411, line 114: What is the purpose of the nested `if` statement?

src/braket/parametric/free_parameter.py
```diff
@@ -99,6 +105,22 @@ def to_dict(self) -> dict:
...
+            if name.startswith("__"):
+                raise ValueError("Symbol names must not start with two underscores '__'")
+        else:
+            unicode_category = category(name[0])
+            if not unicode_category:
+                raise ValueError("Symbol names must start with a valid character")
+            if not unicode_category.startswith("L") and unicode_category != "Nl":
```
math411, line 121: This breaks my use of unsightly circuits. 

```
\U0001f600
```
is no longer valid. This raises the question of whether we should exploring fuzzing certain parts of the BDK...

math411: This also seems to break in the case we have:

```
FreeParameter("_a")
```

src/braket/parametric/free_parameter.py
krneta: Good call.

src/braket/parametric/free_parameter.py
krneta: Names with one underscore are allowed. Names with two are not.

src/braket/parametric/free_parameter.py
krneta: _a should still work. \U0001f600 can't be converted to OpenQASM, so we should probably disallow it. What do you think?

src/braket/parametric/free_parameter.py
krneta: This is kind of a big deal because it potentially breaks customers who were using two underlines or other unicode characters before. However, they weren't able to convert those to OpenQASM anyways. So I'm wondering if there's a use case where we let customers use those as long as they bind them afterwards? Or would it be too complicated to check when generating OpenQASM...

test/unit_tests/braket/parametric/test_free_parameter.py
math411: The SDK has been misusing the xfail. Its [intended purpose](https://docs.pytest.org/en/7.1.x/how-to/skipping.html) is more so for bug fixes or items that are not implemented.

src/braket/parametric/free_parameter.py
krneta: Yes, I reduced the scope of the change to only disallow __

src/braket/parametric/free_parameter.py
krneta: This has been reduced as well

src/braket/parametric/free_parameter.py
krneta: Override the setter? That seems like overkill.

krneta: added a setter

src/braket/parametric/free_parameter.py
```diff
@@ -99,6 +100,19 @@ def to_dict(self) -> dict:
...
+    def _set_name(self, name: str) -> None:
+        FreeParameter._validate_name(name)
+        self._name = Symbol(name)
+
+    @staticmethod
+    def _validate_name(name: str) -> None:
+        if not name:
+            raise ValueError("FreeParameter names must be non empty")
+        if not isinstance(name, str):
+            raise TypeError("FreeParameter name must be a string")
```
math411, line 112:
```suggestion
            raise TypeError("FreeParameter names must be strings")
```

math411 (approved):

PR #679, opened 2023-08-21
feature: Add @aq.gate decorator for AutoQASM custom gate definitions

rmshaffer:
*Issue #, if available:*
- [Support gate subroutine as `aq.gate`](https://github.com/orgs/amazon-braket/projects/2/views/1?pane=issue&itemId=32539968)

*Description of changes:*
- Add `@aq.gate` decorator to define a custom gate. This differs from an `@aq.function` in that it can only contain unitary gate operations. It is defined in OpenQASM as a `gate` statement rather than a `def` subroutine.
- **Note:** This PR does not yet enforce that the gate definition only contains calls to other gates. I will do that in a separate PR in order to keep PR size reasonable.

*Testing done:*
- Several tests added to cover supported and unsupported scenarios.

**Example usage:**
```
@aq.gate
def my_gate(q: aq.Qubit, theta: float):
    h(q)
    rz(q, theta)

@aq.function
def main():
    my_gate(0, np.pi / 4)
    my_gate(1, np.pi / 2)

print(main().to_ir())
```
produces the following OpenQASM:
```
OPENQASM 3.0;
gate my_gate(theta) q {
    h q;
    rz(theta) q;
}
qubit[2] __qubits__;
my_gate(pi / 4) __qubits__[0];
my_gate(pi / 2) __qubits__[1];
```

## Merge Checklist

#### General

#### Tests


src/braket/experimental/autoqasm/api.py
```diff
@@ -320,6 +372,60 @@ def _convert_program_as_subroutine(
...
+
+        # Add the gate invocation to the program
+        if len(args) != len(gate_args):
+            raise errors.ParameterTypeError(
+                f'Incorrect number of arguments passed to gate "{gate_name}". '
+                f"Expected {len(gate_args)}, got {len(args)}."
+            )
+        qubit_args = [args[i] for i, (_, is_qubit) in enumerate(gate_args) if is_qubit]
+        angle_args = [args[i] for i, (_, is_qubit) in enumerate(gate_args) if not is_qubit]
+        aq_instructions.instructions._qubit_instruction(gate_name, qubit_args, *angle_args)
```
rmshaffer, line 419: The argument handling here may be somewhat confusing and I'm open to suggestions. I wanted to allow the Python function signature to be flexible and allow qubit and angle parameters in any arbitrary order. But because OpenQASM treats qubit and angle argument lists separately, the Python argument list has to be separated in order to add the gate to the oqpy program.

src/braket/experimental/autoqasm/program/program.py
```diff
@@ -87,6 +88,7 @@ class ProgramConversionContext:
     def __init__(self, user_config: Optional[UserConfig] = None):
         self.oqpy_program_stack = [oqpy.Program()]
         self.subroutines_processing = set()  # the set of subroutines queued for processing
+        self.gate_definitions_processing = []
```
rmshaffer, line 91: I'll change this to be a private variable.

src/braket/experimental/autoqasm/program/program.py
```diff
@@ -171,52 +194,40 @@ def get_oqpy_program(self) -> oqpy.Program:
...
-            self.oqpy_program_stack = oqpy_program_stack
-
-        def __enter__(self):
-            self.oqpy_program_stack.append(self.oqpy_program)
-
-        def __exit__(self, exc_type, exc_value, exc_tb):
-            self.oqpy_program_stack.pop()
-
-    def push_oqpy_program(self, oqpy_program: oqpy.Program) -> OqpyProgramContextManager:
+    @contextlib.contextmanager
```
rmshaffer, line 197: For consistency with the new code, I also refactored/simplified the two existing context managers here to use `@contextlib.contextmanager` rather than manually implementing a class with `__enter__` and `__exit__`.

test/unit_tests/braket/experimental/autoqasm/test_decorator_gate.py
```diff
@@ -0,0 +1,271 @@
...
+        h(q0)
+        x(q1)
+
+    @aq.function
+    def incorrect_arg_count():
+        my_gate(0)
+
+    with pytest.raises(
+        errors.ParameterTypeError,
+        match='Incorrect number of arguments passed to gate "my_gate". ' "Expected 2, got 1.",
```
rmshaffer, line 107:
```suggestion
        match='Incorrect number of arguments passed to gate "my_gate". Expected 2, got 1.",
```

src/braket/experimental/autoqasm/__init__.py
```diff
@@ -40,8 +40,8 @@ def my_program():
     result[1] = measure __qubits__[1];
 """
 
-from .api import function  # noqa: F401
-from .instructions import QubitIdentifierType  # noqa: F401
+from .api import function, gate  # noqa: F401
+from .instructions import QubitIdentifierType as Qubit  # noqa: F401
```
rmshaffer, line 44: This was the simplest way to get a user-visible `Qubit` type for use as a type hint in gate definitions. I'm open to suggestions here (or we can defer this until we decide what to do with PhysicalQubit).

src/braket/experimental/autoqasm/api.py
```diff
@@ -71,6 +72,40 @@ def _function_with_params(f):
...
+            gate definition is ignored, and only the gate signature is declared so that it can be
+            called from elsewhere in the program. This may be used, for example, if the program
+            is being combined with an external list of gate definitions. Defaults to False.
+
+    Returns:
+        Callable[[Any], None]: A callable which can be used as a custom gate inside an
+        aq.function or inside another aq.gate.
+    """
+    if not args:
+        # In this case, declaration_only was supplied, and we don't know `f` yet.
```
yitchen-tim, line 89: How does a service interpret a gate without a definition? Are you think about the gate that service supports but AutoQASM doesn't?

rmshaffer: Yes, basically. I'm thinking of a case where the developer wants to generate an OpenQASM program which uses gates defined somewhere else - maybe in the service, maybe in a local `.inc` file, etc. (For example, we could potentially use this to declare gates which are supported by a particular device, but we don't want to redefine them in the program that we generate.)

src/braket/experimental/autoqasm/program/program.py
laurencap: Great improvement!

test/unit_tests/braket/experimental/autoqasm/test_decorator_gate.py
```diff
@@ -0,0 +1,271 @@
+# Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
```
laurencap, line 1: Nit: should this not be `test_gate_decorator.py`?

src/braket/experimental/autoqasm/__init__.py
laurencap: Looks fine to me 👍

test/unit_tests/braket/experimental/autoqasm/test_decorator_gate.py
```diff
@@ -0,0 +1,271 @@
...
+import numpy as np
+import pytest
+from test_api import _test_on_local_sim
+
+import braket.experimental.autoqasm as aq
+from braket.experimental.autoqasm import errors
+from braket.experimental.autoqasm.instructions import h, measure, rx, rz, x
+
+
+@aq.gate
```
laurencap, line 26: Fantastic tests! Really top-notch, catching lots of edge cases and interesting/creative usage paths. I feel like I understand the feature well reading these :)

src/braket/experimental/autoqasm/api.py
```diff
@@ -321,6 +379,62 @@ def _convert_program_as_subroutine(
...
+        if not declaration_only:
+            # Process the gate definition
+            angles = [oqpy.AngleVar(name=name) for name, is_qubit in gate_args if not is_qubit]
+            with program_conversion_context.gate_definition(gate_name, qubits, angles):
+                # TODO - enforce that nothing gets added to the program inside here except gates
+                wrapped_f(qubits, *angles)
+
+            # Add the gate definition to the root-level program if necessary
+            root_oqpy_program = program_conversion_context.oqpy_program_stack[0]
+            if gate_name not in root_oqpy_program.gates:
```
laurencap, line 416: What happens if you decorate two different functions with the same name? Can you add a couple tests for it?

src/braket/experimental/autoqasm/api.py
laurencap: It feels a bit confusing to me, but I think it's just that it's lower-level than most of the rest of the function. Could this be factored out into a function for readability?

laurencap (changes requested):
Couple minor suggestions. Pretty cool, and definitely bar-raising tests

src/braket/experimental/autoqasm/api.py
```diff
@@ -72,6 +73,40 @@ def _function_with_params(f: Callable) -> Callable[[Any], aq_program.Program]:
...
 
 
+def gate(*args, declaration_only: bool = False) -> Callable[[Any], None]:
+    """Decorator that converts a function into a callable gate definition.
+
+    Args:
+        declaration_only (bool): If declaration_only is True, the provided body of the
+            gate definition is ignored, and only the gate signature is declared so that it can be
+            called from elsewhere in the program. This may be used, for example, if the program
+            is being combined with an external list of gate definitions. Defaults to False.
```
laurencap, line 83: I'm still a little confused about the use case. If `ccnot` is defined elsewhere, why would someone define it a second time?

rmshaffer: The gate may be defined may be in an external OpenQASM file. The user wants to call the gate in an AutoQASM program, but they do not want to provide an implementation for it.

laurencap: Ah, so it wouldn't appear in the resulting program, but it would run successfully. Could we do something simple and tailored just for that? Like
```
@aq.function
def main():
    my_gate = aq.gate(num_qubits=1, args=["lambda"])
    my_gate(0, lambda=3.14)
```

test/unit_tests/braket/experimental/autoqasm/test_decorator_gate.py
rmshaffer: Thanks! Although I'd also credit the code coverage requirement for forcing me down some of these unnatural paths.

src/braket/experimental/autoqasm/api.py
rmshaffer: No, I don't think the Braket SDK has this concept yet. One advantage of doing this with a decorated function is that you get a named function with named parameters. Also you could, say, build a library of all of your supported/native gate declarations (if you are a quantum hardware provider) and provide it to your users.

However, I think that you and @yitchen-tim have convinced me that the use case here isn't really clear or immediate enough yet. I'll remove the `declaration_only` option from the PR for now.

src/braket/experimental/autoqasm/api.py
rmshaffer: I'll remove the `declaration_only` option from this PR for now - see note here: https://github.com/aws/amazon-braket-sdk-python/pull/679#discussion_r1304179497.

src/braket/experimental/autoqasm/api.py
rmshaffer: Adding a couple of tests. Basically I think the Python semantics should be respected here, the latest-defined function will be the one that is used.

test/unit_tests/braket/experimental/autoqasm/test_gate_decorator.py
```diff
@@ -92,6 +92,65 @@ def main():
...
+
+
+def test_duplicate_gate_names_in_subroutine() -> None:
+    """Verify that gates can only be defined at the top level."""
+
+    @aq.function
+    def define_gate_in_subroutine():
+        @aq.gate
+        def my_gate(q: aq.Qubit):
+            h(q)
```
laurencap, line 126: This is cool! Wouldn't be able to do this like this in openqasm I think

PR #684, opened 2023-08-25
Move error handling into the context manager

laurencap:
Doing an AutoQASM `api.py` refactor. Making this as a first, small PR, because it is separable, and a gitdiff with indentation changes and other changes can be confusing.

There are NO code changes beyond de-dentation and moving code

Source: https://vaclavkosar.com/software/Python-Context-Manager-With-Statement-Exception-Handling


rmshaffer (approved):
And also getting rid of some code duplication, nice!

PR #685, opened 2023-08-25
Use a callback to simplify autoqasm/api logic

laurencap:
Refactor api.py part 2

 - Was able to consolidate duplicated code through a callback function
     - Minus 100 LOC!
 - Also discovered that NullContext doesn't have any affect (used in place of optional context managers: https://docs.python.org/3/library/contextlib.html)

The diff is a bit confusing, so here's a more detailed description:
 - There are three decorators: `main`, `subroutine` and `gate`. Each call `_function_wrapper`
 - There are three decorator converters: `_convert_main`, `_convert_subroutine` and `_convert_gate`
 - `_convert_program_wrapper` and `_convert_gate_wrapper` are merged into `_function_wrapper`
 - `_function_wrapper` accepts a callback, which is one of the `_convert_X` functions
 - `_convert_program` is deleted, and its code is subsumed into `_convert_main` and/or `_convert_subroutine`, as necessary
 - `is_subroutine` is no longer necessary
 - user_config is the only extra argument to the `_convert_X` functions, and only needed by `_convert_main`
 - converter option for `recursive` was always set to `False`, and it defaults to `False`, so I removed it
 - `_add_qubit_declaration` was moved to be local to its caller without code changes
 - moved private functions to follow "step down" ordering -- this means functions are called in closer proximity to where they are defined. This is particularly appropriate here because most of the helper functions are called from only one other function
 - removed unnecessary `_decorator` wrapper which did nothing

### DIFF
--- a/src/braket/experimental/autoqasm/api.py
+++ b/src/braket/experimental/autoqasm/api.py
@@ -60,7 +60,7 @@ def main(*args, num_qubits: Optional[int] = None) -> Callable[[Any], aq_program.
         #    @aq.main
         #    def my_func(...):
         # Equivalently, `main(my_func, num_qubits=None)`
-        return _function_wrapper(args[0], user_config=user_config)
+        return _function_wrapper(args[0], _convert_main, wrapper_args={"user_config": user_config})
     else:
         # In this case, num_qubits was supplied, and we don't know `f` yet.
         # Matches the following syntax:
@@ -68,7 +68,7 @@ def main(*args, num_qubits: Optional[int] = None) -> Callable[[Any], aq_program.
         #    def my_func(...):
         # Equivalently: `main(num_qubits=x)(my_func)`
         def _function_with_params(f: Callable) -> Callable[[Any], aq_program.Program]:
-            return _function_wrapper(f, user_config=user_config)
+            return _function_wrapper(f, _convert_main, wrapper_args={"user_config": user_config})
 
         return _function_with_params
 
@@ -81,7 +81,7 @@ def subroutine(*args) -> Callable[[Any], aq_program.Program]:
         Callable[[Any], Program]: A callable which returns the converted
         quantum program when called.
     """
-    return _function_wrapper(*args, user_config=None, is_subroutine=True)
+    return _function_wrapper(args[0], _convert_subroutine)
 
 
 def gate(*args) -> Callable[[Any], None]:
@@ -91,27 +91,21 @@ def gate(*args) -> Callable[[Any], None]:
         Callable[[Any],]: A callable which can be used as a custom gate inside an
         aq.function or inside another aq.gate.
     """
-    f = args[0]
-    if is_autograph_artifact(f):
-        return f
-
-    wrapper_factory = _convert_gate_wrapper()
-
-    return autograph_artifact(wrapper_factory(f))
+    return _function_wrapper(args[0], _convert_gate)
 
 
 def _function_wrapper(
     f: Callable,
-    user_config: aq_program.UserConfig,
-    is_subroutine: bool = False,
+    callback: Callable,
+    wrapper_args: Optional[Dict[str, Any]] = None,
 ) -> Callable[[Any], aq_program.Program]:
     """Wrapping and conversion logic around the user function `f`.
 
     Args:
         f (Callable): The target function to be converted which represents
             the entry point of the quantum program.
-        user_config (UserConfig): User-specified settings that influence program building.
-        is_subroutine (bool): If the function corresponds to a subroutine or main.
+        callback (Callable): TODO
+        wrapper_args (Optional[Dict[str, Any]]): TODO
 
     Returns:
         Callable[[Any], Program]: A callable which returns the converted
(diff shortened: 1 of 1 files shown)


src/braket/experimental/autoqasm/api.py
```diff
@@ -60,15 +60,15 @@ def main(*args, num_qubits: Optional[int] = None) -> Callable[[Any], aq_program.
...
         # In this case, num_qubits was supplied, and we don't know `f` yet.
         # Matches the following syntax:
         #    @aq.main(num_qubits=x)
         #    def my_func(...):
         # Equivalently: `main(num_qubits=x)(my_func)`
         def _function_with_params(f: Callable) -> Callable[[Any], aq_program.Program]:
-            return _function_wrapper(f, user_config=user_config)
+            return _function_wrapper(f, _convert_main, wrapper_args={"user_config": user_config})
 
         return _function_with_params
```
rmshaffer, line 73: Not directly related to your changes, but I realized when doing the `gate()` decorator that the implementation of `main()` could be simplified a little bit:

```
def main(*args, num_qubits: Optional[int] = None) -> Callable[[Any], aq_program.Program]:
   """docstring here"""
    if not args:
        # Only keyword args were passed to the decorator, so we don't know `f` yet.
        # Matches the following syntax:
        #    @aq.main(num_qubits=x)
        #    def my_func(...):
        # Equivalently: `aq.main(num_qubits=x)(my_func)`
        def _function_with_params(*args) -> Callable[[Any], aq_program.Program]:
            return main(*args, num_qubits=num_qubits)
        return _function_with_params

    user_config = aq_program.UserConfig(num_qubits=num_qubits)
    return _function_wrapper(args[0], _convert_main, wrapper_args={"user_config": user_config})
```

src/braket/experimental/autoqasm/api.py
```diff
@@ -91,27 +91,21 @@ def gate(*args) -> Callable[[Any], None]:
...
 ) -> Callable[[Any], aq_program.Program]:
     """Wrapping and conversion logic around the user function `f`.
 
     Args:
         f (Callable): The target function to be converted which represents
             the entry point of the quantum program.
-        user_config (UserConfig): User-specified settings that influence program building.
-        is_subroutine (bool): If the function corresponds to a subroutine or main.
+        callback (Callable): TODO
+        wrapper_args (Optional[Dict[str, Any]]): TODO
```
rmshaffer, line 108: Couple of TODOs here

src/braket/experimental/autoqasm/api.py
```diff
@@ -120,74 +114,26 @@ def _function_wrapper(
...
+            return callback(f, options, args, kwargs, **wrapper_args)
 
         if inspect.isfunction(f) or inspect.ismethod(f):
             _wrapper = functools.update_wrapper(_wrapper, f)
 
         decorated_wrapper = tf_decorator.make_decorator(f, _wrapper)
         return autograph_artifact(decorated_wrapper)
 
-    return _decorator
+    return autograph_artifact(_decorator(f))
```
rmshaffer, line 136: It looks like `_decorator` already calls `autograph_artifact` on its return value - can it be removed here?
```suggestion
    return _decorator(f)
```

src/braket/experimental/autoqasm/api.py
rmshaffer: In fact, I wonder if it would be possible to move that `if not args:` block into `_function_wrapper` as well? This would also make it so that, for example, `@aq.subroutine` and `@aq.subroutine()` always behave equivalently, for all of our decorators - which could potentially be a paper cut. Then `main()` could end up being just two lines:
```
    user_config = aq_program.UserConfig(num_qubits=num_qubits)
    return _function_wrapper(args[0], _convert_main, wrapper_args={"user_config": user_config})
```

rmshaffer (approved):
Couple of relatively minor comments. This is already a great win for code size and maintainability!

src/braket/experimental/autoqasm/api.py
laurencap: Oh, we actually don't need the `_decorator` function at all!

src/braket/experimental/autoqasm/api.py
laurencap: thank you!

src/braket/experimental/autoqasm/api.py
laurencap: I couldn't get it to work inside _function_wrapper, but I could try again as a follow up

laurencap: For now I deleted the comments and removed the "else"

src/braket/experimental/autoqasm/api.py
```diff
@@ -120,74 +103,23 @@ def _function_wrapper(
...
+    if not converter_args:
+        converter_args = {}
 
+    def _wrapper(*args, **kwargs) -> Callable:
+        """Wrapper that calls the converted version of f."""
+        options = converter.ConversionOptions(
+            user_requested=True,
+            optional_features=_autograph_optional_features(),
+        )
+        # Call the appropriate function converter
```
laurencap, line 115: the whole _decorator wrapper was superfluous

rmshaffer (approved):

PR #713, opened 2023-09-27
doc: Add FAQs to AutoQASM README

laurencap:
Update README

### DIFF
--- a/src/braket/experimental/autoqasm/README.md
+++ b/src/braket/experimental/autoqasm/README.md
@@ -161,16 +161,16 @@ AutoQASM functionality such as subroutines.
 
 ### 2. What is the relationship between AutoQASM and OpenQASM?
 
-AutoQASM can be seen as implementing a builder pattern for OpenQASM. It
-allows you serialize your program to OpenQASM with `Program.to_ir()`. The
-interface is not strongly tied to OpenQASM so we could serialize to other
-formats in the future.
-
-In other words, AutoQASM is a quantum programming interface built in Python.
+AutoQASM is a quantum programming interface built in Python.
 OpenQASM is a quantum assembly language, often used as a serialization format
 for quantum programming frameworks and quantum hardware providers. We can
 represent a quantum program equivalently in either format, but using AutoQASM
 allows one to also make use of Python, including the Amazon Braket SDK.
+
+AutoQASM can be seen as implementing a builder pattern for OpenQASM. It
+allows you serialize your program to OpenQASM with `Program.to_ir()`. The
+interface is not strongly tied to OpenQASM, so we could serialize to other
+formats in the future.
 
 ### 3. What is the relationship between AutoQASM and the Amazon Braket SDK?
 
@@ -179,11 +179,6 @@ branch. It supplements the program building experience and integrates with
 Amazon Braket SDK features. For instance, one can build a program through
 AutoQASM, and then use the SDK to run the program on a local simulator or on
 an Amazon Braket device.
-
-Quantum programs are serialized to OpenQASM before executing on Amazon
-Braket, and AutoQASM programs can be serialized to OpenQASM. Thus, we have a
-very lightweight integration to run AutoQASM programs through the Amazon
-Braket SDK.
 
 ### 4. Does AutoQASM support other providers beyond Amazon Braket?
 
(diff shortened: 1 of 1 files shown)


src/braket/experimental/autoqasm/README.md
```diff
@@ -136,10 +142,100 @@ Please tag your question with "Amazon Braket" and mention AutoQASM in the questi
...
+### 6. Do the devices available through Amazon Braket support all of AutoQASM's features?
+
+No, for example, the `reset` instruction is not supported by all devices. In
+general, different QPUs and QHPs support different sets of features, so
+AutoQASM will often support features that a particular device doesn't
+support. We intend that AutoQASM will eventually be able to generate any
+program representable by OpenQASM 3.0, with additional Python-side features
+such as validation and visualization.
+
+### 7. Is there a difference between classical conditionals and quantum conditionals?
```
laurencap, line 213: Is this more confusing than it's worth?

src/braket/experimental/autoqasm/README.md
```diff
@@ -136,10 +142,100 @@ Please tag your question with "Amazon Braket" and mention AutoQASM in the questi
...
+AutoQASM can be seen as implementing a builder pattern for OpenQASM. It
+allows you serialize your program to OpenQASM with `Program.to_ir()`. The
+interface is not strongly tied to OpenQASM so we could serialize to other
+formats in the future.
+
+In other words, AutoQASM is a quantum programming interface built in Python.
+OpenQASM is a quantum assembly language, often used as a serialization format
+for quantum programming frameworks and quantum hardware providers. We can
+represent a quantum program equivalently in either format, but using AutoQASM
+allows one to also make use of Python, including the Amazon Braket SDK.
```
rmshaffer, line 173: I would probably swap the order of these two paragraphs - the second paragraph more directly answers the question.

src/braket/experimental/autoqasm/README.md
```diff
@@ -136,10 +142,100 @@ Please tag your question with "Amazon Braket" and mention AutoQASM in the questi
...
+
+### 3. What is the relationship between AutoQASM and the Amazon Braket SDK?
+
+AutoQASM lives alongside the Amazon Braket SDK as an experimental feature
+branch. It supplements the program building experience and integrates with
+Amazon Braket SDK features. For instance, one can build a program through
+AutoQASM, using Amazon Braket SDK device objects, and then run the program
+with the SDK.
```
rmshaffer, line 181:
```suggestion
AutoQASM, and then use the SDK to run the program on a local simulator or on
an Amazon Braket device.
```

src/braket/experimental/autoqasm/README.md
```diff
@@ -136,10 +142,100 @@ Please tag your question with "Amazon Braket" and mention AutoQASM in the questi
...
+
+Quantum programs are serialized to OpenQASM before executing on Amazon
+Braket, and AutoQASM programs can be serialized to OpenQASM. Thus, we have a
+very lightweight integration to run AutoQASM programs through the Amazon
+Braket SDK.
```
rmshaffer, line 186: Consider deleting this paragraph - I think (?) the point is that the Braket service accepts OpenQASM, we already said that during the main part of the README above, and I'm not sure it directly helps to answer the question here about the SDK.

src/braket/experimental/autoqasm/README.md
rmshaffer: I'm not sure it's really a question about AutoQASM. Your answer does start to explain the difference between trace-time and runtime conditionals, but that's also not what the question is asking. Maybe something like this could go in an example notebook that introduces control flow and mid-circuit measurement?

rmshaffer (approved):
Thanks, these are very helpful improvements! Left a few comments and suggestions.

src/braket/experimental/autoqasm/README.md
```diff
@@ -136,10 +142,100 @@ Please tag your question with "Amazon Braket" and mention AutoQASM in the questi
...
 ## Tests
 
-To run only AutoQASM tests (and skip the rest of the unit tests), run:
+To run only AutoQASM tests (and skip the rest of the Amazon Braket SDK unit tests), run:
 ```
 tox -e unit-tests -- test/unit_test/braket/experimental/autoqasm
 ```
 
 Note that you may first need to run `pip install -e .[test]`. More information on running tests
 can be found in the [top-level README](../../../../README.md).
```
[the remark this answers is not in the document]
laurencap: I feel like this is a pretty standard section in READMEs, and our own top level readme for the BDK has such a section

src/braket/experimental/autoqasm/README.md
laurencap: I was trying to include an example of the benefit of integrating with the BDK beyond running on a device, but maybe we don't have enough functionality there yet

PR #743, opened 2023-10-13
feat: use results dir as working directory for decorator jobs

ajberdy:
*Issue #, if available:*

*Description of changes:*

use results dir as working directory for decorator jobs

*Testing done:*

Unit tests and manual verification of functionality. Integ tests passing and updated to test new functionality

## Merge Checklist

#### General

#### Tests


src/braket/jobs/_entry_point_template.py
```diff
@@ -1,8 +1,16 @@
...
 import cloudpickle
-from braket.jobs import save_job_result
+import os
+from braket.jobs import get_results_dir, save_job_result
 from braket.jobs_data import PersistedJobDataFormat
 
+
+# set working directory to results dir
+os.chdir(get_results_dir())
+
```
mbeach-aws, line 10: is it also possible to mkdir if they don't exist? Example
```
np.save("my_dir/file.npy")
```
should this automatically create the dir "my_dir"?

ajberdy: I don't know the answer to that, but that's squarely a numpy functionality question. We shouldn't be inspecting path arguments to check whether they map to existing directories etc

src/braket/jobs/_entry_point_template.py
```diff
@@ -1,8 +1,16 @@
...
+# create symlinks to input data
+link_input()
```
kshitijc, line 12: Can we prevent the input data from getting populated in the results? Since the "recovered()" function encapsulates the entirety of the job script, could we unlink the input data in our template after that function finishes execution? This would ensure that the function has had a chance to consume the input data, but remove the input data before the results are packaged.

ajberdy: Sure, we can add that. I can make sure that each link is a symlink and points to the original destination before deleting, in case it has been overwritten explicitly by the user.

src/braket/jobs/_entry_point_template.py
```diff
@@ -1,46 +1,68 @@
...
+# create symlinks to input data
+links = link_input()
+
 # load and run serialized entry point function
 recovered = cloudpickle.loads({serialized})
 def {function_name}():
     result = recovered()
     if result is not None:
         save_job_result(result, data_format=PersistedJobDataFormat.PICKLED_V4)
+    # clean_links(links)
```
kshitijc, line 20: Should this be commented?

kshitijc (commented):
Could we update tests to assert this behavior?

src/braket/jobs/_entry_point_template.py
ajberdy: Ah right, that was just me testing default behavior

ajberdy:
> Could we update tests to assert this behavior?

Yes, I'm working on the tests currently

test/integ_tests/test_create_quantum_job.py
```diff
@@ -227,9 +227,24 @@ def decorator_job(a, b: int, c=0, d: float = 1.0, **extras):
...
+    with tempfile.TemporaryDirectory() as temp_dir:
+        os.chdir(temp_dir)
+        job.download_result()
+        assert (
+            Path(job.name, "results.json").exists()
+            and not Path(job.name, "test", "integ_tests", "requirements.txt").exists()
+        )
+        with open(Path(job.name, "output_file.txt"), "r") as f:
+            assert f.read() == "hello"
+        os.chdir(current_dir)
```
kshitijc, line 246: What happens if the test fails before this?

ajberdy: good point. Copied this pattern from above, so I'll update in both places. The reason for switching the cwd is because download_result downloads to the cwd without an option to specify another location

kshitijc (approved):

PR #751, opened 2023-10-20
test: pytest skip python version mismatch

ajberdy:
*Issue #, if available:*

*Description of changes:*

xfail unsupported python versions

*Testing done:*

Ran integ tests from 3.9 and saw 
```
[gw1] [ 50%] XFAIL test/integ_tests/test_create_quantum_job.py::test_decorator_job_submodule 
[gw0] [100%] XFAIL test/integ_tests/test_create_quantum_job.py::test_decorator_job 

```

Manually verified integ tests FAIL if the test either succeeds or raises an incorrect type of error

## Merge Checklist

#### General

#### Tests

### DIFF
--- a/test/integ_tests/test_create_quantum_job.py
+++ b/test/integ_tests/test_create_quantum_job.py
@@ -198,6 +198,11 @@ def test_completed_quantum_job(aws_session, capsys):
         assert data in log_data
 
 
+@pytest.mark.xfail(
+    (sys.version_info.major, sys.version_info.minor) != decorator_python_version,
+    raises=RuntimeError,
+    reason="Python version mismatch",
+)
 def test_decorator_job():
     class MyClass:
         attribute = "value"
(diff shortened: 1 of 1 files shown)


math411:
Can we use tox to avoid running this test based on the env?

ajberdy:
> Can we use tox to avoid running this test based on the env? Or alternatively, can we use
> 
> ```
> @pytest.mark.skipif(sys.version_info != (3, 10))
> ```

Actually, I think there's value in running these tests and the skip is only reached upon correctly raising the python version mismatch error. We'd be losing coverage by not running these at all. The cost is low, since the python check happens at job definition time and no jobs are created.

I think it's ambiguous whether to count this as a "PASSED" or "SKIPPED" test, since it is technically testing a valid codepath, but due to its severely limited scope, I felt that skipped is semantically closer to the truth (we're not getting the full coverage that test was designed for).

math411:
In that case, I would prefer to use `mark.xfail` as we expect these failures during particular use cases. During those use cases (running on py3.9), if it passes, we will be alerted of an issue.

ajberdy:
> In that case, I would prefer to use `mark.xfail` as we expect these failures during particular use cases. During those use cases (running on py3.9), if it passes, we will be alerted of an issue.

That makes sense. I'll update

ajberdy:
We're losing a bit of coverage with this change, since xfail won't match the error message, but we're gaining coverage by asserting that it does indeed fail. The specific error is covered by unit tests, so I am not concerned

math411 (approved):

Issue #757, opened 2023-10-24
AutoQASM: Add support for FreeParameter as gate arguments


laurencap:
This allows statements such as `rx(0, FreeParameter("theta"))` in AutoQASM programs.

When converted to OpenQASM, the resulting program has an input statement, such as `input float[32] theta;`.

https://openqasm.com/language/directives.html#input-output

> The input modifier can be used to indicate that one or more variable declarations represent values which will be provided at run-time, upon invocation. This allows the programmer to use the same compiled circuits which only differ in the values of certain parameters. For backward compatibility, OpenQASM3 does not require an input declaration to be provided. When an input declaration is provided, the compiler produces an executable that leaves these free parameters unspecified: a circuit run would take as input both the executable and some choice of the parameters.

input type for FreeParameter (`input float32 a`)

laurencap:
Closed by https://github.com/amazon-braket/amazon-braket-sdk-python/pull/762
Additional work _is_ required, but new issues have been made to track that work.

PR #762, opened 2023-10-26
[feature] AutoQASM: Free parameters

laurencap:
*Issue #, if available:*
Part 1 of #757

*Description of changes:*

This allows statements such as `rx(0, FreeParameter("theta"))` in AutoQASM programs.

When converted to OpenQASM, the resulting program has an input statement, such as `input float[32] theta;`.

This PR will be followed by these additional works to complete the feature:
1. Support expressions of parameters
2. Support other types besides float
3. Support free parameters in conditional statements
4. Documentation -- when the feature is more complete

*Testing done:*
New test file!

## Merge Checklist

#### General

#### Tests

### DIFF
--- a/src/braket/experimental/autoqasm/instructions/gates.py
+++ b/src/braket/experimental/autoqasm/instructions/gates.py
@@ -12,8 +12,9 @@
 # language governing permissions and limitations under the License.
 
 
-"""Quantum gates, unitary instructions, that apply to qubits.
-"""
+"""Quantum gates, unitary instructions, that apply to qubits."""
+
+from typing import Union
 
 from braket.circuits.free_parameter_expression import FreeParameterExpression
 
@@ -54,14 +55,14 @@ def cnot(
 def cphaseshift(
     control: QubitIdentifierType,
     target: QubitIdentifierType,
-    angle: float | FreeParameterExpression,
+    angle: Union[float, FreeParameterExpression],
 ) -> None:
     """Controlled phase shift gate.
 
     Args:
         control (QubitIdentifierType): Control qubit.
         target (QubitIdentifierType): Target qubit.
-        angle (float | FreeParameterExpression): Rotation angle in radians.
+        angle (Union[float, FreeParameterExpression]): Rotation angle in radians.
 
     """
     _qubit_instruction("cphaseshift", [control, target], angle)
@@ -70,14 +71,14 @@ def cphaseshift(
 def cphaseshift00(
     control: QubitIdentifierType,
     target: QubitIdentifierType,
-    angle: float | FreeParameterExpression,
+    angle: Union[float, FreeParameterExpression],
 ) -> None:
     """Controlled phase shift gate for phasing the \\|00> state.
 
     Args:
         control (QubitIdentifierType): Control qubit.
         target (QubitIdentifierType): Target qubit.
-        angle (float | FreeParameterExpression): Rotation angle in radians.
+        angle (Union[float, FreeParameterExpression]): Rotation angle in radians.
 
     """
     _qubit_instruction("cphaseshift00", [control, target], angle)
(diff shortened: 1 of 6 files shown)


test/unit_tests/braket/experimental/autoqasm/test_parameters.py
```diff
@@ -0,0 +1,267 @@
...
+rx_alpha(2);"""
+    assert parametric().to_ir() == expected
+
+
+def test_captured_parameter():
+    """Test that a parameter declared in a larger scope is captured
+    and functions correctly.
+    """
+
+    alpha = FreeParameter("alpha")
```
rmshaffer, line 139: This is cool, I hadn't considered this use case.

src/braket/experimental/autoqasm/instructions/gates.py
```diff
@@ -52,14 +55,14 @@ def cnot(
 def cphaseshift(
     control: QubitIdentifierType,
     target: QubitIdentifierType,
-    angle: float,
+    angle: Union[float, FreeParameterExpression],
```
rmshaffer, line 58: I wonder if there is a good way to define an alias for this union type. Seems like user-defined gates should also technically be using this as a type hint, and we wouldn't want users to have to copy this. Does it make sense to have an `AngleIdentifierType` that we expose as `aq.Angle` (just like `aq.Qubit` for `QubitIdentifierType`)? Or is that too specific - are there other types that should be supported with this pattern too? Gates can only take angles, but in general `input` parameters to a program can be other types.

rmshaffer (approved):
🎉

src/braket/experimental/autoqasm/instructions/gates.py
laurencap: Ah I'm noticing now that `gate_calibrations` produce signatures with `angle`, so we should prioritize consistency using `angles` everywhere. I didn't realize they are already used in some places

ajberdy: > Aaron pointed out that we'll want to migrate from float to OpenQASM angles, because gates are not supposed to take floats.

This is just a consideration for the IR we generate. Users will still certainly provide floats

test/unit_tests/braket/experimental/autoqasm/test_parameters.py
```diff
@@ -0,0 +1,289 @@
...
+    task = device.run(program, shots=10, inputs=inputs)
+    assert isinstance(task, LocalQuantumTask)
+    assert isinstance(task.result().measurements, dict)
+    return task.result().measurements
+
+
+@aq.main
+def simple_parametric():
+    rx(0, FreeParameter("theta"))
+    measure(0)
```
ajberdy, line 36: I'm a bit confused by this ux over something like
```
@aq.main
def simple_parametric(theta):
    rx(0, theta)
    measure(0)
```
It seems like ^ is simpler and gives users a spot to designate inputs as ints etc

src/braket/experimental/autoqasm/instructions/gates.py
rmshaffer: Yeah, gate definitions in OQ3 must use `angle` params, so we do use that for `@aq.gate` and `@aq.gate_calibration`. OQ3 allows implicitly casting `float` to `angle`, so technically everything else could use `float` and be fine, I think. But I also agree that if possible we should try to avoid exposing that to users, who just want to use `float` everywhere.

test/unit_tests/braket/experimental/autoqasm/test_parameters.py
rmshaffer: I believe this already works with this implementation. User can then call either `simple_parametric(pi/2)` or `simple_parametric(FreeParameter("theta"))` depending on whether they want a fixed or parameterized program.

src/braket/experimental/autoqasm/instructions/gates.py
laurencap: Ah okay, wasn't aware of the implicit cast. That's good news

PR #767, opened 2023-10-27
infra: code freeze workflow

ajberdy:
*Issue #, if available:*

*Description of changes:*

Support for selectively freezing main branch

*Testing done:*

Manual testing of the following cases:

- Freeze without specifying feature branch: 
  - Feature branch can be merged
  - Critical fix can be merged
  - Other code cannot be merged
- Freeze with specifying feature branch:
  - Matching feature branch can be merged (exact and prefix match)
  - Other feature branch cannot be merged
- Freeze off
  - Any code can be merged (check is skipped)

## Merge Checklist

#### General

#### Tests

### DIFF
--- a/.github/workflows/code-freeze.yml
+++ b/.github/workflows/code-freeze.yml
@@ -7,9 +7,10 @@ on:
 
 permissions:
   contents: read
+
 env:
-  FROZEN: "false"  # Set to "true" to freeze branches, "false" otherwise
-  FEATURE_NAME: ""  # Set your feature prefix here (leave empty to allow all feature branches)
+  FROZEN: ${{ vars.FROZEN }}
+  UNFROZEN_PREFIX: ${{ vars.UNFROZEN_PREFIX }}
 
 jobs:
   check-branch-name-and-title:
@@ -30,8 +31,8 @@ jobs:
         echo $BRANCH_NAME
         echo $PR_TITLE
 
-        if [[ "$BRANCH_NAME" != feature/$FEATURE_NAME* ]] && 
+        if [[ "$BRANCH_NAME" != $UNFROZEN_PREFIX* ]] && 
            [[ "$PR_TITLE" != fix:* && "$PR_TITLE" != *"[critical]"* ]]; then
-          echo "Error: You can only merge from branches that start with 'feature/$FEATURE_NAME', or PRs titled with 'fix: ' and containing '[critical]'."
+          echo "Error: You can only merge from branches that start with '$UNFROZEN_PREFIX', or PRs titled with 'fix: ' and containing '[critical]'."
           exit 1
         fi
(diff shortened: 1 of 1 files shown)


.github/workflows/code-freeze.yml
```diff
@@ -0,0 +1,35 @@
+name: Code Freeze
+
+on:
+  pull_request:
+    branches:
+      - main
+
```
math411, line 7: Can we add a top level permissions limit for the workflow:
```suggestion

permissions:
  contents: read
```

.github/workflows/code-freeze.yml
```diff
@@ -0,0 +1,35 @@
...
+env:
+  FROZEN: "false"  # Set to "true" to freeze branches, "false" otherwise
```
math411, line 9: Does this make use of action variables? https://github.com/amazon-braket/amazon-braket-sdk-python/settings/variables/actions

ajberdy: No these are just set within the workflow. Is there a benefit to moving to action variables until we want to reuse these values?

ajberdy: I don't mind the workflow for updating action variables, but there's a limitation that you can't set an empty value (for allowing all feature branch merges) and there's no documentation to attach to the place they get updated.

We could update this workflow to use a more generable/intuitive env variable `UNFROZEN_PREFIX` to allow us to specify `feature` or `feature/my-feature` and not rely on an empty value

ajberdy: Updated to use repository variables

math411: For code freezes, will we need to publish PRs each time? If so, I think using the action variables would allow for easier changes and cleaner commit history.

ajberdy: Makes sense, I've update the script to use action variables

math411:
If the branch is frozen will updates from main break?

ajberdy:
> If the branch is frozen will updates from main break?

There are two settings:
- FROZEN: true/false
- UNFROZEN_PREFIX: str

This workflow will run for any PR into main with the following logic:

```
FROZEN: bool
UNFROZEN_PREFIX: str

def can_merge_into_main(pr):
    if not FROZEN:
        return True
    else:
        if pr.branch.startswith(UNFROZEN_PREFIX):
            return True
        elif pr.title.startswith("fix:") and pr.title.contains("[critical]"):
            return True
        else:
            return False
```

math411 (dismissed):

ajberdy:
It seems like a benefit of using variables within the file is if we update the workflow on main, then any PRs will have to update their branch (and rerun the code freeze workflow) before merging, avoiding the race condition we get with repository variables. I think that's enough of a reason to at least move the FROZEN variable to the file. We can leave the unfrozen prefix in the action variables if you think it makes sense, or might as well have them together

math411 (approved):

PR #779, opened 2023-11-01
feature: support FreeParameter binding via `make_bound_program`

### LINKED ISSUE #769: AutoQASM: Add support for FreeParameter binding

laurencap:
*Issue #, if available:*
Closes #769 

*Description of changes:*

Support `make_bound_program`

```
@aq.main
def my_circuit(main_argument: float):
    free_param = FreeParameter("alpha")
    rx(0, main_argument)
    rx(1, free_param)
    measure([0, 1])

my_circuit(3.14/2).make_bound_program({"alpha": 3.14}).to_ir()
```

The name and signature is patterned after `make_bound_circuit`. This name also helps inform the user that the method will make and return a new object, with the original left unchanged.

```
    def make_bound_program(self, param_values: dict[str, float]) -> Program:
    def make_bound_circuit(self, param_values: dict[str, Number], strict: bool = False) -> Circuit:
```

*Testing done:*

New unit tests

## Merge Checklist

#### General

#### Tests

### DIFF
--- a/src/braket/experimental/autoqasm/program/program.py
+++ b/src/braket/experimental/autoqasm/program/program.py
@@ -16,12 +16,14 @@ from __future__ import annotations
 
 import contextlib
 import threading
+import copy
 from collections.abc import Callable, Iterable
 from dataclasses import dataclass
 from enum import Enum
 from typing import Any, Optional, Union
 
 import oqpy.base
+from openqasm3 import ast
 
 from braket.circuits.free_parameter import FreeParameter
 from braket.circuits.free_parameter_expression import FreeParameterExpression
@@ -123,12 +125,17 @@ class Program(SerializableProgram):
         combined_oqpy_program += self._oqpy_program
         return Program(combined_oqpy_program, has_pulse_control=True)
 
-    def make_bound_program(self, param_values: dict[str, float]) -> Program:  # Number
-        # FIXME
-        import copy
-        from openqasm3 import ast
+    def make_bound_program(self, param_values: dict[str, float]) -> Program:
+        """TODO
 
+        Args:
+            param_values (dict[str, float]): TODO
+        Returns:
+            TODO
+        """
+        # We have to copy the program so that we don't modify the original, unbound program
         oqpy_program_copy = copy.deepcopy(self._oqpy_program)
+        # Let's break early if we've processed all the parameter assignments
         params_to_process = set(param_values.keys())
         for state in oqpy_program_copy.stack:
             for i in range(len(state.body)):
@@ -137,13 +144,16 @@ class Program(SerializableProgram):
                     name = inst.identifier.name
                     target = oqpy_program_copy.declared_vars[name]
                     target.init_expression = param_values[name]
+                    # TODO: how do I create the right statement directly?
                     inst = oqpy.Program().declare(target).stack[0].body[0]
                     state.body[i] = inst
                     params_to_process.remove(name)
-                if params_to_process == set():
-                    break
+                    if params_to_process == set():
+                        break
             else:
                 continue
+            # If we broke out of the inner loop, we should break out of the outer loop, too,
+            # because there's no more work to do
             break
 
         return Program(oqpy_program_copy, self._has_pulse_control)
@@ -332,21 +342,21 @@ class ProgramConversionContext:
             name (str): The identifier for the parameter.
         """
         if name not in self._free_parameters:
-            self._free_parameters[name] = (oqpy.FloatVar("input", name=name), FreeParameter(name))
+            self._free_parameters[name] = oqpy.FloatVar("input", name=name)
 
     def get_free_parameters(self) -> list[oqpy.FloatVar]:
         """Return a list of named oqpy.Vars that are used as free parameters in the program."""
-        return list(val[0] for val in self._free_parameters.values())
+        return list(self._free_parameters.values())
 
     def add_io_declarations(self) -> None:
         free_parameters = self.get_free_parameters()
         root_oqpy_program = self.get_oqpy_program(scope=ProgramScope.MAIN)
         for parameter in free_parameters[::-1]:
+            # TODO: is it possible to save a pointer to the declaration statement?
             root_oqpy_program.declare(
                 parameter,
                 to_beginning=True,
             )
-            # TODO: is it possible to save a pointer to the declaration?
 
     def get_target_device(self) -> Optional[Device]:
         """Return the target device for the program, as specified by the user.
--- a/src/braket/experimental/autoqasm/program/program.py
+++ b/src/braket/experimental/autoqasm/program/program.py
@@ -126,12 +126,15 @@ class Program(SerializableProgram):
         return Program(combined_oqpy_program, has_pulse_control=True)
 
     def make_bound_program(self, param_values: dict[str, float]) -> Program:
-        """TODO
+        """Binds FreeParameters based upon their name and values passed in. If parameters
+        share the same name, all the parameters of that name will be set to the mapped value.
 
         Args:
-            param_values (dict[str, float]): TODO
+            param_values (dict[str, float]): A mapping of FreeParameter names
+                to a value to assign to them.
         Returns:
-            TODO
+            Program: Returns a program with all present parameters fixed to their respective
+            values.
         """
         # We have to copy the program so that we don't modify the original, unbound program
         oqpy_program_copy = copy.deepcopy(self._oqpy_program)
@@ -142,11 +145,17 @@ class Program(SerializableProgram):
                 inst = state.body[i]
                 if isinstance(inst, ast.IODeclaration) and inst.identifier.name in param_values:
                     name = inst.identifier.name
-                    target = oqpy_program_copy.declared_vars[name]
-                    target.init_expression = param_values[name]
-                    # TODO: how do I create the right statement directly?
-                    inst = oqpy.Program().declare(target).stack[0].body[0]
-                    state.body[i] = inst
+                    # TODO: which is better?
+                    # Option 1
+                    # target = oqpy_program_copy.declared_vars[name]
+                    # target.init_expression = param_values[name]
+                    # inst = oqpy.Program().declare(target).stack[0].body[0]
+                    # Option 2
+                    del oqpy_program_copy.declared_vars[name]
+                    var = oqpy.FloatVar(init_expression=param_values[name], name=name, needs_declaration=False)
+                    state.body[i] = var.make_declaration_statement(oqpy_program_copy)
+                    # Option 3: is there a better way?
+
                     params_to_process.remove(name)
                     if params_to_process == set():
                         break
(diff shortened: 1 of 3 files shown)


src/braket/experimental/autoqasm/program/program.py
```diff
@@ -123,6 +125,48 @@ def with_calibrations(self, gate_calibrations: Union[Callable, list[Callable]])
...
+        oqpy_program_copy = copy.deepcopy(self._oqpy_program)
+        # Let's break early if we've processed all the parameter assignments
+        params_to_process = set(param_values.keys())
+        for state in oqpy_program_copy.stack:
+            for i in range(len(state.body)):
+                inst = state.body[i]
+                if isinstance(inst, ast.IODeclaration) and inst.identifier.name in param_values:
+                    name = inst.identifier.name
+                    # TODO: which is better?
+                    # Option 1
```
laurencap, line 149: Looking for feedback on this

src/braket/experimental/autoqasm/program/program.py
```diff
@@ -123,6 +125,48 @@ def with_calibrations(self, gate_calibrations: Union[Callable, list[Callable]])
...
+                to a value to assign to them.
+        Returns:
+            Program: Returns a program with all present parameters fixed to their respective
+            values.
+        """
+        # We have to copy the program so that we don't modify the original, unbound program
+        oqpy_program_copy = copy.deepcopy(self._oqpy_program)
+        # Let's break early if we've processed all the parameter assignments
+        params_to_process = set(param_values.keys())
+        for state in oqpy_program_copy.stack:
```
laurencap, line 143: It shouldn't be too bad going through these, because all the inputs should be at the start of the program. We break after we've seen the params requested. So, the worst case scenario is if there's a param that doesn't appear in the program.

We can enable a "strict" option like the circuit class has, then check against `program_conversion_context._free_parameters` (would have to pass that at Program creation). That would prevent the slow case.

I also considered trying to maintain a pointer to the IODeclaration when it's created, but that only gives me a reference to the stmt, not the location of the statement in `state.body`, which is what I need here. I could try to use `list.index` for that, but I'm not sure it's worth the extra complex (and hacky) code in `.add_io_declarations`

src/braket/experimental/autoqasm/program/program.py
```diff
@@ -123,6 +125,39 @@ def with_calibrations(self, gate_calibrations: Union[Callable, list[Callable]])
...
+        Returns:
+            Program: Returns a program with all present parameters fixed to their respective
+            values.
+        """
+        # Copy the program so that we don't modify the original program
+        oqpy_program_copy = copy.deepcopy(self._oqpy_program)
+        # Break early if all the parameter assignments have been processed
+        params_to_process = set(param_values.keys())
+
+        # Parameter expressions only occur at the top level scope
```
laurencap, line 144:
```suggestion
        # Input declarations only occur at the top level scope
```

src/braket/experimental/autoqasm/program/program.py
```diff
@@ -123,6 +125,39 @@ def with_calibrations(self, gate_calibrations: Union[Callable, list[Callable]])
...
+        # Input declarations only occur at the top level scope
+        state = oqpy_program_copy.stack[0]
+        for i in range(len(state.body)):
+            inst = state.body[i]
+            if isinstance(inst, ast.IODeclaration) and inst.identifier.name in param_values:
+                name = inst.identifier.name
+                target = oqpy_program_copy.declared_vars[name]
+                target.init_expression = param_values[name]
+                new_inst = oqpy.Program().declare(target).stack[0].body[0]
+                state.body[i] = new_inst
```
rmshaffer, line 153: Would this work?
```
state.body[i] = target.make_declaration_statement(bound_oqpy_program)
```

src/braket/experimental/autoqasm/program/program.py
```diff
@@ -123,6 +125,39 @@ def with_calibrations(self, gate_calibrations: Union[Callable, list[Callable]])
...
+
+                params_to_process.remove(name)
+                if params_to_process == set():
```
rmshaffer, line 156:
```suggestion
                if not params_to_process:
```

src/braket/experimental/autoqasm/program/program.py
```diff
@@ -313,6 +348,16 @@ def get_free_parameters(self) -> list[oqpy.FloatVar]:
         """Return a list of named oqpy.Vars that are used as free parameters in the program."""
         return list(self._free_parameters.values())
 
+    def add_io_declarations(self) -> None:
+        """Add input and output declaration statements to the program."""
+        free_parameters = self.get_free_parameters()
+        root_oqpy_program = self.get_oqpy_program(scope=ProgramScope.MAIN)
+        for parameter in free_parameters[::-1]:
```
rmshaffer, line 355: [nit] to avoid an extra variable declaration and a list copy:
```suggestion
        root_oqpy_program = self.get_oqpy_program(scope=ProgramScope.MAIN)
        for parameter in reversed(self.get_free_parameters()):
```

test/unit_tests/braket/experimental/autoqasm/test_parameters.py
```diff
@@ -289,3 +289,142 @@ def my_program():
...
+    """Test that binding behaves well on empty programs."""
+
+    @aq.main
+    def empty_program():
+        pass
+
+    qasm = empty_program().to_ir()
+    bound_program1 = empty_program().make_bound_program({}).to_ir()
+    bound_program2 = empty_program().make_bound_program({"alpha": 0.5}).to_ir()
+    assert qasm == bound_program1 == bound_program2
```
rmshaffer, line 430: cool, I had somehow never realized that Python allowed chaining equality tests like this.

src/braket/experimental/autoqasm/program/program.py
```diff
@@ -123,6 +125,39 @@ def with_calibrations(self, gate_calibrations: Union[Callable, list[Callable]])
...
+                target = oqpy_program_copy.declared_vars[name]
+                target.init_expression = param_values[name]
+                new_inst = oqpy.Program().declare(target).stack[0].body[0]
+                state.body[i] = new_inst
+
+                params_to_process.remove(name)
+                if params_to_process == set():
+                    break
+
+        return Program(oqpy_program_copy, self._has_pulse_control)
```
rmshaffer, line 159: what if `params_to_process` still has element(s) at this point? it means the user passed in some variable names that aren't in the program - should we raise an exception, or at least print a warning?

rmshaffer (changes requested):
I like the API choice! Left a few minor comments, and also it would be good to get @jcjaskula-aws's opinion on the oqpy program manipulation here.

src/braket/experimental/autoqasm/program/program.py
```diff
@@ -123,6 +125,39 @@ def with_calibrations(self, gate_calibrations: Union[Callable, list[Callable]])
         combined_oqpy_program += self._oqpy_program
         return Program(combined_oqpy_program, has_pulse_control=True)
 
+    def make_bound_program(self, param_values: dict[str, float]) -> Program:
+        """Binds FreeParameters based upon their name and values passed in. If parameters
+        share the same name, all the parameters of that name will be set to the mapped value.
```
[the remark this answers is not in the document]
laurencap: I suppose so? I copied this from `make_bound_circuit`, I should try to test that behavior. Do we want to match the circuit behavior, or should we fail on such inputs?

test/unit_tests/braket/experimental/autoqasm/test_parameters.py
laurencap: I wasn't sure either :D I followed a whim

src/braket/experimental/autoqasm/program/program.py
laurencap: @rmshaffer 
> what if params_to_process still has element(s) at this point? it means the user passed in some variable names that aren't in the program - should we raise an exception, or at least print a warning?

We could add a `strict` option here, like circuits. Shall we? I feel like it's probably what users would prefer.

src/braket/experimental/autoqasm/program/program.py
laurencap: replied in another thread

src/braket/experimental/autoqasm/program/program.py
laurencap: That does seem to be a better implementation. For some unfathomable reason, it changes the output of an `np.pi` assignment to:
```
float[64] beta = 3.141592653589793;
```
whereas the current solution results in:
```
float[64] beta = pi;
```
There must be a reason for this, might be worth looking into some more.

src/braket/experimental/autoqasm/program/program.py
```diff
@@ -313,6 +348,16 @@ def get_free_parameters(self) -> list[oqpy.FloatVar]:
         """Return a list of named oqpy.Vars that are used as free parameters in the program."""
         return list(self._free_parameters.values())
 
+    def add_io_declarations(self) -> None:
+        """Add input and output declaration statements to the program."""
+        free_parameters = self.get_free_parameters()
+        root_oqpy_program = self.get_oqpy_program(scope=ProgramScope.MAIN)
+        for parameter in free_parameters[::-1]:
+            root_oqpy_program.declare(
```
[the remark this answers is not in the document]
laurencap: `to_ir` is called before `make_bound_program`

src/braket/experimental/autoqasm/program/program.py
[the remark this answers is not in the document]
laurencap: Thinking a bit more, I think this just means that `FreeParameter("theta") == FreeParameter("theta")`. Multiple parameters with the same name is the same as one parameter used multiple times. I feel like this sentence just makes it confusing though. I can remove it here

src/braket/experimental/autoqasm/program/program.py
rmshaffer: Yes I think having a `strict` option here would make sense.

By the way, the docstring for the `strict` param of `Circuit.make_bound_circuit` seems wrong:

https://github.com/amazon-braket/amazon-braket-sdk-python/blob/4446f4bd03f29d1f7d09e433e2727f13a55b5077/src/braket/circuits/circuit.py#L858-L859

but the behavior is actually that it raises an exception if *any* of the parameters do not appear:

https://github.com/amazon-braket/amazon-braket-sdk-python/blob/4446f4bd03f29d1f7d09e433e2727f13a55b5077/src/braket/circuits/circuit.py#L884-L886

laurencap: Yes, I almost included an update to the circuits docstring here, but I will open a second PR to `main` so we can update that documentation now

src/braket/experimental/autoqasm/program/program.py
```diff
@@ -123,6 +124,34 @@ def with_calibrations(self, gate_calibrations: Union[Callable, list[Callable]])
...
+            Program: Returns a program with all present parameters fixed to their respective
+            values.
+        """
+        # Copy the program so that we don't modify the original program
+        bound_oqpy_program = copy.deepcopy(self._oqpy_program)
+        for name, value in param_values.items():
+            if name in bound_oqpy_program.undeclared_vars:
+                target = bound_oqpy_program.undeclared_vars[name]
+                assert target.init_expression == "input", "Only free parameters can be bound."
+                target.init_expression = value
```
rmshaffer, line 149: this is nice!

rmshaffer (approved):

PR #782, opened 2023-11-02
fix: fixing some type hints for optional params

krneta:
*Issue #, if available:*

*Description of changes:*

*Testing done:*

## Merge Checklist

#### General

#### Tests

### DIFF
--- a/src/braket/annealing/problem.py
+++ b/src/braket/annealing/problem.py
@@ -37,16 +37,16 @@ class Problem:
     def __init__(
         self,
         problem_type: ProblemType,
-        linear: Dict[int, float] = None,
-        quadratic: Dict[Tuple[int, int], float] = None,
+        linear: Dict[int, float] | None = None,
+        quadratic: Dict[Tuple[int, int], float] | None = None,
     ):
         """
 
         Args:
             problem_type (ProblemType): The type of annealing problem
-            linear (Dict[int, float]): The linear terms of this problem,
+            linear (Dict[int, float] | None): The linear terms of this problem,
                 as a map of variable to coefficient
-            quadratic (Dict[Tuple[int, int], float]): The quadratic terms of this problem,
+            quadratic (Dict[Tuple[int, int], float] | None): The quadratic terms of this problem,
                 as a map of variables to coefficient
 
         Examples:
--- a/src/braket/aws/aws_quantum_job.py
+++ b/src/braket/aws/aws_quantum_job.py
@@ -64,22 +64,22 @@ class AwsQuantumJob(QuantumJob):
         cls,
         device: str,
         source_module: str,
-        entry_point: str = None,
-        image_uri: str = None,
-        job_name: str = None,
-        code_location: str = None,
-        role_arn: str = None,
+        entry_point: str | None = None,
+        image_uri: str | None = None,
+        job_name: str | None = None,
+        code_location: str | None = None,
+        role_arn: str | None = None,
         wait_until_complete: bool = False,
-        hyperparameters: dict[str, Any] = None,
-        input_data: str | dict | S3DataSourceConfig = None,
-        instance_config: InstanceConfig = None,
-        distribution: str = None,
-        stopping_condition: StoppingCondition = None,
-        output_data_config: OutputDataConfig = None,
-        copy_checkpoints_from_job: str = None,
-        checkpoint_config: CheckpointConfig = None,
-        aws_session: AwsSession = None,
-        tags: dict[str, str] = None,
+        hyperparameters: dict[str, Any] | None = None,
+        input_data: str | dict | S3DataSourceConfig | None = None,
+        instance_config: InstanceConfig | None = None,
+        distribution: str | None = None,
+        stopping_condition: StoppingCondition | None = None,
+        output_data_config: OutputDataConfig | None = None,
+        copy_checkpoints_from_job: str | None = None,
+        checkpoint_config: CheckpointConfig | None = None,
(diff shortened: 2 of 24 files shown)


src/braket/annealing/problem.py
```diff
@@ -37,16 +37,16 @@ class Problem:
     def __init__(
         self,
         problem_type: ProblemType,
-        linear: Dict[int, float] = None,
-        quadratic: Dict[Tuple[int, int], float] = None,
+        linear: Dict[int, float] | None = None,
```
math411, line 40:
```suggestion
        linear: dict[int, float] | None = None,
```

Python supports `dict`, `list`, `tuple` as a built in type now. This can also be removed from imports.

src/braket/aws/aws_session.py
```diff
@@ -716,7 +716,7 @@ def describe_log_streams(
         self,
         log_group: str,
         log_stream_prefix: str,
-        limit: int = None,
+        limit: Optional[int] = None,
```
math411, line 719:
```suggestion
        limit: int | None = None,
```

src/braket/aws/aws_session.py
```diff
@@ -725,7 +725,7 @@ def describe_log_streams(
         Args:
             log_group (str): Name of the log group.
             log_stream_prefix (str): Prefix for log streams to include.
-            limit (int): Limit for number of log streams returned.
+            limit (Optional[int]): Limit for number of log streams returned.
```
math411, line 728:
```suggestion
            limit (int | None): Limit for number of log streams returned.
```

math411:
Starting with Python 3.10, implicitly defining optional is not needed. We should check which we wish to use:
```
int | none
or
Optional[int]
```

src/braket/aws/aws_quantum_job.py
```diff
@@ -64,22 +64,22 @@ def create(
         cls,
         device: str,
```
ajberdy, line 65: This needs to be `str | None` as this parameter is explicitly allowed to be None, as mentioned in the doc string

ajberdy:
> Starting with Python 3.10, implicitly defining optional is not needed. We should check which we wish to use:
> 
> ```
> int | none
> or
> Optional[int]
> ```

+1 to `int | None`, since `Optional` is confusing and doesn't actually align with whether a parameter is optional.

ajberdy:
What's the scope of this pull request? I know there are other files (`hybrid_job.py` at least) that require updates. Is this meant to be all of the updates or just some?

krneta:
> Starting with Python 3.10, implicitly defining optional is not needed. We should check which we wish to use:
> 
> ```
> int | none
> or
> Optional[int]
> ```

I use Optional[] if other parameters in this function already use the Optional syntax. I don't think we need to go through all the places where we use that syntax and change it to | None.

krneta:
> > Starting with Python 3.10, implicitly defining optional is not needed. We should check which we wish to use:
> > ```
> > int | none
> > or
> > Optional[int]
> > ```
> 
> +1 to `int | None`, since `Optional` is confusing and doesn't actually align with whether a parameter is optional.

See comment above

math411:
> What's the scope of this pull request? I know there are other files (`hybrid_job.py` at least) that require updates. Is this meant to be all of the updates or just some?

I am working on adding linter checks for these missing types so I am not opposed to merging this as long as we have a consistent convention. This saves me an immense amount of time later so having some finished is better than none!

ajberdy:
> > Starting with Python 3.10, implicitly defining optional is not needed. We should check which we wish to use:
> > ```
> > int | none
> > or
> > Optional[int]
> > ```
> 
> I use Optional[] if other parameters in this function already use the Optional syntax. I don't think we need to go through all the places where we use that syntax and change it to | None.

I think it likely makes sense to converge on a single style that we stick to, though if we don't care to prioritize that update we could allow both methods in the meantime

krneta:
> What's the scope of this pull request? I know there are other files (`hybrid_job.py` at least) that require updates. Is this meant to be all of the updates or just some?

This should be all of them, it's weird that it wasn't picked up. I'll investigate.

krneta:
> > What's the scope of this pull request? I know there are other files (`hybrid_job.py` at least) that require updates. Is this meant to be all of the updates or just some?
> 
> This should be all of them, it's weird that it wasn't picked up. I'll investigate.

Oh, I see. I'm not handling keyword only args... have to fix that.

src/braket/annealing/problem.py
krneta: Yes, but I'd prefer not to make those types of changes in this PR (this PR is broad enough without them).

src/braket/aws/aws_session.py
krneta: I use Optional[] if other parameters in this function already use the Optional syntax. I don't think we need to go through all the places where we use that syntax and change it to | None.

src/braket/jobs/hybrid_job.py
```diff
@@ -46,22 +46,22 @@
 def hybrid_job(
     *,
     device: str,
```
ajberdy, line 48: This annotation should be `str | None` (no default), since `None` is explicitly an allowed value for `device`, as called out in the doc string

src/braket/annealing/problem.py
ajberdy: Maybe we can add a rule to the linter that flags imports of `Dict`, `List`, etc from `typing`

ajberdy (approved):

PR #783, opened 2023-11-03
Fix: Invalidate importlib cache for decorator job

shpface:
*Issue #, if available:*

*Description of changes:*

This change will make a call to invalidate the internal caches in importlib when using decorator jobs. From [imprortlib docs](https://docs.python.org/3/library/importlib.html#importlib.invalidate_caches),
> This function should be called if any modules are created/installed while your program is running to guarantee all finders will notice the new module’s existence.

This is the situation when the hybrid job decorator creates a new source module to create a new job while the program is running.

*Testing done:*
* local unit & integ tests
* https://github.com/amazon-braket/amazon-braket-examples/pull/447

## Merge Checklist

#### General

#### Tests

### DIFF
--- a/src/braket/jobs/quantum_job_creation.py
+++ b/src/braket/jobs/quantum_job_creation.py
@@ -330,6 +330,7 @@ def _validate_entry_point(source_module_path: Path, entry_point: str) -> None:
     sys.path.append(str(source_module_path.parent))
     try:
         # second argument allows relative imports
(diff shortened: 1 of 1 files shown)


krneta (approved):
The thing is that the issue also goes away if I put a time.sleep() in the loop. So it could be that all removing the cache does is slow it down enough that it doesn't reproduce. But we may as well try it and see if it comes back.

PR #787, opened 2023-11-07
feature: Add get_compiled_circuit convenience method

### LINKED ISSUE #780: Convenience method: get compiled circuit
**Describe the feature you'd like**
I would like a convenience method on `task.result()` to grab the compiled circuit, or return None if it's not available.

**How would this feature be used? Please describe.**
Previous method:
https://docs.aws.amazon.com/braket/latest/developerguide/braket-compiled-circuits-inspecting.html

Ex:
```
compiled_program = task.result().additional_metadata.rigettiMetadata.compiledProgram
compiledProgram = task_result.additional_metadata.oqcMetadata.compiledProgram
```

Proposed method:

```
compiled_program = task.result().get_compiled_program()
```

Another benefit of this feature is that the user does not have to make adjustments to their code depending on the provider. The method can handle checking either `rigettiMetadata` or `oqcMetadata`, as appropriate.

**Describe alternatives you've considered**
 - One alternative is to leave it as is.
 - It can also be helpful to get the OpenQASM source that is associated with the task. We could add another convenience method for that.
 - Due to the blocking nature of calling `task.result()`, I don't recommend putting the convenience method in the task itself.
 - We could call it something else, like `compiled_artifact`. I'm not attached to the name I proposed -- there may be better names.

**Additional context**

laurencap:
Add `get_compiled_circuit` to GateModelQuantumTaskResult

*Issue #, if available:*
Closes #780 

*Description of changes:*

Proposed method:

```
compiled_program = task.result().get_compiled_program()
```

For Rigetti and OQC tasks, this could be used instead of the method described here: https://docs.aws.amazon.com/braket/latest/developerguide/braket-compiled-circuits-inspecting.html

*Testing done:*
New unit tests, local testing with successful quantum tasks ran on hardware

## Merge Checklist

#### General

#### Tests

### DIFF
--- a/src/braket/tasks/gate_model_quantum_task_result.py
+++ b/src/braket/tasks/gate_model_quantum_task_result.py
@@ -134,6 +134,21 @@ class GateModelQuantumTaskResult:
             return self.task_metadata.id == other.task_metadata.id
         return NotImplemented
 
+    def get_compiled_circuit(self) -> str:
+        """
+        Get the compiled circuit, if one is available.
+
+        Returns:
+            str: The compiled circuit or None.
+        """
+        metadata = self.additional_metadata
+        if not metadata:
+            return
+        if metadata.rigettiMetadata:
+            return metadata.rigettiMetadata.compiledProgram
+        elif metadata.oqcMetadata:
+            return metadata.oqcMetadata.compiledProgram
+
     @staticmethod
     def measurement_counts_from_measurements(measurements: np.ndarray) -> Counter:
         """
--- a/src/braket/tasks/gate_model_quantum_task_result.py
+++ b/src/braket/tasks/gate_model_quantum_task_result.py
@@ -143,11 +143,13 @@ class GateModelQuantumTaskResult:
         """
         metadata = self.additional_metadata
         if not metadata:
-            return
+            return None
         if metadata.rigettiMetadata:
             return metadata.rigettiMetadata.compiledProgram
         elif metadata.oqcMetadata:
             return metadata.oqcMetadata.compiledProgram
+        else:
+            return None
 
     @staticmethod
     def measurement_counts_from_measurements(measurements: np.ndarray) -> Counter:
--- a/src/braket/tasks/gate_model_quantum_task_result.py
+++ b/src/braket/tasks/gate_model_quantum_task_result.py
@@ -134,12 +134,12 @@ class GateModelQuantumTaskResult:
             return self.task_metadata.id == other.task_metadata.id
         return NotImplemented
 
-    def get_compiled_circuit(self) -> str:
+    def get_compiled_circuit(self) -> Optional[str]:
         """
         Get the compiled circuit, if one is available.
 
         Returns:
-            str: The compiled circuit or None.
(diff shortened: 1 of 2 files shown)


src/braket/tasks/gate_model_quantum_task_result.py
```diff
@@ -134,6 +134,21 @@ def __eq__(self, other) -> bool:
             return self.task_metadata.id == other.task_metadata.id
         return NotImplemented
 
+    def get_compiled_circuit(self) -> str:
```
laurencap, line 137: TODO: Update to `Optional[str]`

src/braket/tasks/gate_model_quantum_task_result.py
```diff
@@ -134,6 +134,21 @@ def __eq__(self, other) -> bool:
...
+    def get_compiled_circuit(self) -> Optional[str]:
+        """
+        Get the compiled circuit, if one is available.
+
+        Returns:
+            Optional[str]: The compiled circuit or None.
+        """
+        metadata = self.additional_metadata
+        if not metadata:
+            return
```
ajberdy, line 146: Return values should be consistent, in this case, all returning an expression (`return None` here) ([pep 8](https://peps.python.org/pep-0008/#:~:text=Be%20consistent%20in,function%20(if%20reachable))).  Technically it would also have us add a return None at the end of the function, but I don't believe we're consistent about that rule.

We could restructure the logic to work with that by inverting the condition
```
if metadata:
    if metadata.rigettiMetadata:
        ...
    elif ...:
        ...
return None
```

test/unit_tests/braket/tasks/test_gate_model_quantum_task_result.py
```diff
@@ -234,6 +331,29 @@ def test_openqasm_shots_calculate_result_types(openqasm_result_obj_shots):
 ]
 
 
+def test_get_compiled_circuit_rigetti(result_rigetti, quil_program):
+    """Test get_compiled_circuit method."""
+    assert result_rigetti.get_compiled_circuit() == quil_program
+
+
+def test_get_compiled_circuit_oqc(result_oqc, qasm2_program):
```
ajberdy, line 339: nit: several of these tests/fixtures can be parametrized with pytest instead of duplicated

laurencap: It looks like I should be able to do something like this -- but afaiu parameter inputs can't be fixtures. Do you know how to do it, beyond accepting the fixture name as a string and looking up the function from it? If you have other more specific recommendations please let me know.

```
@pytest.mark.parametrize("result,program", [(result_rigetti, quil_program), (result_oqc, qasm2_program)])
def test_get_compiled_circuit(result, program):
    """Test get_compiled_circuit method."""
    assert result.get_compiled_circuit() == program
```

laurencap: Updated the result fixtures to remove duplicated code

ajberdy (approved):

PR #789, opened 2023-11-07
feature: Add support for FreeParameters in AutoQASM conditional statements

### LINKED ISSUE #785: AutoQASM: support for FreeParameters in conditional statements
This is a follow up from https://github.com/amazon-braket/amazon-braket-sdk-python/issues/757

`FreeParameter` is supported now for gate parameters, but not yet for conditional statements. For example:

```
@aq.main
def parametric(val: float | FreeParameter):
    threshold = 0.9
    if val > threshold:
        x(0)
    measure(0)

parametric(FreeParameter("val"))
```

This issue adds support for this use case. Right now, we get `TypeError: '>' not supported between instances of 'FreeParameter' and 'float'`

laurencap:
*Issue #, if available:*
Closes #785 

*Description of changes:*
This is a follow up from https://github.com/amazon-braket/amazon-braket-sdk-python/issues/757

This adds support for FreeParameters in conditional statements. For example:

```
@aq.main
def parametric(val: float | FreeParameter):
    threshold = 0.9
    if val > threshold:
        x(0)
    measure(0)

parametric(FreeParameter("val"))
```

*Details*
 - New `ComparisonTransformer` to replace comparison operators with functions that we can later process ourselves
 - Comparison operator processing for <, <=, >, >=
 - arguments processed by `autoqasm/operators/` (both logical and comparison stmts) now check for FreeParameters. If found, they're registered with the program conversion context and converted to the newly associated FloatVar
 - Minor API cleanup for program conversion context
 - transpiler updated to use the new `ComparisonTransformer`
 - Minor improvements in types: if a name kwarg is supplied, it is no longer overwritten.
 - Lots of new tests

*Testing done:*
New unit tests

## Merge Checklist

#### General

#### Tests

### DIFF
--- a/src/braket/experimental/autoqasm/converters/comparisons.py
+++ b/src/braket/experimental/autoqasm/converters/comparisons.py
@@ -1,0 +1,70 @@
+# Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You
+# may not use this file except in compliance with the License. A copy of
+# the License is located at
+#
+#     http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+
+"""Converters for comparison nodes."""
+
+import ast
+
+import gast
+
+from braket.experimental.autoqasm.autograph.core import ag_ctx, converter
+from braket.experimental.autoqasm.autograph.pyct import templates
+
+COMPARISON_OPERATORS = {
+    gast.Lt: "ag__.lt_",
+    gast.LtE: "ag__.lteq_",
+    gast.Gt: "ag__.gt_",
+    gast.GtE: "ag__.gteq_",
+}
+
+
+class ComparisonTransformer(converter.Base):
+    """Transformer for comparison nodes."""
+
+    def visit_Compare(self, node: ast.stmt) -> ast.stmt:
+        """Transforms a comparison node.
+
+        Args:
+            node (ast.stmt): AST node to transform.
+
+        Returns:
+            ast.stmt: Transformed node.
+        """
+        node = self.generic_visit(node)
+
+        op_type = type(node.ops[0])
+        if op_type not in COMPARISON_OPERATORS:
+            return node
+
+        template = f"{COMPARISON_OPERATORS[op_type]}(lhs_, rhs_)"
+
+        return templates.replace(
+            template,
+            lhs_=node.left,
+            rhs_=node.comparators[0],
+            original=node,
+        )[0].value
+
+
+def transform(node: ast.stmt, ctx: ag_ctx.ControlStatusCtx) -> ast.stmt:
+    """Transform comparison nodes.
+
+    Args:
+        node (ast.stmt): AST node to transform.
+        ctx (ag_ctx.ControlStatusCtx): Transformer context.
+
+    Returns:
+        ast.stmt: Transformed node.
+    """
+
+    return ComparisonTransformer(ctx).visit(node)
--- a/src/braket/experimental/autoqasm/operators/comparisons.py
+++ b/src/braket/experimental/autoqasm/operators/comparisons.py
@@ -14,10 +14,7 @@
 
 """Operators for comparison operators: <, <=, >, and >=."""
 
-from collections.abc import Callable
 from typing import Any, Union
-
-import oqpy.base
 
 from braket.experimental.autoqasm import program
 from braket.experimental.autoqasm import types as aq_types
--- a/src/braket/experimental/autoqasm/operators/comparisons.py
+++ b/src/braket/experimental/autoqasm/operators/comparisons.py
@@ -1,0 +1,137 @@
+# Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You
+# may not use this file except in compliance with the License. A copy of
+# the License is located at
+#
+#     http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+
+
+"""Operators for comparison operators: <, <=, >, and >=."""
+
+from collections.abc import Callable
+from typing import Any, Union
+
+import oqpy.base
+
+from braket.experimental.autoqasm import program
+from braket.experimental.autoqasm import types as aq_types
+
+from .utils import _convert_parameters
+
+
+def lt_(a: Any, b: Any) -> Union[bool, aq_types.BoolVar]:
+    """Functional form of "<".
+
+    Args:
+        a (Any): Callable that returns the first expression.
+        b (Any): Callable that returns the second expression.
+
+    Returns:
+        Union[bool, BoolVar]: Whether the first expression is less than the second.
+    """
+    if aq_types.is_qasm_type(a) or aq_types.is_qasm_type(b):
+        return _aq_lt(a, b)
+    else:
+        return a < b
+
+
+def _aq_lt(a: Any, b: Any) -> aq_types.BoolVar:
+    program_conversion_context = program.get_program_conversion_context()
+    program_conversion_context.register_args([a, b])
+    a, b = _convert_parameters(a, b)
+
+    oqpy_program = program_conversion_context.get_oqpy_program()
+    result = aq_types.BoolVar()
+    oqpy_program.declare(result)
+    oqpy_program.set(result, a < b)
+    return result
+
+
+def lteq_(a: Any, b: Any) -> Union[bool, aq_types.BoolVar]:
+    """Functional form of "<=".
+
+    Args:
+        a (Any): Callable that returns the first expression.
+        b (Any): Callable that returns the second expression.
+
+    Returns:
+        Union[bool, BoolVar]: Whether the first expression is less than or equal to the second.
+    """
+    if aq_types.is_qasm_type(a) or aq_types.is_qasm_type(b):
+        return _aq_lteq(a, b)
+    else:
+        return a <= b
+
+
+def _aq_lteq(a: Any, b: Any) -> aq_types.BoolVar:
+    program_conversion_context = program.get_program_conversion_context()
+    program_conversion_context.register_args([a, b])
+    a, b = _convert_parameters(a, b)
+
+    oqpy_program = program_conversion_context.get_oqpy_program()
+    result = aq_types.BoolVar()
+    oqpy_program.declare(result)
+    oqpy_program.set(result, a <= b)
+    return result
+
+
+def gt_(a: Any, b: Any) -> Union[bool, aq_types.BoolVar]:
+    """Functional form of ">".
+
+    Args:
+        a (Any): Callable that returns the first expression.
+        b (Any): Callable that returns the second expression.
+
+    Returns:
+        Union[bool, BoolVar]: Whether the first expression greater than the second.
+    """
+    if aq_types.is_qasm_type(a) or aq_types.is_qasm_type(b):
+        return _aq_gt(a, b)
+    else:
+        return a > b
+
+
+def _aq_gt(a: Any, b: Any) -> aq_types.BoolVar:
+    program_conversion_context = program.get_program_conversion_context()
+    program_conversion_context.register_args([a, b])
+    a, b = _convert_parameters(a, b)
+
+    oqpy_program = program_conversion_context.get_oqpy_program()
+    result = aq_types.BoolVar()
+    oqpy_program.declare(result)
+    oqpy_program.set(result, a > b)
+    return result
+
+
+def gteq_(a: Any, b: Any) -> Union[bool, aq_types.BoolVar]:
+    """Functional form of ">=".
+
+    Args:
+        a (Any): Callable that returns the first expression.
+        b (Any): Callable that returns the second expression.
+
+    Returns:
+        Union[bool, BoolVar]: Whether the first expression greater than or equal to the second.
+    """
+    if aq_types.is_qasm_type(a) or aq_types.is_qasm_type(b):
+        return _aq_gteq(a, b)
+    else:
+        return a >= b
+
+
+def _aq_gteq(a: Any, b: Any) -> aq_types.BoolVar:
+    program_conversion_context = program.get_program_conversion_context()
+    program_conversion_context.register_args([a, b])
+    a, b = _convert_parameters(a, b)
+
+    oqpy_program = program_conversion_context.get_oqpy_program()
+    result = aq_types.BoolVar()
+    oqpy_program.declare(result)
+    oqpy_program.set(result, a >= b)
+    return result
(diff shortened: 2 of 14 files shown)


src/braket/experimental/autoqasm/converters/comparisons.py
```diff
@@ -0,0 +1,70 @@
...
+# the License is located at
+#
+#     http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+
+"""Converters for comparison nodes."""
```
laurencap, line 14: This grabs anything like `"a < b"` and replaces it with a function call that can later be hooked into and modified as we like.

In particular, `"a < b"` becomes `"ag__.lt_(a, b)"` which will then call the function `lt_` defined now in `operators/comparisons.py`

src/braket/experimental/autoqasm/transpiler/transpiler.py
```diff
@@ -139,6 +144,7 @@ def transform_ast(
         # canonicalization creates.
         node = continue_statements.transform(node, ctx)
         node = return_statements.transform(node, ctx)
+        node = comparisons.transform(node, ctx)
```
laurencap, line 147: This is how we insert the `visit_Compare` conversion from `converters/comparisons.py`

src/braket/experimental/autoqasm/types/types.py
```diff
@@ -81,16 +85,19 @@ def __init__(self, *args, **kwargs):
...
-        self.name = program.get_program_conversion_context().next_var_name(oqpy.BoolVar)
+        if "name" not in kwargs:
+            self.name = program.get_program_conversion_context().next_var_name(oqpy.BoolVar)
 
 
 class FloatVar(oqpy.FloatVar):
     def __init__(self, *args, **kwargs):
         super(FloatVar, self).__init__(*args, **kwargs)
-        self.name = program.get_program_conversion_context().next_var_name(oqpy.FloatVar)
+        if "name" not in kwargs:
```
laurencap, line 95: if I do `aq.FloatVar(name="name", val)`, I don't want that name to be replaced

src/braket/experimental/autoqasm/operators/utils.py
```diff
@@ -0,0 +1,45 @@
...
+"Utility methods for operators."
+
+from typing import Any
+
+from braket.circuits import FreeParameter
+from braket.experimental.autoqasm import program
+from braket.experimental.autoqasm import types as aq_types
+
+
+def _register_and_convert_parameters(*args: list[Any]) -> list[aq_types.FloatVar]:
```
laurencap, line 24: This has side effects. Usually I wouldn't be too much of a fan of compound side effect + return val functions like this, but it encapsulates free parameter handling for all the operator code. Also keeps the other code cleaner.

test/unit_tests/braket/experimental/autoqasm/test_api.py
```diff
@@ -131,7 +131,9 @@ def do_h(int[32] q) {
 }
 def recursive_h(int[32] q) {
     do_h(q);
-    if (q > 0) {
+    bool __bool_0__;
```
laurencap, line 134: @rmshaffer one of your PRs simplified code like this recently -- should I investigate this more and see if I can get simplified output from this?

src/braket/experimental/autoqasm/operators/comparisons.py
```diff
@@ -0,0 +1,126 @@
...
+
+def gt_(a: Any, b: Any) -> Union[bool, aq_types.BoolVar]:
+    """Functional form of ">".
+
+    Args:
+        a (Any): Callable that returns the first expression.
+        b (Any): Callable that returns the second expression.
+
+    Returns:
+        Union[bool, BoolVar]: Whether the first expression greater than the second.
```
rmshaffer, line 85:
```suggestion
        Union[bool, BoolVar]: Whether the first expression is greater than the second.
```

src/braket/experimental/autoqasm/operators/comparisons.py
```diff
@@ -0,0 +1,126 @@
...
+
+def gteq_(a: Any, b: Any) -> Union[bool, aq_types.BoolVar]:
+    """Functional form of ">=".
+
+    Args:
+        a (Any): Callable that returns the first expression.
+        b (Any): Callable that returns the second expression.
+
+    Returns:
+        Union[bool, BoolVar]: Whether the first expression greater than or equal to the second.
```
rmshaffer, line 111:
```suggestion
        Union[bool, BoolVar]: Whether the first expression is greater than or equal to the second.
```

src/braket/experimental/autoqasm/operators/comparisons.py
```diff
@@ -0,0 +1,126 @@
...
+
+from .utils import _register_and_convert_parameters
+
+
+def lt_(a: Any, b: Any) -> Union[bool, aq_types.BoolVar]:
+    """Functional form of "<".
+
+    Args:
+        a (Any): Callable that returns the first expression.
+        b (Any): Callable that returns the second expression.
```
rmshaffer, line 30: For all of the docstrings in this file - it doesn't look like any of these are `Callable`, is this just a copy/paste artifact?
```suggestion
        a (Any): The first expression.
        b (Any): The second expression.
```

src/braket/experimental/autoqasm/operators/utils.py
rmshaffer, line 24: should `args` be annotated as a `tuple` rather than a `list`? also the return type hint is wrong in the case that only a single item is returned.

src/braket/experimental/autoqasm/operators/utils.py
```diff
@@ -0,0 +1,45 @@
...
+    Notes: Adding a parameter to the registry twice is safe. Conversion is a pass through
+    for non-FreeParameter inputs.
+
+    FloatVars are more compatible with the program conversion operations.
+
+    Returns:
+        list[FloatVar]: FloatVars for program conversion.
+    """
+    program_conversion_context = program.get_program_conversion_context()
+    program_conversion_context.register_args(args)  # TODO could be one item
```
rmshaffer, line 37: Do you need to address this TODO? Maybe set `args = list(args)` before calling this?

src/braket/experimental/autoqasm/program/program.py
```diff
@@ -321,22 +321,38 @@ def register_args(self, args: list[Any]) -> None:
...
+
+        Raises:
+            ValueError: If there is no parameter with the given name registered with the program.
+
+        Returns:
+            FloatVar: The associated variable.
         """
         if name not in self._free_parameters:
-            self._free_parameters[name] = oqpy.FloatVar("input", name=name)
+            raise ValueError(f"Free parameter '{name}' was not found.")
```
rmshaffer, line 354: Is it worth making a new type in the autoqasm `errors` module for this?

src/braket/experimental/autoqasm/transpiler/transpiler.py
rmshaffer: Do you know how sensitive this is to the order of the `transform` calls? I would assume `comparisons` is more closely related to `logical_expressions` and so it might logically go next to that one. The `assignments` one turned out to be very sensitive to the ordering, but probably this new one is not so much.

src/braket/experimental/autoqasm/types/types.py
rmshaffer: Are you using this functionality somewhere (other than in the tests)? Just curious why we would need it - I think internally we have been using `oqpy.FloatVar` directly when we need to specify a name, and the AutoQASM user scenario for this is to assign to the desired name, e.g. `my_var_name = aq.FloatVar(val)` and the assignments operator takes care of naming the oqpy variable correctly.

test/unit_tests/braket/experimental/autoqasm/test_api.py
rmshaffer: I believe this is happening because the assignments converter is running after your new comparisons converter. I wonder if just moving your new `comparisons.transform()` call to happen _after_ the `assignments.transform()` call in `transpiler.py` would fix this.

rmshaffer (commented):
These new comparison converters are very necessary in general - thanks for adding them! A few comments and suggestions here, will let @yitchen-tim give his thoughts as well.

src/braket/experimental/autoqasm/types/types.py
laurencap: I had at one point, I think I'm not anymore. I still think we should have this, but I could move it into a separate PR if you want.

One reason we'd want it is that we are trying to move away from calling oqpy directly anywhere we want. I also feel like it's incorrect to accept a kwarg (well, oqpy accepts it, and the object has such an attribute) and then overwrite it.

test/unit_tests/braket/experimental/autoqasm/test_api.py
laurencap: Hm I moved it down below logical expressions, but no effect on the test output

rmshaffer: Ok, interesting. I wonder if to fix this somehow your comparison converters would have to return oqpy expressions rather than variables - if you're declaring a variable I don't see how to avoid this. Anyway, it doesn't seem important for this PR (and maybe not important at all).

src/braket/experimental/autoqasm/operators/utils.py
laurencap: I forgot to remove it -- I had expected it to fail but it works fine.

src/braket/experimental/autoqasm/operators/comparisons.py
```diff
@@ -0,0 +1,126 @@
...
+    else:
+        return a < b
+
+
+def _aq_lt(a: Any, b: Any) -> aq_types.BoolVar:
+    a, b = _register_and_convert_parameters(a, b)
+
+    oqpy_program = program.get_program_conversion_context().get_oqpy_program()
+    result = aq_types.BoolVar()
+    oqpy_program.declare(result)
```
yitchen-tim, line 46: I remember we switched to autodeclare because some issue in nested if-else (https://github.com/amazon-braket/amazon-braket-sdk-python/pull/650).  I think it was because of simulator.

src/braket/experimental/autoqasm/types/types.py
rmshaffer: Yeah, makes sense. I guess ultimately these `aq.*Var` types will probably not derive from the `oqpy` types anymore, so this logic might change again at that point.

If we do support `name` here, would it make sense to add it explicitly as an (optional) parameter to `__init__`? Otherwise someone would have to look at the implementation to know that it exists. Same goes for `val`, actually - if we want to truly abstract the oqpy API away, we should have an explicit signature for `__init__` here rather than just passing all the args and kwargs to oqpy.

And yes since it's independent, moving this part to a separate PR might make sense, if that's not too much overhead for you.

src/braket/experimental/autoqasm/operators/comparisons.py
```diff
@@ -0,0 +1,126 @@
...
+    oqpy_program.set(result, a <= b)
+    return result
+
+
+def gt_(a: Any, b: Any) -> Union[bool, aq_types.BoolVar]:
+    """Functional form of ">".
+
+    Args:
+        a (Any): Callable that returns the first expression.
+        b (Any): Callable that returns the second expression.
```
rmshaffer, line 82:
```suggestion
        a (Any): The first expression.
        b (Any): The second expression.
```

src/braket/experimental/autoqasm/operators/comparisons.py
```diff
@@ -0,0 +1,126 @@
...
+    oqpy_program.set(result, a > b)
+    return result
+
+
+def gteq_(a: Any, b: Any) -> Union[bool, aq_types.BoolVar]:
+    """Functional form of ">=".
+
+    Args:
+        a (Any): Callable that returns the first expression.
+        b (Any): Callable that returns the second expression.
```
rmshaffer, line 108:
```suggestion
        a (Any): The first expression.
        b (Any): The second expression.
```

src/braket/experimental/autoqasm/operators/comparisons.py
```diff
@@ -0,0 +1,126 @@
...
+    oqpy_program.set(result, a < b)
+    return result
+
+
+def lteq_(a: Any, b: Any) -> Union[bool, aq_types.BoolVar]:
+    """Functional form of "<=".
+
+    Args:
+        a (Any): Callable that returns the first expression.
+        b (Any): Callable that returns the second expression.
```
rmshaffer, line 56:
```suggestion
        a (Any): The first expression.
        b (Any): The second expression.
```

test/unit_tests/braket/experimental/autoqasm/test_parameters.py
```diff
@@ -484,5 +486,264 @@ def test_binding_variable_fails():
...
+
+    expected = """OPENQASM 3.0;
+input float[64] basis;
+qubit[1] __qubits__;
+bool __bool_0__;
+__bool_0__ = basis == 1;
+if (__bool_0__) {
+    h __qubits__[0];
+} else {
+    bool __bool_1__;
```
yitchen-tim, line 602: Does this work with simulator? Per https://github.com/amazon-braket/amazon-braket-sdk-python/pull/650 we biased toward not having variable declare in if-else scope. @rmshaffer do you remember why we didn't want declaration in the if-else scope? Was it solely because of simulator support?

rmshaffer (approved):

src/braket/experimental/autoqasm/types/types.py
laurencap: Yes I'll move it 👍

test/unit_tests/braket/experimental/autoqasm/test_api.py
laurencap: Might be nice to revisit in Q1 -- I will make an issue

test/unit_tests/braket/experimental/autoqasm/test_parameters.py
rmshaffer: For user-defined variables in Python, they can be separately assigned in `if` and `else` blocks like:
```
if cond:
    a = aq.IntVar(1)
else:
    a = aq.IntVar(2)
return a
```
In the OpenQASM the `a` variable needs to be declared prior to the `if` block in order for this to work.

In this case since it is an auto-generated variable that only exists in the OpenQASM, it will not be used in multiple Python blocks, so I think there should be no issue declaring it here.

src/braket/experimental/autoqasm/operators/comparisons.py
laurencap: oh, my bad, sorry!

test/unit_tests/braket/experimental/autoqasm/test_parameters.py
yitchen-tim: I see, it was the declaration inside if branch being in different scope of else branch. Nested scope should work.

src/braket/experimental/autoqasm/operators/comparisons.py
laurencap: How does that work? I tried to just swap this for `autodeclare`, but I get the following error:
```
RuntimeError: OQPy expressions cannot be converted to bool. This can occur if you try to check the equality
of expressions using == instead of expr_matches.
```

test/unit_tests/braket/experimental/autoqasm/test_parameters.py
laurencap: I'll add some tests with the simulator

laurencap: Added new tests!

src/braket/experimental/autoqasm/types/types.py
laurencap: Working on a draft now: https://github.com/amazon-braket/amazon-braket-sdk-python/pull/793/files

yitchen-tim (approved):

PR #796, opened 2023-11-13
update: default no longer returning RETIRED devices from get_devices

krneta:
*Issue #, if available:*

*Description of changes:*

*Testing done:*

## Merge Checklist

#### General

#### Tests

### DIFF
--- a/src/braket/aws/aws_device.py
+++ b/src/braket/aws/aws_device.py
@@ -571,7 +571,7 @@ class AwsDevice(Device):
                 searched for the region of the current session.
             statuses (Optional[list[str]]): device status filter, default is `None`. When `None`
                 is used, RETIRED devices will not be returned. To include RETIRED devices in
-                the results, use a filter that includes `"RETIRED"` for this parameter.
+                the results, use a filter that includes "RETIRED" for this parameter.
             provider_names (Optional[list[str]]): provider name filter, default is `None`
             order_by (str): field to order result by, default is `name`.
                 Accepted values are ['arn', 'name', 'type', 'provider_name', 'status']
--- a/src/braket/aws/aws_device.py
+++ b/src/braket/aws/aws_device.py
@@ -569,7 +569,9 @@ class AwsDevice(Device):
             types (Optional[list[AwsDeviceType]]): device type list, default is `None`
                 QPUs will be searched for all regions and simulators will only be
                 searched for the region of the current session.
-            statuses (Optional[list[str]]): device status list, default is `None`
+            statuses (Optional[list[str]]): device status list, default is `None`. When `None`
+                is used, RETIRED devices will not be returned. To include RETIRED devices in
+                the results, add `RETIRED` to the list passed to this parameter.
             provider_names (Optional[list[str]]): provider name list, default is `None`
             order_by (str): field to order result by, default is `name`.
                 Accepted values are ['arn', 'name', 'type', 'provider_name', 'status']
--- a/src/braket/aws/aws_device.py
+++ b/src/braket/aws/aws_device.py
@@ -571,7 +571,7 @@ class AwsDevice(Device):
                 searched for the region of the current session.
             statuses (Optional[list[str]]): device status filter, default is `None`. When `None`
                 is used, RETIRED devices will not be returned. To include RETIRED devices in
-                the results, use a filter that includes `RETIRED` for this parameter.
+                the results, use a filter that includes `"RETIRED"` for this parameter.
             provider_names (Optional[list[str]]): provider name filter, default is `None`
             order_by (str): field to order result by, default is `name`.
                 Accepted values are ['arn', 'name', 'type', 'provider_name', 'status']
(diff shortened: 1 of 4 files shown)


src/braket/aws/aws_device.py
```diff
@@ -569,7 +569,9 @@ def get_devices(
             types (Optional[list[AwsDeviceType]]): device type list, default is `None`
                 QPUs will be searched for all regions and simulators will only be
                 searched for the region of the current session.
-            statuses (Optional[list[str]]): device status list, default is `None`
+            statuses (Optional[list[str]]): device status list, default is `None`. When `None`
+                is used, RETIRED devices will not be returned. To include RETIRED devices in
+                the results, add `RETIRED` to the list passed to this parameter.
```
ajberdy, line 574: We typically state the default semantically used, even if the function signature has `=None`. Here, I believe the default is `["ONLINE", "OFFLINE"]`. We can add a note in the doc string that the default behavior is to filter out `"RETIRED"`

src/braket/aws/aws_session.py
```diff
@@ -645,6 +647,8 @@ def search_devices(
                     continue
                 if statuses and result["deviceStatus"] not in statuses:
                     continue
+                if statuses is None and result["deviceStatus"] == "RETIRED":
+                    continue
```
ajberdy, line 651: Am I missing something, or does this not actually impact the `search_devices` call? Does this fix our time issue? It looks like it still gets every device but then filters after; is this a limitation of the search devices api?

src/braket/aws/aws_device.py
krneta: I am generally not a fan of listing this stuff out here since it can more easily go out-of-date (like if we add another state). Then, for users of older versions of the BDK, it would actually be wrong as it would return more than those two.

src/braket/aws/aws_session.py
krneta: It does fix our time issue. Even though the API returns all values, we will filter our the retired ones. This means that when we actually do get_device for all returned devices, we will be doing many fewer calls.

ajberdy: Ah that's the piece I was missing, that makes sense if `get_device` is what's taking a long time

src/braket/aws/aws_device.py
ajberdy: The docs are versioned by release, so I'm not sure I see the issue. The doc string is wrong for older versions of the bdk regardless; I think it's just a question of the phrasing here

ajberdy: Oh I see what you mean by listing stuff out. Would it be better to just add a new parameter `filter_retired=True` that kicks in when `statuses` is `None` to separate the logic?

krneta: I considered it, butI think it would be confusing to add a parameter that can have a direct opposite effect of another parameter (ie. have a filter_retired=True and statuses=["RETIRED"])

ajberdy: I see that "default is `None`" is consistent with the other doc strings. I don't love how the argument doc strings don't actually specify the default behavior, just the default value of the parameter.

I'd prefer to see these called out as filters explicitly, maybe even just changing "device name list, default is `None`" to "device name filter, default is `None`". Then we could clarify statuses like "device status filter, default is `None`. When `None` is used, RETIRED devices will not be returned. To include RETIRED devices in the results, use a filter including `"RETIRED"` for this parameter."

ajberdy (dismissed):
Approved with some doc nits, feel free to take or leave

src/braket/aws/aws_device.py
krneta: Good call

krneta:
> Approved with some doc nits, feel free to take or leave

Done.

src/braket/aws/aws_device.py
```diff
@@ -564,13 +564,15 @@ def get_devices(
...
+            arns (Optional[list[str]]): device ARN filter, default is `None`
+            names (Optional[list[str]]): device name filter, default is `None`
+            types (Optional[list[AwsDeviceType]]): device type filter, default is `None`
                 QPUs will be searched for all regions and simulators will only be
                 searched for the region of the current session.
-            statuses (Optional[list[str]]): device status list, default is `None`
-            provider_names (Optional[list[str]]): provider name list, default is `None`
+            statuses (Optional[list[str]]): device status filter, default is `None`. When `None`
+                is used, RETIRED devices will not be returned. To include RETIRED devices in
+                the results, use a filter that includes `RETIRED` for this parameter.
```
ajberdy, line 574: Smallest nit: in this context we mean the string `"RETIRED"`, unless we have an enum somewhere defining `RETIRED`

ajberdy (approved):
Looks good! Smallest optional nit

ajberdy (approved):

ajberdy (approved):

PR #756, opened 2023-10-23
doc: Add matrix expressions to docstrings

ykharkov:
Resolves issue https://github.com/amazon-braket/amazon-braket-sdk-python/issues/597

*Description of changes:*
Added definitions of unitary matrices for each gate (docstrings)

*Testing done:*

## Merge Checklist

_Put an `x` in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your pull request._

#### General

- [x] I have read the [CONTRIBUTING](https://github.com/amazon-braket/amazon-braket-sdk-python/blob/main/CONTRIBUTING.md) doc
- [x] I used the commit message format described in [CONTRIBUTING](https://github.com/amazon-braket/amazon-braket-sdk-python/blob/main/CONTRIBUTING.md#commit-your-change)
- [x] I have updated any necessary documentation, including [READMEs](https://github.com/amazon-braket/amazon-braket-sdk-python/blob/main/README.md) and [API docs](https://github.com/amazon-braket/amazon-braket-sdk-python/blob/main/CONTRIBUTING.md#documentation-guidelines) (if appropriate)

#### Tests

- [ ] I have added tests that prove my fix is effective or that my feature works (if appropriate)
- [ ] I have checked that my tests are not configured for a specific region or account (if appropriate)

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #799, opened 2023-11-14
feat: add U and GPhase gates

jcjaskula-aws:
*Issue #, if available:*
Adding the U and GPhase gates so customers can build circuits from OpenQASM3 and simulate them with the local simulator.
This allows to import "stdgates.inc".

*Description of changes:*
Created a U and GPhase class and adding them to the translator dict.

*Testing done:*
Adding a series of new tests.

## Merge Checklist

#### General

#### Tests

### DIFF
--- a/src/braket/circuits/ascii_circuit_diagram.py
+++ b/src/braket/circuits/ascii_circuit_diagram.py
@@ -132,11 +132,11 @@ class AsciiCircuitDiagram(CircuitDiagram):
         Compute the integrated phase at a certain moment.
 
         Args:
-            global_phase (float|None): The integrated phase up to the computed moment
+            global_phase (float | None): The integrated phase up to the computed moment
             items (list[Instruction]): list of instructions
 
         Returns:
-            float|None: The updated integrated phase.
+            float | None: The updated integrated phase.
         """
         moment_phase = 0
         for item in items:


src/braket/circuits/gates.py
```diff
@@ -1091,6 +1091,118 @@ def phaseshift(
...
+        """
+        Generate parameterized Unitary matrix.
+        https://openqasm.com/language/gates.html#built-in-gates
+
+        Returns:
+            np.ndarray: U Matrix
+        """
+        theta = self.angle_1
+        phi = self.angle_2
+        lam = self.angle_3
```
ajberdy, line 1131: nit: imo `_lambda` would be more readable/standard

src/braket/circuits/gates.py
```diff
@@ -1091,6 +1091,118 @@ def phaseshift(
...
+                ],
+                [
+                    np.exp(1j * phi) * np.sin(theta / 2),
+                    np.exp(1j * (phi + lam)) * np.cos(theta / 2),
+                ],
+            ]
+        )
+
+    def adjoint(self) -> list[Gate]:
+        return [U(self.angle_1, np.pi - self.angle_3, np.pi - self.angle_2)]
```
ajberdy, line 1146: Is this equivalent to the [mapping stated in the docs](https://openqasm.com/language/gates.html#:~:text=The%20base%20case%20is%20given%20by%20replacing%20inv%20%40%20U(%CE%B8%2C%20%CF%95%2C%20%CE%BB)%20by%20U(%2D%CE%B8%2C%20%2D%CE%BB%2C%20%2D%CF%95))?

src/braket/circuits/gates.py
```diff
@@ -1091,6 +1091,118 @@ def phaseshift(
...
+
+    @property
+    def _qasm_name(self) -> str:
+        return "U"
+
+    def to_matrix(self) -> np.ndarray:
+        """
+        Generate parameterized Unitary matrix.
+        https://openqasm.com/language/gates.html#built-in-gates
+
```
ajberdy, line 1125: Let's add a unitary matrix as per https://github.com/amazon-braket/amazon-braket-sdk-python/pull/756

src/braket/circuits/gates.py
```diff
@@ -1091,6 +1091,118 @@ def phaseshift(
...
+        angle_3: Union[FreeParameterExpression, float],
+        *,
+        control: Optional[QubitSetInput] = None,
+        control_state: Optional[BasisStateInput] = None,
+        power: float = 1,
+    ) -> Iterable[Instruction]:
+        """Registers this function into the circuit class.
+
+        Args:
+            target1 (QubitInput): Target qubit 1 index.
```
ajberdy, line 1170: type: `target1` -> `target` (also doc string looks wrong)

test/unit_tests/braket/circuits/test_gates.py
```diff
@@ -54,6 +54,7 @@ class TripleAngle:
     (Gate.Rx, "rx", ir.Rx, [SingleTarget, Angle], {}),
     (Gate.Ry, "ry", ir.Ry, [SingleTarget, Angle], {}),
     (Gate.Rz, "rz", ir.Rz, [SingleTarget, Angle], {}),
+    (Gate.U, "u", None, [SingleTarget, TripleAngle], {}),
```
ajberdy, line 57: should this be "U"?

src/braket/circuits/gates.py
ajberdy: I don't believe the local simulator defines an adjoint for `U`. The OpenQASM spec states that the inverse of `U` can be expressed as `U(self.angle_1, -self.angle_3, -self.angle_2)`, which seems simpler here. Can we confirm they're equivalent and opt for the simpler expression?

src/braket/circuits/gates.py
```diff
@@ -189,6 +189,80 @@ def i(
 Gate.register_gate(I)
 
 
+class GPhase(AngledGate):
+    """Z-axis rotation gate.
```
rmshaffer, line 193: update docstring

src/braket/circuits/gates.py
```diff
@@ -189,6 +189,80 @@ def i(
...
+                `control`. Will be ignored if `control` is not present. May be represented as a
+                string, list, or int. For example "0101", [0, 1, 0, 1], 5 all represent
+                controlling on qubits 0 and 2 being in the \\|0⟩ state and qubits 1 and 3 being
+                in the \\|1⟩ state. Default "1" * len(control).
+            power (float): Integer or fractional power to raise the gate to. Negative
+                powers will be split into an inverse, accompanied by the positive power.
+                Default 1.
+
+        Returns:
+            Instruction: GPhase instruction d.
```
rmshaffer, line 253:
```suggestion
            Instruction: GPhase instruction.
```

src/braket/circuits/gates.py
```diff
@@ -189,6 +189,80 @@ def i(
...
+
+        Examples:
+            >>> circ = Circuit().gphase(0.45)
+        """
+        return [
+            Instruction(GPhase(angle), control=control, control_state=control_state, power=power)
+        ]
```
rmshaffer, line 260: type hint says this returns `Instruction`, but this is inconsistent

src/braket/circuits/gates.py
ajberdy: Oh actually local simulator uses `lambda_` that would probably be my vote for consistency, but I don't feel too strongly

test/unit_tests/braket/circuits/test_gates.py
ajberdy: Ah makes sense

src/braket/circuits/gates.py
```diff
@@ -189,6 +189,80 @@ def i(
...
+
+    @property
+    def _qasm_name(self) -> str:
+        return "gphase"
+
+    def adjoint(self) -> list[Gate]:
+        return [GPhase(-self.angle)]
+
+    def to_matrix(self) -> np.ndarray:
+        return np.exp(1j * self.angle) * np.eye(1)
```
ajberdy: Can you manually verify that gphase works correctly in the following cases, running on the local simulator *and* with Circuit.to_unitary():
- gphase no target
- gphase targets (1, multiple)
- controlled gphase
- neg-controlled gphase fails (but passes if controlled first)

src/braket/circuits/gates.py
```diff
@@ -189,6 +189,104 @@ def i(
...
+
+    def adjoint(self) -> list[Gate]:
+        return [GPhase(-self.angle)]
+
+    def to_matrix(self) -> np.ndarray:
+        r"""
+        Generate a 1x1 matrix containing the global phase.
+
+        Unitary matrix:
+            .. math:: \mathtt{gphase}(\gamma) = e^(i \gamma) I_m.
```
ajberdy, line 223: `m` doesn't mean anything in this scope

src/braket/circuits/gates.py
```diff
@@ -189,6 +189,104 @@ def i(
...
+        """
+        if control is not None:
+            if control_state is None:
+                inv = 0
+            elif isinstance(control_state, int):
+                inv = 1 if control_state % 2 else 0
+                control_state = control_state - 1
+            elif isinstance(control_state, (str, list)):
+                inv = 1 - control_state[0]
+                control_state = control_state[1:]
```
ajberdy, line 275: This is really hard to follow. Can we separate out extracting the control state from the parameter and the actual logic? I'd prefer to use if/else for boolean logic then bit flags and -1^0, as it's more readable

src/braket/circuits/gates.py
```diff
@@ -189,6 +189,104 @@ def i(
...
+            return PhaseShift.phaseshift(
+                control[0],
+                angle * (-1) ** inv,
+                control=control[1:],
+                control_state=control_state,
+                power=power,
+            )
+
+        return Instruction(GPhase(angle), control=control, control_state=control_state, power=power)
```
ajberdy, line 284: I'm not sure how we can have a zero-target gate with our current BDK Circuit class, since we don't define a set number of qubits for the circuit, and we don't even know all qubits used at the time of processing this instruction. It seems like we would need to give Circuit a new "global_phase" attribute that gets modified directly and gets used to add a gphase to the OpenQASM IR or update the generated matrix with to_unitary

src/braket/circuits/ascii_circuit_diagram.py
```diff
@@ -288,7 +296,7 @@ def _ascii_diagram_column(
                         else ascii_symbols[item_qubit_index]
                     )
                 elif qubit in control_qubits:
-                    symbols[qubit] = "C"
+                    symbols[qubit] = "C" if map_control_qubit_states[qubit] else "N"
```
ajberdy, line 299: I think this is a good improvement. Would love to get a second opinion from @speller26 to make sure I'm not missing anything (ie. name conflicts or existing conventions) with this notation

src/braket/circuits/gates.py
```diff
@@ -215,6 +215,123 @@ def i(
...
+                    X.x(control_qubits[-1]),
+                ]
+
+            highest_control_qubit = max(
+                [qubit for qubit, state in zip(control_qubits, control_basis_state) if state]
+            )
+            highest_control_qubit_index = control_qubits.index(highest_control_qubit)
+
+            control_qubits.pop(highest_control_qubit_index)
+            control_basis_state.pop(highest_control_qubit_index)
```
ajberdy, line 319: I'm a bit confused by this logic. The ordering of the control modifiers is well-defined, with the rightmost qubit/modifier getting applied first. This shouldn't change depending on whether that modifier is ctrl or negctrl.

src/braket/circuits/circuit.py
```diff
@@ -1196,6 +1201,9 @@ def _to_openqasm(
             ]
         )
 
+        if self.global_phase:
+            ir_instructions.append(f"gphase({self.global_phase});")
```
ajberdy, line 1205: Are we ok with combining all the gphase instructions? What if someone has the instruction in multiple places and they expect to see them there? I understand they are equivalent, but we don't remove `x x` from a Circuit in the BDK

src/braket/circuits/ascii_circuit_diagram.py
ajberdy: Yes, it might be a good time to update to that. Might want to think if that's a breaking change, but I suspect we can assume users use the printed representation purely visually, though it may trigger some test failures for libraries built on Braket

src/braket/circuits/gates.py
ajberdy: An operation with `control=[2, 5, 3], control_state="011"` should first apply positive control to qubit 3, then positive control to qubit 5, then negative control to qubit 2. It looks like this code uses qubit value (rather than position in the control list) to sort, starting in this example with control on qubit 5. The order of ctr/negctrl applied should be 3, 5, 2, regardless of control state.

I think we want something like
```
if control_qubits:
    phase_target = control_qubits[-1]
    target_control_state = control_basis_state[-1]
    phase_instruction = PhaseShift.phaseshift(
        phase_target,
        angle,
        control=control_qubits[:-1],
        control_state=control_basis_state[:-1],
        power=power,
    ),
    cphase_instructions = (
        [phase_instruction]
        if target_control_state:
        else [X.x(phase_target), phase_instruction, X.x(phase_target)]
    )
```
Though maybe we want to add another gate for "neg" phaseshift instead of relying on the Xs

src/braket/circuits/gates.py
```diff
@@ -215,6 +215,123 @@ def i(
...
+
+        Examples:
+            >>> circ = Circuit().gphase(0.45)
+        """
+        if control is not None:
+            control_qubits = QubitSet(control)
+
+            if control_state is None:
+                control_state = len(control_qubits)
+            control_basis_state = list(BasisState(control_state, len(control_qubits)).as_tuple)
```
ajberdy, line 298:
```suggestion
            control_basis_state = BasisState(control_state, len(control_qubits))
```
This should work without calling list or as_tuple, as methods like len, iter, etc. are implemented. If we're running into a case that isn't supported (slice indexing maybe?) I think we should add that to the `BasisState` class

ajberdy: I've gone ahead and implemented these improvements in https://github.com/amazon-braket/amazon-braket-sdk-python/pull/808, feel free to pull those changes into this branch

src/braket/circuits/circuit.py
ajberdy: Yes, that makes sense. I'd love for more people to weigh in here before we decide which approach to use. Thanks for explaining the options

src/braket/circuits/circuit.py
```diff
@@ -1423,7 +1431,9 @@ def to_unitary(self) -> np.ndarray:
         qubits = self.qubits
         if not qubits:
             return np.zeros(0, dtype=complex)
-        return calculate_unitary_big_endian(self.instructions, qubits)
+        return calculate_unitary_big_endian(self.instructions, qubits) * np.exp(
+            1j * self.global_phase
+        )
```
ajberdy, line 1436: good catch

src/braket/circuits/gates.py
```diff
@@ -215,6 +215,121 @@ def i(
...
+                        control=control_qubits[:-1],
+                        control_state=control_basis_state[:-1],
+                        power=power,
+                    ),
+                    X.x(control_qubits[-1]),
+                ]
+
+            rightmost_control_qubit_index = (
+                len(control_qubits) - control_basis_state[::-1].index(1) - 1
+            )
```
ajberdy, line 315: I still don't agree with this interpretation. I don't think we should be filtering through the control state to find a positive control. The qubit used here should simply be the rightmost.

src/braket/circuits/gates.py
ajberdy: I don't believe the added code complexity is worth it here. It may also confuse customers to see their qubits misordered, even if the operations are commutative. If the x-phase-x is dispreferred, then I believe we should use a unitary or add a gate.

Maybe @speller26 can weigh in here too

src/braket/circuits/moments.py
```diff
@@ -181,6 +182,8 @@ def _add(self, instruction: Instruction, noise_index: int = 0) -> None:
             self._time_all_qubits = time
         elif isinstance(operator, Noise):
             self.add_noise(instruction)
+        elif instruction.target == QubitSet([]):
+            self._global_phase += operator.angle
```
rmshaffer, line 186: if you're assuming this can only happen for a global phase instruction, maybe first `assert isinstance(operator, GPhase)`?

src/braket/circuits/gates.py
```diff
@@ -215,6 +215,121 @@ def i(
...
+                powers will be split into an inverse, accompanied by the positive power.
+                Default 1.
+
+        Returns:
+            Instruction | Iterable[Instruction]: GPhase instruction.
+
+        Examples:
+            >>> circ = Circuit().gphase(0.45)
+        """
+        if control is not None:
```
rmshaffer, line 293: ok, so just catching myself up, reading [the OQ3 spec](https://openqasm.com/language/gates.html#control-modifiers): 
"the controlled global phase gate `ctrl @ gphase(a)` is equivalent to the single-qubit gate `U(0, 0, a)`."

What does `negctrl @ gphase(a)` produce? is it the same as `ctrl @ gphase(-a)`?

Could we implement it directly using the `U` gate, so that we are more faithful to the spec? (although I agree that this could be done equivalently with the `phaseshift` or `rz` gate)

src/braket/circuits/gates.py
```diff
@@ -215,6 +215,121 @@ def i(
...
+            control_qubits = QubitSet(control)
+
+            if control_state is None:
+                control_state = len(control_qubits)
+            control_basis_state = BasisState(control_state, len(control_qubits))
+
+            if not any(control_basis_state):
```
rmshaffer, line 300: This line is confusing me. Why do we do something different if the `control_state` is all zeros?

src/braket/circuits/gates.py
rmshaffer: I also don't understand why this is necessary - I also thought we should always be acting on the rightmost qubit as the target. Unless I'm misunderstanding what `negctrl @ gphase(a)` means.

src/braket/circuits/gates.py
rmshaffer: Wouldn't the second just become `phaseshift(-a) q; gphase(a);`? On hardware, `phaseshift` and `gphase` are likely to be virtual operations, right? So inserting `x` gates actually may be worse.

src/braket/circuits/moments.py
ajberdy: You may be able to avoid circular imports by further specifying the import (ie. `braket.circuits.gates` vs `braket.circuits`)

src/braket/circuits/ascii_circuit_diagram.py
```diff
@@ -41,26 +43,23 @@ def build_diagram(circuit: cir.Circuit) -> str:
             str: ASCII string circuit diagram.
         """
 
-        if not circuit.instructions:
+        if not circuit.instructions or not any(
+            m for m in circuit._moments if m.moment_type != MomentType.GLOBAL_PHASE
+        ):
```
ajberdy, line 48:
```suggestion
        if not circuit.instructions or all(
            m.moment_type == MomentType.GLOBAL_PHASE for m in circuit._moments
        ):
```

src/braket/circuits/ascii_circuit_diagram.py
ajberdy: Yes, that's so much nicer. I say we do it. We'll need to manually update gates with built-in control qubits as well, like cnot. I will poll the devx team to see if there's alignment

src/braket/circuits/gate.py
```diff
@@ -198,7 +198,7 @@ def _to_openqasm(
 
         return (
             f"{inv_prefix}{power_prefix}{control_prefix}"
-            f"{self._qasm_name}{param_string} {', '.join(qubits)};"
+            f"{self._qasm_name}{param_string}{','.join([f' {qubit}' for qubit in qubits])};"
```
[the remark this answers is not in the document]
ajberdy: We should be able to just do `qubits` here instead of `map(str, qubits)`, since `qubits` already has the serialization properties-based formatting applied to the int values

src/braket/circuits/ascii_circuit_diagram.py
```diff
@@ -216,21 +261,25 @@ def _ascii_diagram_column_set(
 
         first_line = "{:^{width}}|\n".format(col_title, width=len(lines[0]) - 1)
 
-        return first_line + "\n".join(lines)
+        return first_line + "\n".join(lines), global_phase
```
ajberdy, line 264: Not a fan of this design pattern (function returning a tuple of two "unrelated" things). Can we make a separate helper function for this? I wouldn't be concerned about iterating through the moments twice

src/braket/circuits/ascii_circuit_diagram.py
```diff
@@ -138,7 +166,11 @@ def _ascii_group_items(
                 qubits_added = group[0]
                 instr_group = group[1]
                 # Take into account overlapping multi-qubit gates
-                if not qubits_added.intersection(set(qubit_range)):
+                if not qubits_added.intersection(set(qubit_range)) or (
+                    isinstance(item, Instruction)
+                    and isinstance(item.operator, Gate)
+                    and item.operator.name == "GPhase"
+                ):
```
ajberdy, line 173: Can you add a comment explaining this logic please?

src/braket/circuits/ascii_circuit_diagram.py
```diff
@@ -288,20 +352,51 @@ def _ascii_diagram_column(
...
+        global_phase: float | None,
+    ) -> str:
+        symbols_width = max([len(symbol) for symbol in symbols.values()])
         output = ""
-        for qubit in circuit_qubits:
+
+        if global_phase is not None:
+            global_phase = (
+                round(global_phase, 2) if isinstance(global_phase, float) else global_phase
+            )
```
ajberdy, line 379: Why do we need to do this? Can we not just print with limited precision? `f"{global_phase:.2f}"`

src/braket/circuits/ascii_circuit_diagram.py
```diff
@@ -288,20 +352,51 @@ def _ascii_diagram_column(
...
             output += "{0:{width}}\n".format(margins[qubit], width=symbols_width + 1)
             output += "{0:{fill}{align}{width}}\n".format(
                 symbols[qubit], fill="-", align="<", width=symbols_width + 1
             )
         return output
+
+    @staticmethod
+    def _build_map_control_qubits(item: Instruction, control_qubits: QubitSet) -> dict(Qubit, int):
+        control_state = getattr(item, "control_state", None)
+        if control_state is not None:
```
ajberdy, line 395: Haven't gotten to the tests yet, but this is correct because control_state could be `0`. I want to make sure we have a test asserting this case, so a future developer doesn't "simplify" this condition to `if control_state`

src/braket/circuits/gates.py
```diff
@@ -215,6 +215,119 @@ def i(
...
+        angle (Union[FreeParameterExpression, float]): angle in radians.
+    """
+
+    def __init__(self, angle: Union[FreeParameterExpression, float]):
+        # Avoid parent constructor because _qubit_count must be zero
+        self._qubit_count = self.fixed_qubit_count()
+        self._ascii_symbols = []
+
+        if angle is None:
+            raise ValueError("angle must not be None")
```
ajberdy, line 235: I don't understand why we have this validation. Angle isn't optional and there's no expectation this method would work without providing angle.

src/braket/circuits/gates.py
```diff
@@ -215,6 +215,119 @@ def i(
...
+            >>> circ = Circuit().gphase(0.45)
+        """
+        if control is not None:
+            control_qubits = QubitSet(control)
+
+            control_basis_state = (
+                BasisState(control_state, len(control_qubits))
+                if control_state is not None
+                else BasisState((1,) * len(control_qubits), len(control_qubits))
+            )
```
ajberdy, line 305: More explicit if we set the control_state default first.
```suggestion
            control_state = (
                control_state
                if control_state is not None
                else "1" * len(control_qubits)
            )
            control_basis_state = BasisState(control_state, len(control_qubits))
```

src/braket/circuits/gates.py
```diff
@@ -1287,6 +1400,129 @@ def phaseshift(
...
+        angle_1 (Union[FreeParameterExpression, float]): theta angle in radians.
+        angle_2 (Union[FreeParameterExpression, float]): phi angle in radians.
+        angle_3 (Union[FreeParameterExpression, float]): lambda angle in radians.
+    """
+
+    def __init__(
+        self,
+        angle_1: Union[FreeParameterExpression, float],
+        angle_2: Union[FreeParameterExpression, float],
+        angle_3: Union[FreeParameterExpression, float],
```
ajberdy, line 1423: Can we add defaults for this gate as we did [with MS](https://github.com/amazon-braket/amazon-braket-sdk-python/blob/696974da3b4b9bb80b09c3a60549193b3935c6f6/src/braket/circuits/gates.py#L3409)? That way we can users can define a [u2 gate](https://docs.quantum.ibm.com/api/qiskit/qiskit.circuit.library.U2Gate) by only providing relevant params.

ajberdy: Might be tricky since only the first gate would have a default, so maybe we circle back to this in another pr

ajberdy: Maybe we just add documentation that setting angle_1 to pi/2 is equivalent to U2

test/unit_tests/braket/circuits/test_ascii_circuit_diagram.py
```diff
@@ -29,6 +29,10 @@ def test_empty_circuit():
     assert AsciiCircuitDiagram.build_diagram(Circuit()) == ""
 
 
+def test_only_gphase_circuit():
+    assert AsciiCircuitDiagram.build_diagram(Circuit().gphase(0.1)) == ""
```
ajberdy, line 33: Hm are we sure about this? I think this could make sense, but I'd appreciate hearing a bit of context of why we're convinced

test/unit_tests/braket/circuits/test_ascii_circuit_diagram.py
```diff
@@ -58,6 +62,49 @@ def test_one_gate_one_qubit_rotation_with_parameter():
     _assert_correct_diagram(circ, expected)
 
 
+def test_one_gate_with_global_phase():
+    circ = Circuit().x(target=0).gphase(0.15)
+    expected = (
+        "T  : |0| 1  |",
+        "GP : |0|0.15|",
```
ajberdy, line 69: Does this show global phase or delta to global phase?

ajberdy: Nvm I see from other tests that it shows global phase.

src/braket/circuits/gates.py
ajberdy: Oh I see. Still seems weird to me, but I'm ok with it for consistency

src/braket/circuits/ascii_circuit_diagram.py
ajberdy: Ah right, thanks

src/braket/circuits/ascii_circuit_diagram.py
```diff
@@ -97,6 +105,49 @@ def build_diagram(circuit: cir.Circuit) -> str:
...
+    ) -> float | None:
+        """
+        Compute the integrated phase at a certain moment.
+
+        Args:
+            global_phase (float|None): The integrated phase up to the computed moment
+            items (list[Instruction]): list of instructions
+
+        Returns:
+            float|None: The updated integrated phase.
```
speller26, line 139: nit
```suggestion
            float | None: The updated integrated phase.
```

src/braket/circuits/ascii_circuit_diagram.py
```diff
@@ -97,6 +105,49 @@ def build_diagram(circuit: cir.Circuit) -> str:
...
+
+    @staticmethod
+    def _compute_moment_global_phase(
+        global_phase: float | None, items: list[Instruction]
+    ) -> float | None:
+        """
+        Compute the integrated phase at a certain moment.
+
+        Args:
+            global_phase (float|None): The integrated phase up to the computed moment
```
speller26, line 135:
```suggestion
            global_phase (float | None): The integrated phase up to the computed moment
```

src/braket/circuits/gate.py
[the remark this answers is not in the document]
speller26: Both of the suggestions will leave no space between `param_string` and the qubits, no? Being explicit is nice, but I think JC's implementation is elegant enough.

speller26 (dismissed):
Looks good, just some nits

speller26 (approved):

PR #803, opened 2023-11-16
feat: remove *args from api decorators

ajberdy:
*Issue #, if available:*

*Description of changes:*

Remove `*args` from the signatures of the API decorators. Replace with single positional argument for decorated function. Enforce keyword-only for other decorator args, as before.

*Testing done:*

Unit tests passing

## Merge Checklist

#### General

#### Tests

### DIFF
--- a/src/braket/experimental/autoqasm/api.py
+++ b/src/braket/experimental/autoqasm/api.py
@@ -13,6 +13,8 @@
 
 """This module implements the decorator API for generating programs using AutoQASM."""
 
+from __future__ import annotations
+
 import copy
 import functools
 import inspect
@@ -43,10 +45,11 @@ from braket.experimental.autoqasm.program.gate_calibrations import GateCalibrati
 
 
 def main(
-    *args,
+    func: Optional[Callable] = None,
+    *,
     num_qubits: Optional[int] = None,
     device: Optional[Union[Device, str]] = None,
-) -> Callable[[Any], aq_program.Program]:
+) -> Callable[..., aq_program.Program]:
     """Decorator that converts a function into a callable that returns
     a Program object containing the quantum program.
 
@@ -54,20 +57,22 @@ def main(
     function is called, and a new Program object is returned each time.
 
     Args:
+        func (Optional[Callable]): Decorated function. May be `None` in the case where decorator
+            is used with parentheses.
         num_qubits (Optional[int]): Configuration to set the total number of qubits to declare in
             the program.
         device (Optional[Union[Device, str]]): Configuration to set the target device for the
             program. Can be either an Device object or a valid Amazon Braket device ARN.
 
     Returns:
-        Callable[[Any], Program]: A callable which returns the converted
+        Callable[..., Program]: A callable which returns the converted
         quantum program when called.
     """
     if isinstance(device, str):
         device = AwsDevice(device)
 
     return _function_wrapper(
-        *args,
+        func,
         converter_callback=_convert_main,
         converter_args={
             "user_config": aq_program.UserConfig(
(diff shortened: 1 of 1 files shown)


src/braket/experimental/autoqasm/api.py
```diff
@@ -114,40 +127,50 @@ def gate_calibration(*args, implements: Callable, **kwargs) -> Callable[[], Gate
...
 
 
 def _function_wrapper(
-    *args: tuple[Any],
+    func: Optional[Callable],
+    *,
     converter_callback: Callable,
     converter_args: Optional[dict[str, Any]] = None,
-) -> Callable[[Any], aq_program.Program]:
+) -> Callable[..., Optional[Union[aq_program.Program, GateCalibration]]]:
```
rmshaffer, line 141: This return type hint (as for all of the functions above which call it) is still not correct in the case where we `return functools.partial(...)`. Can we make recursive type hints which `_function_wrapper` and all of the decorators can use for their return type hints? Haven't tried it, but I guess something like (maybe this can be simplified):
```
ProgramDecoratorRetVal = Callable[...,   # for main(), subroutine()
    Union['ProgramDecoratorRetVal', aq_program.Program]]
GateDecoratorRetVal = Callable[...,   # for gate()
    Optional['GateDecoratorRetVal']]
GateCalibrationDecoratorRetVal = Callable[...,    # for gate_calibration()
    Union['GateCalibrationDecoratorRetVal', GateCalibration]]
DecoratorRetVal = Union[    # for _function_wrapper()
    ProgramDecoratorRetVal, GateDecoratorRetVal, GateCalibrationDecoratorRetVal]
```

rmshaffer: aside: wish we could write `A | B | None` instead of `Optional[Union[A, B]]`!

src/braket/experimental/autoqasm/api.py
```diff
@@ -114,40 +127,50 @@ def gate_calibration(*args, implements: Callable, **kwargs) -> Callable[[], Gate
...
-        quantum program when called.
+        Callable[..., Optional[Union[Program, GateCalibration]]]: A callable which
+        returns the converted construct, if any, when called.
     """
-    if not args:
-        # This the case where a decorator is called with only keyword args, for example:
-        #     @aq.main(num_qubits=4)
+    if not (func and callable(func)):
+        # This the case where a decorator is called either without a positional argument,
+        # or with a non-callable positional argument, which is as close of an approximation
```
rmshaffer, line 156: What's the point of allowing the non-callable positional argument case? It looks like we just throw away `func` in that case. Maybe we could just first validate (and enforce by raising an exception) that if `func` is not None, it must be callable? (That's how `func` is annotated anyway, so it seems ok to enforce it.) Then this condition could become just `if not func`.

src/braket/experimental/autoqasm/api.py
```diff
@@ -114,40 +127,50 @@ def gate_calibration(*args, implements: Callable, **kwargs) -> Callable[[], Gate
...
+        # we can get to the case where a decorator is called with parentheses.
+        #
+        # There is still a false negative case, where we have something like:
+        #     @aq.main(callable_pos_arg)
         #     def my_function():
+        #
+        # but this is known limitation in python (consider the valid non-decorator usage
+        # `aq.main(my_function)` for an example of why this ambiguity exists).
```
rmshaffer, line 164: brainstorming here (not something we need to do in this PR) - could we ever detect & warn the user if they somehow make this mistake? e.g., the decorated object that is returned could raise some error or warning if it is called again with a function as a first positional argument.

src/braket/experimental/autoqasm/api.py
ajberdy: Yes, I was thinking to update the type hints in another PR. We can use `|` syntax if we import `__annotations__` from future

src/braket/experimental/autoqasm/api.py
ajberdy: Right now, none of the decorators accept any positional args (in fact we enforce all to be keyword-only), so you are correct. If there is a positional arg, it is the function. We may want to keep this system because of that exact benefit.

However, in general, a decorator may take positional args and then it becomes impossible to differentiate without using heuristics.

src/braket/experimental/autoqasm/api.py
ajberdy: I'm not sure exactly which mistake you're referring to. The decorated function returned will (shortly) be an AutoQASM Program, so error handling for calls will be established in that class

PR #804, opened 2023-11-17
feat: build autoqasm program directly from main decorator

### LINKED ISSUE #805: Constraint @aq.main functions to OpenQASM inputs only. Unbound values become FreeParameters

### LINKED ISSUE #806: Remove secondary @aq.main call. A function decorated with @main is a Program

### LINKED ISSUE #813: AutoQASM: support for typed input parameters

ajberdy:
*Issue #, if available:*

Fixes #805, fixes #806, fixes #813

*Description of changes:*

- Update `aq.main` to build AutoQASM program directly.
- Add support for typed input parameters and auto-capture of inputs.
  - usable in qubit indices
- Various typing simplifications

Wrapped testing programs in fixtures so they don't get built during other tests.

*Testing done:*

Unit tests updated and passing, notebook tests passing as well.

## Merge Checklist

#### General

#### Tests


ajberdy:
Two tests failing in test_api: test_nested_function and test_main_from_main

laurencap (commented):
Reviewed `test_api.py`

test/unit_tests/braket/experimental/autoqasm/test_api.py
```diff
@@ -190,22 +202,26 @@ def bell_state_arbitrary_qubits(int[32] q0, int[32] q1) {
...
 
 
-@aq.main
-def bell_measurement_undeclared() -> None:
-    """A function that generates and measures a two-qubit Bell state."""
-    h(0)
-    cnot(0, 1)
-    c = measure([0, 1])  # noqa: F841
+@pytest.fixture
+def bell_measurement_undeclared():
```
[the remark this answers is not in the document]
ajberdy: Not *strictly* required here, but the translation happens at decorator execution time, which here was happening when the function is defined+decorated in global scope. That means the translation happens whenever this file is loaded, even if you're just running one other test. This always leads to a slowdown, but is explicitly problematic with failing example programs (which would now fail at definition time-- in the global scope at load time).

test/unit_tests/braket/experimental/autoqasm/test_api.py
```diff
@@ -163,10 +170,11 @@ def recursive_h(int[32] q) {
         recursive_h(q - 1);
     }
 }
+int[32] n = 4;
 qubit[6] __qubits__;
-recursive_h(4);"""
+recursive_h(n);"""
 
-    assert main(4).to_ir() == expected
+    assert main.make_bound_program(param_values={"n": 4}).to_ir() == expected
```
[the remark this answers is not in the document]
ajberdy: Ah I see. In my head, I interpreted these tests as testing the user story "customer wants to define this program and then build the program with a specific value for its input". I agree `make_bound_program` is a bit clunky today for this, but we can add easier methods for `Program -> Program` maps that bind variables, either by simply replacing the input declaration with an assignment, or fully inlining as an optimization pass.

If instead these tests were meant to test the codepath "non-input python variables are used during the transpilation step", then I can see how your suggestion is more relevant. It's not obvious to me which is the case, maybe we can chat so I can better understand here

test/unit_tests/braket/experimental/autoqasm/test_api.py
```diff
@@ -927,17 +978,16 @@ def bell() -> None:
...
 
-    @aq.main
-    def main():
-        bell(0, 1)
-
     with pytest.raises(errors.AutoQasmTypeError):
-        main()
+
+        @aq.main
+        def main():
```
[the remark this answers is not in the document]
ajberdy: I suspect we may want to update the other api decorators to have a similar breakdown (in another PR lol). Something like:

|                | Definition Time                          | Call Time                               |
|----------------|------------------------------------------|----------------------------------------|
| **Main**       | Validates and builds Program  | Runs Program on Device?   |
| **Subroutine** | Validates and builds subroutine definition | Adds subroutine call to Program |

test/unit_tests/braket/experimental/autoqasm/test_api.py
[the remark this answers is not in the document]
laurencap: From sync review: We will update with the second interpretation

src/braket/experimental/autoqasm/program/program.py
```diff
@@ -341,15 +341,26 @@ def _free_symbol_names(expr: FreeParameterExpression) -> Iterable[str]:
         """
         return sorted([str(s) for s in expr._expression.free_symbols if isinstance(s, Symbol)])
 
-    def register_parameter(self, parameter_name: str) -> None:
+    def register_parameter(
+        self, parameter_name: str, parameter_type: Optional[Union[float, int]] = float
```
rmshaffer, line 345:
```suggestion
        self, parameter_name: str, parameter_type: Optional[Union[bool, float, int]] = float
```

src/braket/experimental/autoqasm/api.py
```diff
@@ -65,13 +66,22 @@ def main(
             program. Can be either an Device object or a valid Amazon Braket device ARN.
 
     Returns:
-        Callable[..., Program]: A callable which returns the converted
-        quantum program when called.
+        Program | partial: A callable
+        which returns the converted quantum program when called.
```
rmshaffer, line 70: Update docstring, no longer returns a callable (in the typical case)

test/unit_tests/braket/experimental/autoqasm/test_api.py
```diff
@@ -22,36 +22,37 @@
...
 def test_sim_bell_state(bell_state_program) -> None:
-    _test_on_local_sim(bell_state_program())
+    _test_on_local_sim(bell_state_program)
 
 
 def test_multiple_calls(empty_subroutine, bell_state_subroutine) -> None:
     """Tests multiple calls to a single aq.main to ensure that each resulting
     Program object has the expected contents.
     """
+    # todo: delete or update this test as it is no longer relevant as written
```
rmshaffer, line 55: reminder to address todo

test/unit_tests/braket/experimental/autoqasm/test_api.py
```diff
@@ -119,12 +120,17 @@ def recursive_h(q: int):
...
+    @aq.main(num_qubits=6)
+    def recursive_h_wrapper(q: int):
+        recursive_h(q)
+
+    return recursive_h_wrapper
 
 
-def test_recursive_h_wrapper():
+def test_recursive_h_wrapper(recursive_h_wrapper):
+    # todo: make sure annotation is captured as int
```
rmshaffer, line 133: reminder to address todo

test/unit_tests/braket/experimental/autoqasm/test_api.py
```diff
@@ -395,48 +429,60 @@ def qasm_simple_condition(do_cnot: bool) -> bool:
...
     return do_cnot;
-}
+}}
+bool do_cnot = {"true" if do_cnot else "false"};
 qubit[2] __qubits__;
 bool __bool_0__;
 """
-    expected += f"__bool_0__ = qasm_simple_condition({'true' if do_cnot else 'false'});"
-    assert qasm_simple_condition_wrapper(do_cnot).to_ir() == expected
+    expected += "__bool_0__ = qasm_simple_condition(do_cnot);"
```
rmshaffer, line 455: you could now include this in the string literal above

src/braket/experimental/autoqasm/instructions/gates.py
```diff
@@ -16,11 +16,15 @@
...
 from typing import Union
 
+import oqpy
+
 from braket.circuits.free_parameter_expression import FreeParameterExpression
 
 from .instructions import _qubit_instruction
 from .qubits import QubitIdentifierType
 
+GateParameterType = Union[float, FreeParameterExpression, oqpy._ClassicalVar]
```
rmshaffer, line 26: thanks for adding this!

test/unit_tests/braket/experimental/autoqasm/test_api.py
ajberdy: True. However, I'll be updating these tests to use a closure around `do_cnot` instead of defining it as an input, as per [Lauren's comment](https://github.com/amazon-braket/amazon-braket-sdk-python/pull/804/files/073a11bec26ffa65632fdb57f9031b32f0f1cf14#r1398044784), so the expected IR will go back to as before

test/unit_tests/braket/experimental/autoqasm/test_parameters.py
```diff
@@ -412,11 +463,11 @@ def cal_1(angle: float):
...
     @aq.main
-    def my_program():
-        rx("$1", FreeParameter("theta"))
+    def my_program(theta):
+        rx("$1", theta)
 
-    qasm1 = my_program().with_calibrations(cal_1).make_bound_program({"theta": 0.6}).to_ir()
-    qasm2 = my_program().make_bound_program({"theta": 0.6}).with_calibrations(cal_1).to_ir()
+    qasm1 = my_program.with_calibrations(cal_1).make_bound_program({"theta": 0.6}).to_ir()
+    qasm2 = my_program.make_bound_program({"theta": 0.6}).with_calibrations(cal_1).to_ir()
```
laurencap, line 470: This `make_bound_progarm` API doesn't make sense anymore. Let's update with `my_program(theta=0.6)`, like the `Circuit` class has

laurencap: This also is halfway towards fixing one of the failing tests -- it gets rid of the TypeError, but we still have to find a new way to raise an error when `main` calls `main`

src/braket/experimental/autoqasm/instructions/gates.py
laurencap: Is it possible to use AutoQASM `types` instead? We are trying to work towards minimizing `oqpy` imports

src/braket/experimental/autoqasm/api.py
```diff
@@ -200,11 +212,12 @@ def _autograph_optional_features() -> tuple[converter.Feature]:
 
 def _convert_main(
     f: Callable,
+    *,
```
laurencap, line 215: This line can be removed (for now, until we revisit in the follow up about the more strict APIs here)

src/braket/experimental/autoqasm/api.py
```diff
@@ -218,16 +231,31 @@ def _convert_main(
...
         user_config (UserConfig): User-specified settings that influence program building.
+
+    Returns:
+        aq_program.Program: Generated AutoQASM Program.
     """
     if aq_program.in_active_program_conversion_context():
         raise errors.AutoQasmTypeError(
             f"Cannot call main function '{f.__name__}' from another main function. Did you mean "
             "to use '@aq.subroutine'?"
         )
```
laurencap, line 242: Presumably, these lines aren't reached anymore. We'll see when code-cov runs

ajberdy:
main from main issue root caused and captured in #809 to be fixed in a follow-up PR

test/unit_tests/braket/experimental/autoqasm/test_api.py
```diff
@@ -832,7 +884,7 @@ def prog() -> None:
 __bit_0__ = measure __qubits__[0];
 a = __bit_0__;
 b = a;"""
-    assert prog().to_ir() == expected
+    assert prog.to_ir() == expected
 
 
 def test_nested_function():
```
rmshaffer, line 890: `test_nested_function` fails now because the inner function `ghz` is a pure Python function but is now being called with a `FreeParameter`, which causes infinite recursion. (When called with an integer `n >= 1`, it terminates at some point.) Not sure we can really avoid this.

I guess a simple way to fix this would be to use a Python variable instead of a FreeParameter, for example:

```
def test_nested_function():
    ghz_size = 5

    @aq.main
    def make_ghz() -> None:
        def ghz(n: int):
            if n == 1:
                h(0)
            else:
                ghz(n - 1)
                cnot(0, n - 1)

        ghz(ghz_size)

    expected = """OPENQASM 3.0;
qubit[5] __qubits__;
h __qubits__[0];
cnot __qubits__[0], __qubits__[1];
cnot __qubits__[0], __qubits__[2];
cnot __qubits__[0], __qubits__[3];
cnot __qubits__[0], __qubits__[4];"""

    assert make_ghz.to_ir() == expected
```

I tested this and it seems to be fine.

laurencap (commented):
Another partial review

test/unit_tests/braket/experimental/autoqasm/test_api.py
ajberdy: This is awesome! Definitely would love to see us handle free parameters as function inputs for recursive functions. I wonder if there's some check we can do while converting a function to turn off recursive mode if parameters are not initialized (or if that makes sense to do). At the very least we could catch this and raise a more explicit error

src/braket/experimental/autoqasm/instructions/qubits.py
```diff
@@ -128,6 +129,17 @@ def _(qid: str) -> oqpy.Qubit:
         raise ValueError(f"invalid qubit label: '{qid}'")
 
 
+@_qubit.register
+def _(qid: FreeParameterExpression) -> oqpy.Qubit:
```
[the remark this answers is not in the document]
ajberdy: There are a few tests that need this functionality once the program inputs got captured as free parameters. For example [test_multi_angle_gates](https://github.com/amazon-braket/amazon-braket-sdk-python/blob/0d27fb0d8d41a7cc6e58004146fb8c4e6f79d92a/test/unit_tests/braket/experimental/autoqasm/test_parameters.py#L209).

I've also added explicit tests in this PR at the [bottom of test_api](https://github.com/amazon-braket/amazon-braket-sdk-python/blob/0d27fb0d8d41a7cc6e58004146fb8c4e6f79d92a/test/unit_tests/braket/experimental/autoqasm/test_api.py#L1045)

test/unit_tests/braket/experimental/autoqasm/test_operators.py
```diff
@@ -794,29 +791,44 @@ def test_assignment_py(value: Any) -> None:
 @pytest.mark.parametrize("value", [0, 1])
 def test_py_if_stmt(value: int) -> None:
     """Test Python if branches on true and false conditions."""
+    # todo: if statements are no longer optimized out during binding parameters
+    # we will need to be explicit about this compilation pass functionality
+    # we can still add it to `make_bound_program`
```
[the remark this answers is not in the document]
ajberdy: Before, users would define a "Program builder" with parameters that get bound during the function call:
```
@aq.main
def my_program_builder(builder_input: int):
    ...

my_program = my_program_builder(builder_input=3)
```
We still expose that functionality by allowing customers to create builder methods, but customers who simply want to make a program with arguments will likely transition to the following:
```
@aq.main
def my_program(program_input: int):
    ...

my_program = my_program(program_input=3)
```
In this case, "binding" the arguments happens in a different place and has a slightly different effect than before (creates `input` variables vs inlines the variables). My comment is just saying we should be aware of this and intentional about providing an option to "inline" a variable of a Program if we want to expose that functionality to customers outside of meta-programming.

"optimizing out if statements" is one example of a simplification that may occur while inlining a variable

src/braket/experimental/autoqasm/program/program.py
```diff
@@ -341,15 +341,26 @@ def _free_symbol_names(expr: FreeParameterExpression) -> Iterable[str]:
...
+    ) -> None:
         """Register an input parameter if it has not already been registered.
-        Only floats are currently supported.
```
laurencap, line 346: Nice

src/braket/experimental/autoqasm/program/program.py
laurencap: +1, and in the docstring too

src/braket/experimental/autoqasm/program/program.py
```diff
@@ -341,15 +341,26 @@ def _free_symbol_names(expr: FreeParameterExpression) -> Iterable[str]:
...
 
         Args:
             parameter_name (str): The name of the parameter to register with the program.
+            parameter_type (Optional[Union[float, int]]): The type of the parameter to register
+                with the program.
         """
         if parameter_name not in self._free_parameters:
-            self._free_parameters[parameter_name] = oqpy.FloatVar("input", name=parameter_name)
+            if parameter_type in [float, None]:
```
laurencap, line 355: This isn't being called by users, do we need to support `parameter_type=None`? The default is `float` not `None`

test/unit_tests/braket/experimental/autoqasm/test_parameters.py
```diff
@@ -173,32 +214,45 @@ def parametric(qubit_0: int, phi: float, theta: float):
         ms(0, qubit_0, phi, phi, theta)
 
     expected = """OPENQASM 3.0;
+int[32] qubit_0 = 2;
 input float[64] phi;
+float[64] theta = 0.5;
 qubit[5] __qubits__;
-ms(phi, phi, 0.5) __qubits__[0], __qubits__[2];"""
-    assert parametric(2, FreeParameter("phi"), 0.5).to_ir() == expected
+ms(phi, phi, theta) __qubits__[0], __qubits__[qubit_0];"""
```
laurencap, line 221: Cool!

test/unit_tests/braket/experimental/autoqasm/test_parameters.py
```diff
@@ -173,32 +214,45 @@ def parametric(qubit_0: int, phi: float, theta: float):
...
+    assert (
+        parametric.make_bound_program(
```
laurencap, line 223: Really starting to dislike this interface! 😩

test/unit_tests/braket/experimental/autoqasm/test_parameters.py
```diff
@@ -173,32 +214,45 @@ def parametric(qubit_0: int, phi: float, theta: float):
...
+    )
 
 
 def test_parameters_passed_as_main_arg():
     """Test that parameters work when passed as input values."""
 
     @aq.main
-    def parametric(phi: float):
-        cphaseshift(0, 1, phi)
+    def parametric(my_phi: float):
```
laurencap, line 248: Nice that we don't have this ambiguity anymore

test/unit_tests/braket/experimental/autoqasm/test_parameters.py
```diff
@@ -435,58 +486,61 @@ def test_bind_empty_program():
...
+        def parametric_explicit():
+            alpha = aq.FloatVar(1.2)  # noqa: F841
+            rx(0, FreeParameter("alpha"))
 
     with pytest.raises(RuntimeError, match="conflicting variables with name alpha"):
-        parametric()
+
+        @aq.main
+        def parametric(alpha):
+            alpha = aq.FloatVar(1.2)  # noqa: F841
```
laurencap, line 542: Nice solution for covering existing tests and also focusing on the expected user interface. I appreciate it!

test/unit_tests/braket/experimental/autoqasm/test_parameters.py
```diff
@@ -629,22 +683,22 @@ def main(basis: int):
...
+    measurements = _test_parametric_on_local_sim(main, {"basis": 1})
     assert 1 in measurements["c"] and 0 in measurements["c"]
 
 
 def test_sim_comparison_stmts():
     @aq.main
-    def main(basis: int):
+    def main(basis: float):
         if basis > 0.5:
             x(0)
```
laurencap, line 696: Basis seems like a weird name for a float, my bad for writing it like that. Mind updating this for me? No worries if you want to skip, it's definitely out of scope 😝 
```suggestion
    def main(threshold: float):
        if threshold > 0.5:
            x(0)
```

test/unit_tests/braket/experimental/autoqasm/test_parameters.py
```diff
@@ -715,7 +769,7 @@ def parametric(alpha: float, beta: float):
 }
 bit __bit_1__;
 __bit_1__ = measure __qubits__[0];"""
-    assert parametric(FreeParameter("alpha"), FreeParameter("beta")).to_ir() == expected
```
laurencap, line 718: Again so nice not to have to specify these

test/unit_tests/braket/experimental/autoqasm/test_parameters.py
```diff
@@ -729,15 +783,16 @@ def parametric(alpha: float, beta: float):
...
 bool __bool_0__;
-__bool_0__ = alpha && 1.5;
+__bool_0__ = alpha && beta;
 if (__bool_0__) {
     rx(alpha) __qubits__[0];
 }
 bit __bit_1__;
 __bit_1__ = measure __qubits__[0];"""
-    assert parametric(FreeParameter("alpha"), 1.5).to_ir() == expected
+    assert parametric.make_bound_program({"beta": 1.5}).to_ir() == expected
```
laurencap, line 795: Weird, that out of all the outputs only this one changed. I don't think it's anything to worry about for this PR though

test/unit_tests/braket/experimental/autoqasm/test_operators.py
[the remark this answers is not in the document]
laurencap: Can you make a new issue on the board with the change you desire? Please focus the title of the issue on the action that should be completed by the issue. This reads more like a statement: `if statements are no longer optimized out during binding parameters`

Since `if` statement is just one example of this behavior, let's remove this comment. The action taken is likely more broad than this one test or operator, and an issue on our board will be much more effective to call our attention to this while we are planning.

test/unit_tests/braket/experimental/autoqasm/test_program.py
```diff
@@ -87,36 +87,32 @@ def circuit(angle: float):
 
     @aq.main
     def zne(scale: int, angle: float) -> aq.BitVar:
-        for i in aq.range(scale):
+        if scale:
```
laurencap, line 90: Does it break if you change it back? This changes the behavior of this test and the program

laurencap: Alternatively, we can use the meta-programming style for this

laurencap (changes requested):
Finished reviewing code changes. I have 3 more test files to review

src/braket/experimental/autoqasm/program/program.py
```diff
@@ -341,15 +341,26 @@ def _free_symbol_names(expr: FreeParameterExpression) -> Iterable[str]:
...
+    ) -> None:
         """Register an input parameter if it has not already been registered.
-        Only floats are currently supported.
 
         Args:
             parameter_name (str): The name of the parameter to register with the program.
+            parameter_type (Optional[Union[float, int]]): The type of the parameter to register
+                with the program.
         """
         if parameter_name not in self._free_parameters:
```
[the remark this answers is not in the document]
ajberdy: Ah and raise something like `TypeError(f"{parameter_name}" exists as an input variable of type {self._free_parameters[...].type}, cannot create input of the same name of type {parameter_type}.")`? Ya I think that makes a lot of sense, I can add that and a test

src/braket/experimental/autoqasm/program/program.py
ajberdy: Actually, if we check for type collision with existing free parameters, there is a difference between no type specified and float, so maybe None would be a more suitable default if we want to allow code like
```
@aq.main
def func(a: int):
    for i in aq.range(FreeParameter("a")):
        ...
```
default float would raise an exception

src/braket/experimental/autoqasm/program/program.py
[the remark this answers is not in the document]
ajberdy: Related to the comment right below this one, I'm thinking the way we keep the type info gives some edge cases, ie:
```
@aq.main
def func(a):
    for i in aq.range(FreeParameter("a")):
        ...
```
could theoretically (if we capture range index context for a) cause a type conflict for `a` (no type specified = float default, but then used as an int).

We may want to update the way we store input types in the Program class (uninitialized, or with an associated type) and may change our internal structure here as we move Oqpy further away.

I've implemented this check locally, but I propose waiting to add it until the Program refactor, or at least as a separate PR from this one. I've captured it is an issue: #814

src/braket/experimental/autoqasm/api.py
ajberdy: I'd prefer to enforce using argument names over position here, but I'm okay doing that in a separate PR. I'll revert for now

src/braket/experimental/autoqasm/api.py
```diff
@@ -218,16 +231,31 @@ def _convert_main(
...
 
+    kwargs = {}
+    parameters = inspect.signature(f).parameters
+
     with aq_program.build_program(user_config) as program_conversion_context:
+        # Capture inputs
```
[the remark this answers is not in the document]
ajberdy: technically we're talking about the inputs to the decorated function here, not the decorator `main`

src/braket/experimental/autoqasm/instructions/gates.py
ajberdy: We don't have an existing Union type here, but I can make one. I think it might be more fitting to address this together with other similar cases. I'm still in favor of adding the oqpy reference here to make sure this case gets lumped in with the others and not missed.

src/braket/experimental/autoqasm/pulse/pulse.py
```diff
@@ -129,14 +130,15 @@ def capture_v0(frame: Frame) -> None:
 
 def delay(
     qubits_or_frames: Union[Frame, list[Frame], QubitIdentifierType, list[QubitIdentifierType]],
-    duration: Union[float, oqpy.FloatVar],
+    duration: Union[float, oqpy.FloatVar, FreeParameterExpression],
```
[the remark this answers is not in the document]
ajberdy: Yes, we do support this already. Here is the [existing test](https://github.com/amazon-braket/amazon-braket-sdk-python/pull/804/files#diff-5ad2d5434a4a1aa49bf65352ef13aedd563eb9f2b4a42af5d69b24133ae8c651L189)

test/unit_tests/braket/experimental/autoqasm/test_operators.py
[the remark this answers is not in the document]
ajberdy: Yes, I will remove this comment, since it is no longer relevant once I switch this test back to using an in-scope python variable for `a` instead of an input. I've captured the general todo here in https://github.com/amazon-braket/amazon-braket-sdk-python/issues/815

test/unit_tests/braket/experimental/autoqasm/test_parameters.py
```diff
@@ -121,11 +156,14 @@ def parametric():
 cnot __qubits__[0], __qubits__[1];
 rx(theta) __qubits__[0];
 rx(alpha) __qubits__[1];"""
-    assert parametric().to_ir() == expected
+    assert parametric.to_ir() == expected
 
 
 def test_parameter_in_subroutine():
     """Test that parameters in subroutines are declared appropriately."""
+    # todo: the openqasm generated here isn't strictly valid
```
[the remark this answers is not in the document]
ajberdy: To confirm, the issue should be around the invalid OpenQASM, correct? And then the Program class comment is generally about the fact that subroutines should be able to access inputs?

If we want to fully stick to the spec here, we may need the Program class to abstract this functionality by adding extra parameters to subroutines to capture used inputs (or something)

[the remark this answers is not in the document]
ajberdy: I have created a [draft issue](https://github.com/orgs/amazon-braket/projects/2/views/1?pane=issue&itemId=45874521)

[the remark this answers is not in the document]
ajberdy: And I'll update the todos in a batch

test/unit_tests/braket/experimental/autoqasm/test_parameters.py
```diff
@@ -142,13 +180,16 @@ def rx_alpha(int[32] qubit) {
...
+    assert parametric.to_ir() == expected
 
 
 def test_captured_parameter():
     """Test that a parameter declared in a larger scope is captured
     and functions correctly.
     """
+    # todo: if we enable directly calling Programs to run on device,
+    # these tests should be updated to ensure alpha becomes a kwarg
+    # of the Program callable
```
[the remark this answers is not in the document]
ajberdy: I have added a [draft issue](https://github.com/orgs/amazon-braket/projects/2/views/1?pane=issue&itemId=45874986) and will remove this comment

test/unit_tests/braket/experimental/autoqasm/test_parameters.py
ajberdy: As discussed offline and in other comments, the program `__call__` method may yet be used for other purposes, like running the program. This use of `make_bound_program` feels pretty similar to how it was before this change, the IR isn't modified at all; the only thing that's changed is how the FP is declared.

In general, definitely open to finding a smoother interface here though

test/unit_tests/braket/experimental/autoqasm/test_parameters.py
ajberdy: Sure! Definitely got a laugh when this test started failing and I realized that as written, with basis as an int, this should have been failing the whole time and the new failure was an improvement lol

test/unit_tests/braket/experimental/autoqasm/test_parameters.py
ajberdy: It's because beta isn't a free parameter, it's inlined during program creation. I will update this test like all the others

test/unit_tests/braket/experimental/autoqasm/test_program.py
ajberdy: Yes, even with the old explicit syntax, FreeParameters inside an aq range breaks. For this test, I'll flip back to meta-programming and I'll create an issue for the bug

ajberdy: https://github.com/orgs/amazon-braket/projects/2/views/1?pane=issue&itemId=45876614

ajberdy:
Updated all tests and incorporated all currently available feedback. Top of mind remaining to get this merged: 
- `test_transpiler.py` could use review, particularly `test_partial_function` and `test_classmethod`.
- example notebooks need to be updated

examples/autoqasm/1_Getting_started_with_AutoQASM.ipynb
```diff
@@ -80,8 +80,7 @@
     }
```
laurencap, line 80: Looks good. Please update the text, too. For example: "The quantum program is returned by calling the decorated bell_state function."

test/unit_tests/braket/experimental/autoqasm/test_api.py
```diff
@@ -865,16 +916,16 @@ def empty_program() -> None:
...
-    assert empty_program().to_ir() == expected
+    assert empty_program.to_ir() == expected
 
 
 def test_main_return():
-    @aq.main
-    def main() -> int:
-        return 1
-
+    # TODO: outputs?
```
laurencap, line 923: bump -- reminder to address this

test/unit_tests/braket/experimental/autoqasm/test_api.py
```diff
@@ -954,4 +1003,60 @@ def nothing() {
 }
 nothing();"""
 
-    assert main().to_ir() == expected
+    assert main.to_ir() == expected
+
+
+def test_input_types():
```
laurencap, line 1009: The title of this test had me looking at the types of x, y and z.
```suggestion
def test_main_decorator_args():
```

test/unit_tests/braket/experimental/autoqasm/test_api.py
```diff
@@ -954,4 +1003,60 @@ def nothing() {
...
+}"""
+    assert (
+        multiple_input_types.to_ir()
+        == multiple_input_types_parens.to_ir()
+        == multiple_input_types_params.to_ir()
+        == expected_ir
+    )
+
+
+def test_input_qubit_indices():
```
laurencap, line 1044: cool!

laurencap (changes requested):
Will the other example notebooks need to be updated?

ajberdy:
> Will the other example notebooks need to be updated?
> 
> If so, can you keep a running TODO list in the PR description to help us stay in sync with you?

Yes, I highlighted the remaining work in the comment above, but I agree the description may be easier to reference. I expect all the examples will need updates, I've only gone through the first two yet though

src/braket/experimental/autoqasm/program/program.py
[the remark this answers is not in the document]
ajberdy: Sounds good! There is a todo here linking to the github issue #814

test/unit_tests/braket/experimental/autoqasm/test_api.py
ajberdy: I believe this is out of scope for this PR. I believe it's on our radar already, so would you prefer I remove this todo?

test/unit_tests/braket/experimental/autoqasm/test_api.py
ajberdy: Yes, that's correct. The test asserts that x, y, z, u get captured as inputs of the correct types. It does this for multiple decorator setups, but the main focus is the input types. I think the test name is valid

examples/autoqasm/1_Getting_started_with_AutoQASM.ipynb
rmshaffer, line 1: Just a note, please also run `tox -e notebooks` to test the service-dependent example notebooks that currently don't run in CI.

test/unit_tests/braket/experimental/autoqasm/test_transpiler.py
```diff
@@ -71,50 +68,86 @@ def bell(q0: int, q1: int):
...
-    bell_noarg_partial = functools.partial(bell_decorated, 1, 3)
-    assert bell_noarg_partial().to_ir() == expected
+    bell_noarg_partial = aq.main(num_qubits=2)(functools.partial(bell, 1, 3))
+    assert bell_noarg_partial.to_ir() == expected_no_arg_partial
 
 
 def test_classmethod() -> None:
     """Tests aq.main decorator application to a classmethod."""
 
+    # TODO: see if this functionality should work
```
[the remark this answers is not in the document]
ajberdy: Yup makes sense to me. Not supporting is a one way door, as we can always build out some level of support later if necessary

examples/autoqasm/1_Getting_started_with_AutoQASM.ipynb
ajberdy: Ah, thanks for pointing this out! Just confirmed that it passes with the latest updates; I'll note this in the PR description

test/unit_tests/braket/experimental/autoqasm/test_transpiler.py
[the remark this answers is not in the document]
ajberdy: Ended up putting a smaller version of the test back to hit the line lol

test/unit_tests/braket/experimental/autoqasm/test_transpiler.py
```diff
@@ -71,50 +68,74 @@ def bell(q0: int, q1: int):
...
+    expected_partial = """OPENQASM 3.0;
+input int[32] q1;
+qubit[2] __qubits__;
+h __qubits__[1];
+cnot __qubits__[1], __qubits__[q1];"""
 
-    expected = """OPENQASM 3.0;
-qubit[4] __qubits__;
+    expected_no_arg_partial = """OPENQASM 3.0;
+qubit[2] __qubits__;
```
laurencap, line 78: The qubit length needs to be max qubit index + 1

```suggestion
qubit[4] __qubits__;
```

test/unit_tests/braket/experimental/autoqasm/test_transpiler.py
```diff
@@ -71,50 +68,74 @@ def bell(q0: int, q1: int):
...
 h __qubits__[1];
 cnot __qubits__[1], __qubits__[3];"""
 
-    bell_partial = aq.main(functools.partial(bell, 1))
-    assert bell_partial(3).to_ir() == expected
+    bell_partial = aq.main(num_qubits=2)(functools.partial(bell, 1))
```
laurencap, line 82: Why do we have to supply `num_qubits` on this input now? Maybe it's a rare enough use case that we don't have to worry about it. @rmshaffer, thoughts?

test/unit_tests/braket/experimental/autoqasm/test_transpiler.py
```diff
@@ -71,50 +68,74 @@ def bell(q0: int, q1: int):
...
 
-    assert aq.main(MyClass.bell)(1, 3).to_ir() == expected
-    assert MyClass.bell_decorated(1, 3).to_ir() == expected
+    assert aq.main(num_qubits=2)(MyClass.rx).to_ir() == expected
+    assert MyClass.rx_decorated.to_ir() == expected
 
     a = MyClass()
-    assert aq.main(a.bell)(1, 3).to_ir() == expected
-    assert a.bell_decorated(1, 3).to_ir() == expected
+    assert aq.main(num_qubits=2)(a.rx).to_ir() == expected
```
laurencap, line 137: Shouldn't `rx` be called with `q0`?

test/unit_tests/braket/experimental/autoqasm/test_transpiler.py
```diff
@@ -71,50 +68,74 @@ def bell(q0: int, q1: int):
...
-    bell_decorated_partial = functools.partial(bell_decorated, 1)
-    assert bell_decorated_partial(3).to_ir() == expected
-
-    bell_noarg_partial = functools.partial(bell_decorated, 1, 3)
-    assert bell_noarg_partial().to_ir() == expected
+    bell_noarg_partial = aq.main(num_qubits=2)(functools.partial(bell, 1, 3))
+    assert bell_noarg_partial.to_ir() == expected_no_arg_partial
 
 
 def test_classmethod() -> None:
```
laurencap, line 89: @rmshaffer @yitchen-tim is this important functionality? It seems like we lose this feature, of AutoQASM instance and class methods, if we go this route

test/unit_tests/braket/experimental/autoqasm/test_transpiler.py
```diff
@@ -71,50 +68,74 @@ def bell(q0: int, q1: int):
...
+qubit[2] __qubits__;
+h __qubits__[q0];
+cnot __qubits__[q0], __qubits__[q1];"""
+
+    assert aq.main(num_qubits=2)(MyClass.bell).to_ir() == expected
+    assert aq.main(num_qubits=2)(MyClass().bell).to_ir() == expected
+
+
+def test_method() -> None:
+    """Tests aq.main decorator application to a classmethod."""
```
laurencap, line 112:
```suggestion
    """Tests aq.main decorator application to an instance method."""
```

ajberdy: Ah I actually meant to remove this test in favor of the one above it, especially since it's not even passing and @rmshaffer weighed in that we don't need to prioritize this functionality at this time

test/unit_tests/braket/experimental/autoqasm/test_transpiler.py
ajberdy: This line passes the function `a.rx` to the decorator; the function isn't actually called at all. You can see from the expected ir that `q0` is captured from the signature as an input. Regardless, I will delete this test as mentioned above

test/unit_tests/braket/experimental/autoqasm/test_transpiler.py
ajberdy: Do we have any plans to validate this? This scenario seems like an example of when we have sufficient information to assert this

test/unit_tests/braket/experimental/autoqasm/test_transpiler.py
ajberdy: Here we need `num_qubits` because the program is built with variable qubit indices in use. For the case below, where we pass in a partial fully binding the qubit args, `num_qubits` can be omitted. This seems correct to me.

It's worth noting that this function previously always built the Program with all variables bound, even though the `aq.main` call was on varying partials. Now that building happens during the decorator call, incomplete partials will have free parameters.

We could test partials on a meta-programming approach that would more closely match the end result of the previous test, but I think this approach is more fitting, as the essence of this test was asserting the behavior of decorating partial functions.

rmshaffer: Agreed, this new behavior seems reasonable and the test makes sense to me.

test/unit_tests/braket/experimental/autoqasm/test_transpiler.py
ajberdy: Here is some context from [Ryan's comment](https://github.com/amazon-braket/amazon-braket-sdk-python/pull/804#discussion_r1408403018) on this test:
> I believe I added these class method tests primarily for getting to 100% code coverage in transpiler, and since it was pretty easy to make it work in the old model.
>
>I think it's legitimate to say that this no longer makes sense in the world where the program conversion happens at function declaration time, and we don't need to support this.

laurencap (commented):
Reviewed through notebook 4

test/unit_tests/braket/experimental/autoqasm/test_transpiler.py
rmshaffer: I don't think it's important. A user can still easily make a class method which wraps an `@aq.main` function and returns the `Program` (which is essentially what it previously meant to mark an instance/class method as `@aq.main`).

examples/autoqasm/3_Iterative_phase_estimation.ipynb
```diff
@@ -82,13 +82,12 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "@aq.main(num_qubits=2)\n",
-    "def ipe(n_iterations: int):\n",
-    "    \"\"\"Iterative phase estimation algorithm.\n",
+    "# TODO: consider using n_iterations as an input (blocked by #818)\n",
```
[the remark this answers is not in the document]
ajberdy: Yes, this is us todoing; I don't intend for the todo to become a part of the notebook. As written, with `n_iterations` outside of the function scope matches the functionality of the example before this change.

However, if a customer wanted to make a version of the program that worked for multiple values of `n_iterations`, the most accessible and imo recommended approach would be to use an input variable, rather than build a meta-programming wrapper around the program. I think this could be a good place to showcase that functionality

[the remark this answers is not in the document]
ajberdy: Since this would break today, I'm open to dismissing this option for this PR, though maybe we should consider the update with https://github.com/amazon-braket/amazon-braket-sdk-python/pull/820, which fixes #818

I'll remove the todo here

rmshaffer:
@ajberdy what's the current status of this PR... seems like it's getting close? If you could please just ping @laurencap and me when all of the TODOs/comments addressed (or have issues opened) and you are ready for a final PR review pass.

ajberdy:
> @ajberdy what's the current status of this PR... seems like it's getting close? If you could please just ping @laurencap and me when all of the TODOs/comments have been addressed (or have issues opened) and you are ready for a final PR review pass.

@rmshaffer Yes, looks like it's getting close! There are 3 TODOs on this PR, all tagged with issues and to be fixed in later PRs (#809, #814, #816). I believe @laurencap has a couple example notebooks left to review and possibly some remaining test files.

I've responded and acted on all the comments, so there aren't currently any action items for the PR on my end. Once all the files are reviewed, I'd say we're good for a final pass.

rmshaffer (approved):
A few very minor comments. Overall looks good to me, approving from my end but also deferring to @laurencap after she has a chance to re-review.

test/unit_tests/braket/experimental/autoqasm/test_api.py
```diff
@@ -22,30 +22,30 @@
...
-from braket.experimental.autoqasm.instructions import cnot, h, measure, x
+from braket.experimental.autoqasm.instructions import cnot, h, measure, rx, x
 from braket.tasks.local_quantum_task import LocalQuantumTask
 
 
-def _test_on_local_sim(program: aq.Program) -> None:
+def _test_on_local_sim(program: aq.Program, inputs=None) -> None:
     device = LocalSimulator(backend=StateVectorSimulator())
-    task = device.run(program, shots=10)
+    task = device.run(program, shots=10, inputs=inputs or {})
```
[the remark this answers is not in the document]
ajberdy: Yes, Python behaves weirdly when you put a mutable object as a default in a function signature (if you mutate the object, it mutates the default for the function). ([source](https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments))

test/unit_tests/braket/experimental/autoqasm/test_api.py
```diff
@@ -924,20 +972,20 @@ def bell() -> None:
         bell()
 
 
+@pytest.mark.xfail(reason="will fix by adding a checker to the transpiler")
```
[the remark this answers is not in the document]
ajberdy: Yes, I'll update this xfail to link to #809

examples/autoqasm/5_Pulse_programming_and_dynamical_decoupling.ipynb
```diff
@@ -69,8 +69,12 @@
    "metadata": {},
    "outputs": [],
    "source": [
+    "# TODO: determine if this example should make these variables inputs to the program\n",
```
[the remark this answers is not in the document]
ajberdy: Technically, should work with qubit as an int. But agree that's not as straightforward as using it directly

test/unit_tests/braket/experimental/autoqasm/test_api.py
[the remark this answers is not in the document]
rmshaffer: I had no idea.

test/unit_tests/braket/experimental/autoqasm/test_api.py
laurencap: That's very confusing

test/unit_tests/braket/experimental/autoqasm/test_parameters.py
laurencap: bump

laurencap (approved):

Issue #811, opened 2023-11-26
Cannot obtain the unitary of circuits imported via `from_ir` when pi is used


jcjaskula-aws:
**Describe the bug**
Pi is interpreted as a FreeParameter when one imports a circuit from an OpenQASM3. However, it cannot be substituted and it is not recognized as a float or a sympy.Float so it remains a FreeParameter. 

Additionally, the following snippet works but should not:
```
from braket.parametric import FreeParameter
from braket.circuits import Circuit
from braket.devices import LocalSimulator

device = LocalSimulator()
pi=FreeParameter("pi")
c=Circuit().rx(0, pi)
result = device.run(c, shots=100).result()
```

As `pi` is a builtin constant in OQ3, we could forbid the declaration of the pi identifier and standardize the behavior between the BDK and the local simulator.

**To reproduce**
```
from braket.circuits import Circuit

qasm_str = """OPENQASM 3.0;
qubit[1] q;
bit[1] c;
rx(π) q[0];
c[0] = measure q[0];
"""

circuit = Circuit.from_ir(qasm_str)
circuit.to_unitary()
```

**Expected behavior**
As pi seems to be defined as sympy.Pi (see circuit.instructions[0].operator.angle.expression), unitary could be evaluated if the FreeParameterExpression is evaluated in `to_matrix`.

Edit: remove unnecessary code in the second snippet.

PR #697, opened 2023-09-11
feat: update OQpy to version 0.3.5

jcjaskula-aws:
*Issue #, if available:*
OQpy pushed some bug fixes that will be needed for future features such as in-place modification of pulse sequences.

With the new version of OQpy, we could support parametric pulse sequence (especially with duration types where OQpy support notations such `(c) * 1s` out of the box.

*Description of changes:*
- update oqpy to 0.3.5
- add a _to_oqpy_expression method to FreeParameterExpression. Once if is translated to an OQPyExpression, the FPE will handle types and AST generation out of the box. 
- update tests to match new duration variable printing
- closes #828 
- closes #822 

*Testing done:*
Ran the test suite.

## Merge Checklist

_Put an `x` in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your pull request._

#### General

- [x] I have read the [CONTRIBUTING](https://github.com/aws/amazon-braket-sdk-python/blob/main/CONTRIBUTING.md) doc
- [x] I used the commit message format described in [CONTRIBUTING](https://github.com/aws/amazon-braket-sdk-python/blob/main/CONTRIBUTING.md#commit-your-change)
- [x] I have updated any necessary documentation, including [READMEs](https://github.com/aws/amazon-braket-sdk-python/blob/main/README.md) and [API docs](https://github.com/aws/amazon-braket-sdk-python/blob/main/CONTRIBUTING.md#documentation-guidelines) (if appropriate)

#### Tests

- [x] I have added tests that prove my fix is effective or that my feature works (if appropriate)
- [x] I have checked that my tests are not configured for a specific region or account (if appropriate)

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #829, opened 2023-12-07
fix: treating OpenQASM builtin types as constants

krneta:
*Issue #, if available:*
https://github.com/amazon-braket/amazon-braket-sdk-python/issues/811

*Description of changes:*

This fixes a part of the problem, but it has the serious drawback that the OpenQASM from -> to conversions will now transform symbols to constants.

eg.

`qasm_str = """OPENQASM 3.0;
qubit[1] q;
bit[1] c;
rx(pi) q[0];
c[0] = measure q[0];
"""

circuit = Circuit.from_ir(qasm_str)

ir = circuit.to_ir(ir_type=IRType.OPENQASM)
print(ir.source)
`
will print
`
OPENQASM 3.0;
bit[1] b;
qubit[1] q;
rx(3.141592653589793) q[0];
b[0] = measure q[0];
`

*Testing done:*

## Merge Checklist

#### General

#### Tests

### DIFF
--- a/src/braket/circuits/braket_program_context.py
+++ b/src/braket/circuits/braket_program_context.py
@@ -14,7 +14,7 @@
 from typing import Optional, Union
 
 import numpy as np
-from sympy import Expr
+from sympy import Expr, Number
 
 from braket.circuits import Circuit, Instruction
 from braket.circuits.gates import Unitary
@@ -147,5 +147,8 @@ class BraketProgramContext(AbstractProgramContext):
             otherwise wraps the symbolic expression as a `FreeParameterExpression`.
         """
         if isinstance(value, Expr):
-            return FreeParameterExpression(value)
+            evaluated_value = value.evalf()
+            if isinstance(evaluated_value, Number):
+                return evaluated_value
+            return FreeParameterExpression(evaluated_value)
         return value
--- a/test/unit_tests/braket/circuits/test_circuit.py
+++ b/test/unit_tests/braket/circuits/test_circuit.py
@@ -1718,6 +1718,36 @@ def test_circuit_user_gate(pulse_sequence_2):
                 inputs={},
             ),
         ),
+        (
+            Circuit().rx(0, np.pi),
+            OpenQasmProgram(
+                source="\n".join(
+                    [
+                        "OPENQASM 3.0;",
+                        "bit[1] b;",
+                        "qubit[1] q;",
+                        "rx(π) q[0];",
+                        "b[0] = measure q[0];",
+                    ]
+                ),
+                inputs={},
+            ),
+        ),
+        (
+            Circuit().rx(0, 2 * np.pi),
+            OpenQasmProgram(
+                source="\n".join(
+                    [
+                        "OPENQASM 3.0;",
+                        "bit[1] b;",
+                        "qubit[1] q;",
+                        "rx(τ) q[0];",
+                        "b[0] = measure q[0];",
+                    ]
+                ),
+                inputs={},
+            ),
+        ),
     ],
 )
 def test_from_ir(expected_circuit, ir):


src/braket/circuits/braket_program_context.py
```diff
@@ -147,5 +147,8 @@ def handle_parameter_value(
             otherwise wraps the symbolic expression as a `FreeParameterExpression`.
         """
         if isinstance(value, Expr):
-            return FreeParameterExpression(value)
+            evaluated_value = value.evalf()
```
ajberdy, line 150: Is there a list of built-in constants that will be evaluated here?

test/unit_tests/braket/circuits/test_circuit.py
```diff
@@ -1718,6 +1718,36 @@ def foo(
...
+        ),
+        (
+            Circuit().rx(0, 2 * np.pi),
+            OpenQasmProgram(
+                source="\n".join(
+                    [
+                        "OPENQASM 3.0;",
+                        "bit[1] b;",
+                        "qubit[1] q;",
+                        "rx(τ) q[0];",
```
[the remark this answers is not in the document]
ajberdy: Honestly, might as well test all three built-ins

src/braket/circuits/braket_program_context.py
krneta: I'm not sure that I understand. It's not just constants - the reason why I have a tau test is because the expression could actually be a math expression, such as 2 * pi.

test/unit_tests/braket/circuits/test_circuit.py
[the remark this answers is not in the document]
krneta: I think it's overkill to test all 6. I test pi and tau since pi is a constant and tau is a formula (2 * pi), and not both of them worked the same way.

src/braket/circuits/braket_program_context.py
math411: In the case of expressions, we can split the string and check against the set. There are probably many tokens in the [object](https://github.com/openqasm/openqasm-pygments/blob/734ccf7a46f10e37406f1103368c552debdb0fb6/src/openqasm_pygments/qasm3.py#L75) we should check against so that customers does not submit openqasm reserved words in their program.

math411: Or rather, this would be regex matching against an iterable generated from the linked code.

ajberdy: I would think we can avoid using regex here and rely on Sympy for any substitutions/validation

krneta: +1, I do not think we should be doing any additional parsing/interpreting here.

krneta:
> What happens if the constant is also supplied as an input? (That is, if the qasm str has: `input float pi;` at the top, what's the behavior with the service if I pass in a parameter value `{"pi": 0.1}`)

I've had an offline conversation with Aaron about this. The input is ignored in the Local Simulator, before and after this change. This is a separate issue that I'm considering.

krneta:
> On this branch, free parameters with reserved keywords will now get assigned the constant value. We should probably raise an error instead (this should be pretty easy to add to the FP class).

Yes, there are two problems with the issue that's linked above.

1. Customers may expect they can change the value, but they can't. Using a free parameter with one of these default constants will silently invalidate their input and use the default value.

2. We don't use these default constant values elsewhere in the SDK. So, if the customer creates a circuit using one of these constants as a FreeParameter, it will be treated as a FreeParameter in the BDK (meaning it will fail to generate a unitary), but not be treated that way in the Local Simulator.

This PR addresses problem 2, but doesn't really address problem 1. I am hesitant to tie FreeParameter validation to OpenQASM, since they should be separate. I'll consider this problem in a separate PR.

krneta:
> However, on main we have arguably worse behavior where the fp is converted to a sympy symbol (not a free parameter)

It's much worse, IMO, since there is no indication that the FreeParameter won't actually be treated as a FreeParameter, and there's no indication if passing in an input will do anything or not.

kshitijc:
Thanks! Could you also please describe what the behavior with the service is today?

krneta:
I don't think we should make an exception for tau. It's in the spec and it's the current behaviour in the local sim.

krneta:
> Should we consider setting these variables to their constant value in the ApproximationParser (related to [this comment](https://github.com/amazon-braket/amazon-braket-sdk-python/pull/697#discussion_r1344689117))?

This change does change the variables to their constant values.

kshitijc (approved):

PR #838, opened 2023-12-12
feature: Support OpenQASM annotations in AutoQASM programs

### LINKED ISSUE #837: Support OpenQASM annotations in AutoQASM

rmshaffer:
*Issue #, if available:*
Resolves #837 

*Description of changes:*
Support `annotations` argument for subroutines, variables, ranges, and boxes. This will result in an OpenQASM annotation being added to the serialized program.

For example:
```
@aq.subroutine(annotations=["inline"])
def subroutine_test():
    pass

@aq.main
def main():
    subroutine_test()
```
produces:
```
OPENQASM 3.0;
@inline
def subroutine_test() {
}
subroutine_test();
```

*Testing done:*
Tests added to verify new functionality. tox passes.

## Merge Checklist

#### General

#### Tests

### DIFF
--- a/src/braket/experimental/autoqasm/api.py
+++ b/src/braket/experimental/autoqasm/api.py
@@ -95,7 +95,7 @@ def main(
 
 
 def subroutine(
-    func: Optional[Callable] = None, annotations: Optional[Iterable[str | tuple[str, str]]] = None
+    func: Optional[Callable] = None, annotations: Optional[str | Iterable[str]] = None
 ) -> Callable[..., aq_program.Program]:
     """Decorator that converts a function into a callable that will insert a subroutine into
     the quantum program.
@@ -103,9 +103,7 @@ def subroutine(
     Args:
         func (Optional[Callable]): Decorated function. May be `None` in the case where decorator
             is used with parentheses.
-        annotations (Optional[Iterable[str | tuple[str, str]]]): Annotations to be added to
-            the subroutine. The annotations can be either a string or a tuple of the form
-            (annotation_name, annotation_value).
+        annotations (Optional[str | Iterable[str]]): Annotations to be added to the subroutine.
 
     Returns:
         Callable[..., Program]: A callable which returns the converted
@@ -115,7 +113,7 @@ def subroutine(
         func,
         converter_callback=_convert_subroutine,
         converter_args={
-            "annotations": annotations or [],
+            "annotations": aq_types.make_annotations_list(annotations),
         },
     )
 
@@ -316,7 +314,7 @@ def _add_qubit_declaration(program_conversion_context: aq_program.ProgramConvers
 def _convert_subroutine(
     f: Callable,
     options: converter.ConversionOptions,
-    annotations: Iterable[str | tuple[str, str]],
+    annotations: Iterable[str],
     args: list[Any],
     kwargs: dict[str, Any],
 ) -> None:
@@ -329,9 +327,7 @@ def _convert_subroutine(
     Args:
         f (Callable): The function to be converted.
         options (converter.ConversionOptions): Converter options.
-        annotations (Iterable[str | tuple[str, str]]): Annotations to be added to
-            the subroutine. The annotations can be either a string or a tuple of the form
-            (annotation_name, annotation_value).
+        annotations (Iterable[str]): Annotations to be added to the subroutine.
         args (list[Any]): Arguments passed to the program when called.
         kwargs (dict[str, Any]): Keyword arguments passed to the program when called.
     """
--- a/src/braket/experimental/autoqasm/api.py
+++ b/src/braket/experimental/autoqasm/api.py
@@ -20,7 +20,7 @@ import functools
 import inspect
 from collections.abc import Callable
 from types import FunctionType
-from typing import Any, Optional, Union, get_args
+from typing import Any, Iterable, Optional, Union, get_args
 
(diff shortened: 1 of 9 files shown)


src/braket/experimental/autoqasm/api.py
```diff
@@ -94,19 +94,30 @@ def main(
...
     Returns:
         Callable[..., Program]: A callable which returns the converted
         quantum program when called.
     """
-    return _function_wrapper(func, converter_callback=_convert_subroutine)
+    return _function_wrapper(
+        func,
+        converter_callback=_convert_subroutine,
+        converter_args={
+            "annotations": annotations or [],
```
laurencap, line 118: Nice, seems like it was pretty straightforward to add this arg

src/braket/experimental/autoqasm/api.py
```diff
@@ -336,7 +351,9 @@ def _convert_subroutine(
 
             # Convert the function via autograph into an oqpy subroutine
             # NOTE: Process a clone of the function so that we don't modify the original object
-            oqpy_sub = oqpy.subroutine(_wrap_for_oqpy_subroutine(_clone_function(f), options))
+            oqpy_sub = oqpy.subroutine(annotations=annotations)(
```
laurencap, line 354: Interesting interface from oqpy for adding annotations. Easy enough though

src/braket/experimental/autoqasm/program/program.py
```diff
@@ -620,18 +629,19 @@ def else_block(self) -> contextlib._GeneratorContextManager:
         return self._control_flow_block(oqpy.Else(oqpy_program))
 
     def for_in(
-        self, iterator: oqpy.Range, iterator_name: Optional[str]
+        self, iterator: aq_types.Range, iterator_name: Optional[str]
```
laurencap, line 632: Nice one

src/braket/experimental/autoqasm/types/types.py
```diff
@@ -13,7 +13,9 @@
 
 """AutoQASM types and type utilities."""
 
-from typing import Any, Optional
+from __future__ import annotations
```
laurencap, line 16: Confusing, this is not the same annotations as we're mostly talking about here, this is some Python lang thing?

src/braket/experimental/autoqasm/types/types.py
```diff
@@ -41,21 +43,29 @@ def is_qasm_type(val: Any) -> bool:
...
+    ):
+        """Creates a range definition.
+
+        Args:
+            start (int): Start of the range.
+            stop (Optional[int]): End of the range. Defaults to None.
+            step (Optional[int]): Step of the range. Defaults to 1.
+            annotations (Optional[Iterable[str | tuple[str, str]]]): Annotations for the range.
+                The annotations can be either a string or a tuple of the form
+                (annotation_name, annotation_value).
```
laurencap, line 62: What's this? Why not only simple strings?

laurencap (approved):
Some comments. I think it's fine for a first iteration

src/braket/experimental/autoqasm/types/types.py
ajberdy: I would guess this maps to `@annotation_name annotation_value`, but I agree it's confusing. The OpenQASM spec (and AST) only defines annotations as containing a string. If my understanding of this is correct and we want to expose this feature, maybe we can give an example in the doc string.

src/braket/experimental/autoqasm/api.py
rmshaffer: It seems like `oqpy.subroutine(f, annotations=annotations)` should have worked, but it doesn't. @jcjaskula-aws may have more insight. I think the `oqpy.subroutine` is optimized for use as a decorator rather than being called directly. But in any case, yes, this is easy enough.

src/braket/experimental/autoqasm/types/types.py
rmshaffer: This is a good point, I was blindly copying oqpy behavior here. I'd be open to simplifying this in AutoQASM to just support a list of strings. After all, if someone wants the OQ3 style `@annotation_name annotation_value` they could just pass the string `"annotation_name annotation_value"`.

rmshaffer: See the OQ3 spec, which implies the name/value structure: https://openqasm.com/language/directives.html#annotations

But I do agree we could just represent this as a single string.

ajberdy: +1 I think this is a case where we can think about what feature *we* want to create and then map that to the IR feature.

rmshaffer: Ok, I've updated this PR so that our API takes annotations as `Optional[str, Iterable[str]]`. It's a little bit of an annoying type hint, but if we only accept `Iterable[str]` we lead the user right into the mistake of:
```
@aq.subroutine(annotations="inline")
```
which, because `str` is itself `Iterable[str]`, would produce
```
@i
@n
@l
@i
@n
@e
```

ajberdy: I don't mind the `Optional[str, Iterable[str]]` type. [Job decorator](https://github.com/amazon-braket/amazon-braket-sdk-python/blob/main/src/braket/jobs/hybrid_job.py#L49) does the same thing for a couple arguments

ajberdy (approved):

Issue #832, opened 2023-12-07
Bug: NameErrors dependent on function ordering


ajberdy:
In Python, I can do this:

```
def func1():
    func2()

def func2():
    pass
```

but now we get a NameError if we have

```
@aq.main
def func1():
    func2()

@aq.subroutine
def func2():
    pass
```

This should be valid code

ajberdy:
I wonder if we can fix this by keeping a registry of defined subroutines that the Program can references when run. In python we see the following behavior:
```
def f1():
    f2()


def f2():
    print("f2")


f1()
# prints "f2"


def f2():
    print("F2")

f1()
prints "F2"

```
Before the subroutine is defined, we can label it as an extern function in the Program object so the openqasm is valid

rmshaffer:
I wonder if we could make the decorators use a lazy initialization pattern. So nothing really happens when the function is defined - it just returns a `Program` object in which the conversion has not happened yet. Then the first time the `Program` object is _used_ (i.e. any method or property is accessed), then the conversion happens.

rmshaffer:
Closing this now that #841 has been merged into feature/autoqasm.

PR #841, opened 2023-12-19
feat: Explicit build API for AutoQASM main program

rmshaffer:
*Issue #, if available:*
#832 

*Description of changes:*
Wait to perform the main AutoQASM program conversion until `.build()` is called.

**Note:** Although we will want to enforce that `.build()` is called before calling `.to_ir()`, for the purposes of this PR, `.to_ir()` still calls `.build()` implicitly. This avoids introducing a massive amount of mechanical & documentation changes that will be required when we flip over to requiring a call to `.build()`. I'll make those changes in a follow-up PR so that they are not interspersed with the API and functional changes here.

*Testing done:*
tox passes, manual testing done and unit tests added/updated to confirm new behavior.

## Merge Checklist

#### General

#### Tests


ajberdy:
Thanks for putting this together, this is is really cool. I think it's worth asking if we want to have lazy initialization for all parts of the program, or just a subset of functionality (like subroutine definitions). This arrangement leads to some edge cases that technically match Python's behavior, but may be surprising:
```
theta = 0.1

@aq.main
def my_program():
    rx(0, theta)


theta = 0.2
my_program.display()
```
outputs
```
OPENQASM 3.0;
qubit[1] __qubits__;
rx(0.2) __qubits__[0];
```

Laid out linearly as above seems easy enough to follow, but I wonder if it could become confusing in a notebook. It feels different than the Python behavior for functions, since there the functions are actually being called, whereas here the Program is getting implicitly built because some method is called. Maybe it would be better if less methods triggered the building (i.e. display could just include the subroutine signature with `...` in the body or something, but `to_ir` would actually require building the program), though that may be more confusing too.

I'm also not sure if it's worth having two separate classes for the program wrapper and generated program, especially once we have a better defined `Program` class. Seems like it may be simpler to have one class that is either in an initialized or uninitialized state. Undecided on this though, as I haven't fully internalized the code, open to discussion on all of the above points.

yitchen-tim:
>I think it's worth asking if we want to have lazy initialization for all parts of the program, or just a subset of functionality (like subroutine definitions). 
I am not sure if partial lazy initialization is viable. If a subroutine is defined after main, main does not know if the function is a python function or an AutoQASM function. The two interact differently with aq.main. The later may be simply a replacement of subroutine signature with the real built subroutine. But the former may be computing some numbers that is used in the later part of the main. The main cannot be built without these numbers in this case. 

But I do see that it's strange to the building happen when `program.display()` is called. `display` implies just displaying an built program, but in fact it builds the program. 

I think this is related to the fact that aq.main now means a mixture of function and object. If we think it’s a function, it should only be built when called. If it's an object, it should be built right away. Now aq.main is in the middle. There is no direct call method, but it is also not built when defined.

src/braket/experimental/autoqasm/program/program.py
```diff
@@ -89,6 +90,129 @@ class ProgramMode(Enum):
...
+                the program.
+
+        Returns:
+            Program: Returns a program with all present parameters fixed to their respective
+            values.
+        """
+        self._ensure_initialized()
+        return self._program.make_bound_program(param_values, strict)
+
+    def display(self, ir_type: IRType = IRType.OPENQASM) -> None:
```
laurencap, line 202: This is `print(prog.to_ir())`? We might want to reserve `display` for drawing the program, or some other improved visualization.

rmshaffer:
@ajberdy @yitchen-tim, thanks for the comments. These are very good points. I agree that the `@aq.main` decorated function has some fundamental differences in behavior as compared to ordinary Python functions, so the question is how much of this we can abstract away from the user without introducing potentially confusing behavior.

Another confusing thing with this implementation is that the _first_ time `.to_ir()` is called, the program is built, but on subsequent calls it is not. That confusion may offset any benefit of caching the built program; it may be better to just rebuild it every time (e.g. so that any changes to variables in the outer scope will be picked up).

Moving this PR back to Draft status for now - we can continue to think about this and alternatives.

src/braket/experimental/autoqasm/program/program.py
rmshaffer: Today this is the "pretty print" API that uses `pygments`. But I agree with you. I am also thinking that `display` and similar functionality belong in some separate visualization module rather than being part of the `Program` API.

laurencap:
Very clean solution! I'm impressed how simple it is.

We should pause and think critically about these changes, because it shows that our ideas about the basic AutoQASM interface have been vacillating.

### "Original"
Deferred eval, invoked by user. But calling the program doesn't "run" the program (same as most quantum SDKs, but weird for Python)
```
@aq.main
def my_func(my_args):
    ...
my_func()  # program is built and validated
```

### Existing
Immediate eval. Non-pythonic behavior such as NameErrors for forward references.
```
@aq.main
def my_func(my_args):  # program is built and validated
    ...
my_func()  # No defined behavior
```

### New
Deferred eval, but no single interface for requesting the build -- all program methods have side effects
```
@aq.main
def my_func(my_args):
    ...

my_func.call_any_method()  # program is built and validated
# or
device.run(my_func)  # program is built and validated
```

### Final thoughts
Perhaps we should go for something more explicit:
```
@aq.main
def my_func(arg1, arg2):
    ...

my_func.validate()
built_func = my_func.build()  # Could even have `new_build` and `idempotent_build`, but that 
                              # feels like overengineering
# Can still reserve `my_func()` for calling `device.run(program.build())`
```

yitchen-tim:
An example use case that may be broken by lazy initialization + implicit build.
```
programs = []
for theta in theta_list:
    @aq.main
    def my_program():
        rx(0, theta)
    programs.append(my_program)

task_batch = device.run(programs)
```
The programs are expected to scan through `theta_list`, but they will also use `theta_list[-1]`.

src/braket/circuits/serialization.py
```diff
@@ -38,15 +38,20 @@ class SerializableProgram(ABC):
     def to_ir(
         self,
         ir_type: IRType = IRType.OPENQASM,
+        allow_implicit_build: bool = True,
```
rmshaffer, line 41: As mentioned in the PR description, I'll make a follow-up PR which flips this default to `False` and makes the resulting avalanche of downstream changes, since every existing call to `.to_ir()` will need to be updated to `.build().to_ir()`.

src/braket/experimental/autoqasm/program/program.py
```diff
@@ -87,6 +88,65 @@ class ProgramMode(Enum):
     """For program conversion inside a context where only pulse operations are allowed."""
 
 
+class MainProgram(SerializableProgram):
```
rmshaffer, line 91: I'm not sure I like `MainProgram` as the class name here. Suggestions welcome.

laurencap: +1 I think it's good, agree it could be better. But let's not block on it (and I don't have a suggestion offhand). Let's add an action item to think about in a follow up PR?

src/braket/experimental/autoqasm/program/program.py
```diff
@@ -87,6 +88,65 @@ class ProgramMode(Enum):
...
+    """Represents an AutoQASM program that has not yet been generated. The program can be built by
```
laurencap, line 92: "not yet been generated" is a little vague.

```suggestion
    """Represents an AutoQASM program. The program can be built by
```

src/braket/experimental/autoqasm/program/program.py
```diff
@@ -87,6 +88,65 @@ class ProgramMode(Enum):
...
+    calling build(), or it can be executed by passing it to the run() method of a Device object."""
+
+    def __init__(self, program_generator: Callable[[Device | None], Program]):
+        self._program_generator = program_generator
```
laurencap, line 96: Cool

laurencap (approved):

PR #846, opened 2023-12-22
feature: update circuit drawing

jcjaskula-aws:
*Issue #, if available:*
`N`, `C` characters are not conventional to denote (neg)controlled qubits. 
Circuits can also look nicer by using box drawing characters.

*Description of changes:*
- changed N and C to empty/full unicode circle
- use box drawing characters
- keep ascii diagram as legacy builder 
- move ascii_circuit_diagram.py and boxdrawing_circuit_diagram to text_circuit_diagram_builders/
- create a text_circuit_diagram.py containing helper functions to reuse common code. 

*Testing done:*
updated unit tests.

## Merge Checklist

#### General

#### Tests

### DIFF
--- a/src/braket/circuits/text_diagram_builders/unicode_circuit_diagram.py
+++ b/src/braket/circuits/text_diagram_builders/unicode_circuit_diagram.py
@@ -251,8 +251,9 @@ class UnicodeCircuitDiagram(TextCircuitDiagram):
         symbol = f"┤ {symbol} ├"
         return top, symbol, bottom
 
-    @staticmethod
+    @classmethod
     def _build_verbatim_box(
+        cls,
         symbol: Literal["StartVerbatim", "EndVerbatim"],
         connection: Literal["above", "below", "both", "none"],
     ) -> str:
@@ -267,7 +268,7 @@ class UnicodeCircuitDiagram(TextCircuitDiagram):
             top = "║"
             symbol = "╨"
         top = _fill_symbol(top, " ")
-        symbol = _fill_symbol(symbol, UnicodeCircuitDiagram._qubit_line_character())
+        symbol = _fill_symbol(symbol, cls._qubit_line_character())
         bottom = _fill_symbol(bottom, " ")
 
         return top, symbol, bottom


src/braket/circuits/angled_gate.py
```diff
@@ -46,7 +46,7 @@ def __init__(
                 printing a diagram of a circuit. The length must be the same as `qubit_count`, and
                 index ordering is expected to correlate with the target ordering on the instruction.
                 For instance, if a CNOT instruction has the control qubit on the first index and
-                target qubit on the second index, the ASCII symbols should have `["C", "X"]` to
+                target qubit on the second index, the ASCII symbols should have `["⏺", "X"]` to
```
ajberdy, line 49: I believe this is the wrong character being used. On mobile it shows up as ⏺️. I believe ● is the correct character.

image

test/unit_tests/braket/circuits/test_box_drawing_circuit_diagram.py
```diff
@@ -0,0 +1,986 @@
...
+            target=[1, 2],
+        )
+    )
+    expected = (
+        "T  : │     0     │  1  │     Result Types      │",
+        "                  ┌───┐     ┌─────────────┐     ",
+        "q0 : ───●─────────┤ H ├─────┤ Variance(Y) ├─────",
+        "        │         └───┘     └─────────────┘     ",
+        "        │               ┌─────────────────────┐ ",
+        "q1 : ───┼─────●─────────┤ Expectation(MyHerm) ├─",
```
laurencap, line 756: This doesn't support `Program`s with OpenQASM source, is that right? Asking here because I wonder how `measure q;` would look.

laurencap (commented):
outputs lgtm!

src/braket/circuits/box_drawing_circuit_diagram.py
```diff
@@ -0,0 +1,524 @@
...
+from braket.circuits.gate import Gate
+from braket.circuits.instruction import Instruction
+from braket.circuits.moments import MomentType
+from braket.circuits.noise import Noise
+from braket.circuits.result_type import ResultType
+from braket.registers.qubit import Qubit
+from braket.registers.qubit_set import QubitSet
+
+
+class BoxDrawingCircuitDiagram(CircuitDiagram):
```
ajberdy, line 31: I would imagine the high-level logic for constructing these diagrams is quite similar to `AsciiCircuitDiagram`. Can we share more logic between the two classes? Maybe even make `BoxxDrawingCircuitDiagram` a subclass of `AsciiCircuitDiagram` and factor out common logic into helper methods? That would help maintainability significantly

test/unit_tests/braket/circuits/test_box_drawing_circuit_diagram.py
```diff
@@ -0,0 +1,986 @@
...
+        Circuit()
+        .h(0)
+        .h(1)
+        .h(100)
+        .expectation(observable=Observable.Y() @ Observable.Z(), target=[0, 100])
+    )
+    expected = (
+        "T    : │  0  │    Result Types    │",
+        "        ┌───┐ ┌──────────────────┐ ",
+        "q0   : ─┤ H ├─┤ Expectation(Y@Z) ├─",
```
speller26, line 638: Not for this PR, but it'll be worth replacing `@` with `⊗`

test/unit_tests/braket/circuits/test_box_drawing_circuit_diagram.py
```diff
@@ -0,0 +1,986 @@
...
+        .cnot(1, 3)
+        .h(0)
+        .variance(observable=Observable.Y(), target=0)
+        .expectation(observable=Observable.Y(), target=2)
+        .sample(observable=Observable.Y())
+    )
+    expected = (
+        "T  : │     0     │  1  │          Result Types          │",
+        "                  ┌───┐   ┌─────────────┐  ┌───────────┐ ",
+        "q0 : ───●─────────┤ H ├───┤ Variance(Y) ├──┤ Sample(Y) ├─",
```
speller26, line 684: Shouldn't the `Y` observables be unlinked? This should be four separate Y measurements

ajberdy: For my own context, what does unlinked mean here?

ajberdy: Oh got it, in the same way that `Circuit.h([0, 1, 2, 3])` wouldn't put a link between the gates because they are separate 1-qubit operations

test/unit_tests/braket/circuits/test_box_drawing_circuit_diagram.py
```diff
@@ -0,0 +1,1021 @@
...
+# may not use this file except in compliance with the License. A copy of
+# the License is located at
+#
+#     http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+
```
krneta, line 13: Thanks for doing this, it does look A LOT nicer.

src/braket/circuits/circuit.py
```diff
@@ -82,6 +82,7 @@ class Circuit:
     """
 
     _ALL_QUBITS = "ALL"  # Flag to indicate all qubits in _qubit_observable_mapping
+    default_diagram_builder = BoxDrawingCircuitDiagram
```
krneta, line 85: We may not want to change the default until we've confirmed that there's no breaking changes when running our notebooks.

src/braket/circuits/ascii_circuit_diagram.py
```diff
@@ -31,8 +31,13 @@
...
 
-    @staticmethod
-    def build_diagram(circuit: cir.Circuit) -> str:
+    _vdelim = "|"
+    _qubit_line_char = "-"
+    _add_empty_line = True
+    _buffer_size = 0
+
+    @classmethod
+    def build_diagram(cls, circuit: cir.Circuit) -> str:
```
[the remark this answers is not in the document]
rmshaffer: Why does this need to be a `@classmethod`? Can't it still be a `@staticmethod`? Then it would still match the abstract method in `CircuitDiagram` base class, and no worries about backward incompatibility.

Probably changing `BoxDrawingCircuitDiagram` so that it inherits directly from `CircuitDiagram` would help here.

src/braket/circuits/box_drawing_circuit_diagram.py
```diff
@@ -0,0 +1,263 @@
...
+class BoxDrawingCircuitDiagram(AsciiCircuitDiagram):
+    """Builds ASCII string circuit diagrams using box-drawing characters."""
+
+    _vdelim = "│"
+    _qubit_line_char = "─"
+    _add_empty_line = False
+    _buffer_size = 4
+
+    @staticmethod
+    def build_diagram(circuit: cir.Circuit) -> str:
```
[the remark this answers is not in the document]
rmshaffer: I agree, the `Ascii` and `BoxDrawing` are fundamentally different things and probably should both inherit from `CircuitDiagram` directly.

And yes, I agree, any shared code between the two could be factored into static methods in some `TextCircuitDiagramUtilities` class or something like this.

src/braket/circuits/circuit.py
```diff
@@ -82,6 +82,7 @@ class Circuit:
     """
 
     _ALL_QUBITS = "ALL"  # Flag to indicate all qubits in _qubit_observable_mapping
+    default_diagram_builder = AsciiCircuitDiagram
```
krneta, line 85: I don't think we need this. Let's delete it.

src/braket/circuits/circuit.py
```diff
@@ -1495,7 +1496,7 @@ def __repr__(self) -> str:
             )
 
     def __str__(self):
-        return self.diagram(AsciiCircuitDiagram)
+        return self.diagram(self.default_diagram_builder)
```
krneta, line 1499: I think this should just call self.diagram() and use the default.

I think we should change def diagram() to use the new box drawing class as the default class.

src/braket/circuits/text_diagram_builders/ascii_circuit_diagram.py
```diff
@@ -0,0 +1,176 @@
...
+from braket.circuits.instruction import Instruction
+from braket.circuits.result_type import ResultType
+from braket.circuits.text_diagram_builders import text_circuit_diagram_utils
+from braket.registers.qubit_set import QubitSet
+
+
+class AsciiCircuitDiagram(CircuitDiagram):
+    """Builds ASCII string circuit diagrams."""
+
+    _vdelim = "|"  # Character that connects qubits of multi-qubit gates
```
krneta, line 32: These classes seem to have a mix of functionality and configuration and create a (soft) cyclic dependency. I think we can do either one or the other, and avoid the cycle. 

So, either:
 - we have a configuration that we pass into diagram_utils, and have it do the rest of the work based on the configuration
XOR
 - we have the circuit class call the diagram utils and pass in the diagram classes and have them handle the functionality where the configuration is relevant. 
 
I think it's confusing to have both.

test/unit_tests/braket/circuits/test_circuit.py
```diff
@@ -202,6 +203,12 @@ def test_str(h):
     assert str(h) == expected
 
 
+def test_change_diagram_builder(h):
+    Circuit.default_diagram_builder = BoxDrawingCircuitDiagram
```
krneta, line 207: As above, I think that this new one should be the default, and to change it, customers may need to call circuit.diagram(AsciiCircuitDiagram)

src/braket/circuits/text_diagram_builders/text_circuit_diagram.py
```diff
@@ -0,0 +1,265 @@
...
+    def _create_output(
+        cls,
+        symbols: dict[Qubit, str],
+        margins: dict[Qubit, str],
+        qubits: QubitSet,
+        global_phase: float | None,
+    ) -> str:
+        """Creates the ouput for a single column:
+            a. If there was one or more gphase gate, create a first line with the total global
+            phase shift ending with target_class.vdelim, e.g. 0.14|
```
krneta, line 234: nit: doc here and a couple lines below still references the old names.

src/braket/circuits/text_diagram_builders/unicode_circuit_diagram.py
```diff
@@ -0,0 +1,282 @@
...
+
+                elif qubit in control_qubits:
+                    symbols[qubit] = "C" if map_control_qubit_states[qubit] else "N"
+                else:
+                    symbols[qubit] = "┼"
+
+        output = cls._create_output(symbols, connections, circuit_qubits, global_phase)
+        return output
+
+    @staticmethod
```
krneta, line 138: nit: we could have made this a class method, rather than referencing the class inside it.

src/braket/circuits/text_diagram_builders/unicode_circuit_diagram.py
```diff
@@ -0,0 +1,282 @@
...
+                cls._qubit_line_character() if needed.
+            connection (Literal["above", "below", "both", "none"]): specifies if a connection
+                will be drawn above and/or below the box.
+
+        Returns:
+            str: a string representing the symbol.
+        """
+        top = ""
+        bottom = ""
+        if symbol in ["C", "N", "SWAP"]:
```
krneta, line 216: I'm not crazy about hard-coding these (again?)... couldn't we just have put in the correct symbol on line 131? For SWAP, why are we not just changing the symbol in the class, rather than here?

krneta (approved):

Issue #847, opened 2023-12-27
aq.subroutine does not support aq.Qubit argument type


ajberdy:
```
def bell(int[32] q0, int[32] q1) {
    h __qubits__[q0];
    cnot __qubits__[q0], __qubits__[q1];
}
```
should be replaced with
```
def bell(qubit q0, qubit q1) {
    h q0;
    cnot q0, q1;
}
```

PR #849, opened 2023-12-27
feat: support qubits as subroutine args

ajberdy:
*Issue #, if available:*

#847 

*Description of changes:*

Support qubit-type arguments in subroutines

*Testing done:*

Tests to be added. Currently verified with the following manual test:

```
from braket.devices import LocalSimulator
import braket.experimental.autoqasm as aq
from braket.experimental.autoqasm.instructions import h, cnot, measure


@aq.subroutine
def bell(q0: aq.Qubit, q1: aq.Qubit):
    h(q0)
    cnot(q0, q1)


@aq.subroutine
def n_bells(n: int, q0: aq.Qubit, q1: aq.Qubit):
    for i in aq.range(n):
        h(q0)
        cnot(q0, q1)


# note that num_qubits is not needed here
@aq.main
def my_program():
    bell(0, 1)
    n_bells(5, 2, 3)
    measure(range(4))


my_program.display()
print(LocalSimulator().run(my_program, shots=10).result().measurements)

"""
OPENQASM 3.0;
def bell(qubit q0, qubit q1) {
    h q0;
    cnot q0, q1;
}
def n_bells(int[32] n, qubit q0, qubit q1) {
    for int i in [0:n - 1] {
        h q0;
        cnot q0, q1;
    }
}
qubit[4] __qubits__;
bell(__qubits__[0], __qubits__[1]);
n_bells(5, __qubits__[2], __qubits__[3]);
bit[4] __bit_0__ = "0000";
__bit_0__[0] = measure __qubits__[0];
__bit_0__[1] = measure __qubits__[1];
__bit_0__[2] = measure __qubits__[2];
__bit_0__[3] = measure __qubits__[3];

{'__bit_0__': ['0001', '0001', '1110', '1110', '1110', '0001', '1110', '1101', '0010', '1110']}
"""

```
## Merge Checklist

#### General

#### Tests

### DIFF
--- a/src/braket/experimental/autoqasm/api.py
+++ b/src/braket/experimental/autoqasm/api.py
@@ -39,9 +39,8 @@ from braket.experimental.autoqasm.autograph.impl.api_core import (
     is_autograph_artifact,
 )
 from braket.experimental.autoqasm.autograph.tf_utils import tf_decorator
-from braket.experimental.autoqasm.instructions.qubits import QubitIdentifierType as Qubit
-from braket.experimental.autoqasm.instructions.qubits import is_qubit_identifier_type
 from braket.experimental.autoqasm.program.gate_calibrations import GateCalibration
+from braket.experimental.autoqasm.types import QubitIdentifierType as Qubit
 from braket.parametric import FreeParameter
 
 
@@ -649,7 +648,7 @@ def _convert_calibration(
         value = union_deco_func_args[name]
         is_qubit = i in gate_args.qubit_indices
 
-        if is_qubit and not is_qubit_identifier_type(value):
+        if is_qubit and not aq_types.is_qubit_identifier_type(value):
             raise errors.ParameterTypeError(f'Parameter "{name}" must have a type of aq.Qubit.')
 
         if not is_qubit and not isinstance(value, (float, oqpy.AngleVar)):
--- a/src/braket/experimental/autoqasm/api.py
+++ b/src/braket/experimental/autoqasm/api.py
@@ -351,6 +351,20 @@ def _convert_subroutine(
                 _wrap_for_oqpy_subroutine(_clone_function(f), options)
             )
 
+            # todo: need to validate in other codepaths too?
+            # Validate and declare used qubits
+            quantum_indices = {
+                i
+                # todo: is there a better way to access argument types?
+                for i, arg in enumerate(oqpy_sub.subroutine_declaration[1].arguments)
+                if isinstance(arg, qasm_ast.QuantumArgument)
+            }
+            args = [
+                (aq_instructions.qubits._qubit(arg) if i in quantum_indices else arg)
+                for i, arg in enumerate(args)
+            ]
+            # todo: validate targets with program_conversion_context.validate_gate_targets?
+
             # Process the program
             subroutine_function_call = oqpy_sub(oqpy_program, *args, **kwargs)
             program_conversion_context.register_args(args)
--- a/src/braket/experimental/autoqasm/api.py
+++ b/src/braket/experimental/autoqasm/api.py
@@ -350,11 +350,9 @@ def _convert_subroutine(
                 _wrap_for_oqpy_subroutine(_clone_function(f), options)
             )
 
-            # todo: need to validate in other codepaths too?
             # Validate and declare used qubits
             quantum_indices = {
                 i
-                # todo: is there a better way to access argument types?
                 for i, arg in enumerate(oqpy_sub.subroutine_declaration[1].arguments)
                 if isinstance(arg, qasm_ast.QuantumArgument)
             }
@@ -362,7 +360,6 @@ def _convert_subroutine(
                 (aq_instructions.qubits._qubit(arg) if i in quantum_indices else arg)
                 for i, arg in enumerate(args)
             ]
-            # todo: validate targets with program_conversion_context.validate_gate_targets?
 
             # Process the program
             subroutine_function_call = oqpy_sub(oqpy_program, *args, **kwargs)
--- a/src/braket/experimental/autoqasm/api.py
+++ b/src/braket/experimental/autoqasm/api.py
@@ -353,8 +353,8 @@ def _convert_subroutine(
             # Validate and declare used qubits
             quantum_indices = {
                 i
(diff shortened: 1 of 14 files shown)


src/braket/experimental/autoqasm/types/conversions.py
```diff
@@ -37,6 +38,10 @@ def map_parameter_type(python_type: type) -> type:
     """
     origin_type = typing.get_origin(python_type) or python_type
 
+    # todo see if we can reuse `is_qubit_identifier_type` or at least
```
yitchen-tim, line 41: We should avoid circular import if we move `is_qubit_identifier_type` to `types` module. The method is more of a type method than a qubit method anyway. 

However, `is_qubit_identifier_type` checks the type of instantiated object. Here, I think we can use `QubitIdentifierType` directly with `python_type is QubitIdentifierType`.

src/braket/experimental/autoqasm/api.py
```diff
@@ -351,6 +351,20 @@ def _convert_subroutine(
                 _wrap_for_oqpy_subroutine(_clone_function(f), options)
             )
 
+            # todo: need to validate in other codepaths too?
+            # Validate and declare used qubits
+            quantum_indices = {
+                i
+                # todo: is there a better way to access argument types?
```
yitchen-tim, line 358: `aq.gate` and `aq.gate_calibration` use `aq_program.GateArgs()` to separate qubit and non-qubit argument. It only separates qubit and float types there. I think subroutines consider more than these two types. One idea is we could generalize `GateArgs` to include more types beyond just qubit and float.

src/braket/experimental/autoqasm/types/conversions.py
ajberdy: Done, works!

ajberdy:
Diff blew up a bit from the `QubitIdentifierType` move, but it's just a lot of refactoring of where to import from at the top of several files

src/braket/experimental/autoqasm/api.py
ajberdy: Yes, I see. That sounds like a good path to me

src/braket/experimental/autoqasm/types/conversions.py
yitchen-tim: thanks for refactoring!

src/braket/experimental/autoqasm/api.py
rmshaffer: One difference is that `GateArgs` allows arbitrary ordering of the qubit and angle args in Python, and then separates them out as required for OQ3 gate definitions. For subroutines we want the order in Python and the order in OQ3 to match exactly (right? can anyone think of a counterexample to this?).

src/braket/experimental/autoqasm/api.py
```diff
@@ -351,6 +350,20 @@ def _convert_subroutine(
...
+                for i, arg in enumerate(oqpy_sub.subroutine_declaration[1].arguments)
+                if isinstance(arg, qasm_ast.QuantumArgument)
```
rmshaffer, line 359: Is there a reason you need to get these from the `oqpy` subroutine declaration? What if you just look at the signature of the (Python) function returned from `_wrap_for_oqpy_subroutine`, or even just the signature of the (Python) function `f`?

src/braket/experimental/autoqasm/instructions/gates.py
```diff
@@ -20,8 +20,8 @@
 
 from braket.circuits.free_parameter_expression import FreeParameterExpression
 
+from ..types import QubitIdentifierType
```
rmshaffer, line 23: [nit] we've been using the full path for these, i.e. `braket.experimental.autoqasm.types` rather than `..types`.

src/braket/experimental/autoqasm/api.py
ajberdy: Yes, though here we do want to filter out only the qubit args.

src/braket/experimental/autoqasm/instructions/gates.py
ajberdy: I think pycharm matches whatever is used in the file and probably picked up on the relative import right below. I'll update both

src/braket/experimental/autoqasm/api.py
rmshaffer: Yeah, I agree, just trying to think through how this will work when we abstract `oqpy` out of this code. I imagine we will end up using the Python signatures here.

laurencap:
cool feature

src/braket/experimental/autoqasm/instructions/gates.py
ajberdy: Getting import errors when I switch this to absolute imports, maybe that's why the line below is using a relative import? Not sure that this is worth prioritizing before the Program refactor, which will likely alter file structure

src/braket/experimental/autoqasm/program/program.py
```diff
@@ -470,7 +470,6 @@ def validate_gate_targets(self, qubits: list[Any], angles: list[Any]) -> None:
             angles (list[Any]): The list of target angles to validate.
 
         Raises:
-            errors.InvalidTargetQubit: Target qubits are invalid in the current context.
```
rmshaffer, line 473: Did you mean to remove this? It looks like the call to `_validate_verbatim_target_qubits` could still raise this error.

src/braket/experimental/autoqasm/program/program.py
```diff
@@ -330,11 +330,11 @@ def register_gate(self, gate_name: str) -> None:
                     f"block. The native gates of the device are: {native_gates}"
                 )
 
-    def register_args(self, args: list[Any]) -> None:
+    def register_args(self, args: tuple[Any]) -> None:
```
rmshaffer, line 333: could this hint just be `Iterable`?

rmshaffer (commented):
Looks good overall, thanks for adding this feature! Left a few comments.

src/braket/experimental/autoqasm/program/program.py
ajberdy: Ah good call, I totally missed that lol

rmshaffer (approved):
Thanks for making the changes!

PR #850, opened 2023-12-29
feat: support noise model in dm simulators

yitchen-tim:
Issue #, if available:
closes https://github.com/amazon-braket/amazon-braket-sdk-python/issues/851

*Description of changes:*
- Support initializing DM simulator with a noise model, including local dm simulator and DM1. Noises are automatically applied to circuits when calling `device.run`.
- Validate supported pragma and throw error when adding noise model to devices that does not support noise. 

*Testing done:*
tox

## Merge Checklist

#### General

#### Tests

### DIFF
--- a/src/braket/aws/aws_device.py
+++ b/src/braket/aws/aws_device.py
@@ -114,9 +114,8 @@ class AwsDevice(Device):
         self._ports = None
         self._frames = None
         self._noise_model = noise_model
-
         if noise_model:
-            self._validate_noise_model_support()
+            self._validate_device_noise_model_support()
 
     def run(
         self,
@@ -202,8 +201,8 @@ class AwsDevice(Device):
         See Also:
             `braket.aws.aws_quantum_task.AwsQuantumTask.create()`
         """
-        if isinstance(task_specification, Circuit) and self._noise_model:
-            task_specification = self._noise_model.apply(task_specification)
+        if self._noise_model:
+            task_specification = self._apply_noise_model_to_circuit(task_specification)
         return AwsQuantumTask.create(
             self._aws_session,
             self._arn,
@@ -299,6 +298,11 @@ class AwsDevice(Device):
         See Also:
             `braket.aws.aws_quantum_task_batch.AwsQuantumTaskBatch`
         """
+        if self._noise_model:
+            task_specifications = [
+                self._apply_noise_model_to_circuit(task_specification)
+                for task_specification in task_specifications
+            ]
         return AwsQuantumTaskBatch(
             AwsSession.copy_session(self._aws_session, max_connections=max_connections),
             self._arn,
--- a/src/braket/aws/aws_device.py
+++ b/src/braket/aws/aws_device.py
@@ -33,6 +33,7 @@ from braket.aws.aws_session import AwsSession
 from braket.aws.queue_information import QueueDepthInfo, QueueType
 from braket.circuits import Circuit, Gate, QubitSet
 from braket.circuits.gate_calibrations import GateCalibrations
+from braket.circuits.noise_model import NoiseModel
 from braket.device_schema import DeviceCapabilities, ExecutionDay, GateModelQpuParadigmProperties
 from braket.device_schema.dwave import DwaveProviderProperties
 from braket.device_schema.pulse.pulse_device_action_properties_v1 import (  # noqa TODO: Remove device_action module once this is added to init in the schemas repo
@@ -78,11 +79,19 @@ class AwsDevice(Device):
         "Xy": "XY",
     }
 
-    def __init__(self, arn: str, aws_session: Optional[AwsSession] = None):
+    def __init__(
+        self,
+        arn: str,
+        aws_session: Optional[AwsSession] = None,
+        noise_model: Optional[NoiseModel] = None,
+    ):
         """
         Args:
             arn (str): The ARN of the device
             aws_session (Optional[AwsSession]): An AWS session object. Default is `None`.
+            noise_model (Optional[NoiseModel]): The Braket noise model to apply to the circuit
+                before execution. Noise model can only be added to the devices that support
+                noise simulation.
 
         Note:
             Some devices (QPUs) are physically located in specific AWS Regions. In some cases,
@@ -104,6 +113,10 @@ class AwsDevice(Device):
         self._aws_session = self._get_session_and_initialize(aws_session or AwsSession())
         self._ports = None
         self._frames = None
+        self._noise_model = noise_model
+
+        if noise_model:
+            self._validate_noise_model_support()
 
     def run(
         self,
@@ -189,6 +202,8 @@ class AwsDevice(Device):
         See Also:
             `braket.aws.aws_quantum_task.AwsQuantumTask.create()`
         """
+        if isinstance(task_specification, Circuit) and self._noise_model:
+            task_specification = self._noise_model.apply(task_specification)
         return AwsQuantumTask.create(
             self._aws_session,
             self._arn,
--- a/src/braket/aws/aws_device.py
+++ b/src/braket/aws/aws_device.py
@@ -113,9 +113,9 @@ class AwsDevice(Device):
         self._aws_session = self._get_session_and_initialize(aws_session or AwsSession())
         self._ports = None
         self._frames = None
-        self._noise_model = noise_model
         if noise_model:
-            self._validate_device_noise_model_support()
+            self._validate_device_noise_model_support(noise_model)
+        self._noise_model = noise_model
 
     def run(
         self,
--- a/src/braket/circuits/translations.py
+++ b/src/braket/circuits/translations.py
@@ -84,7 +84,7 @@ one_prob_noise_map = {
     "phase_damping": noises.PhaseDamping,
 }
 
-supported_noise_pragma_to_noise = {
+SUPPORTED_NOISE_PRAGMA_TO_NOISE = {
     "braket_noise_bit_flip": noises.BitFlip,
     "braket_noise_phase_flip": noises.PhaseFlip,
     "braket_noise_pauli_channel": noises.PauliChannel,
--- a/src/braket/circuits/translations.py
+++ b/src/braket/circuits/translations.py
@@ -84,6 +84,19 @@ one_prob_noise_map = {
     "phase_damping": noises.PhaseDamping,
 }
 
+supported_noise_pragma_to_noise = {
+    "braket_noise_bit_flip": noises.BitFlip,
+    "braket_noise_phase_flip": noises.PhaseFlip,
+    "braket_noise_pauli_channel": noises.PauliChannel,
+    "braket_noise_depolarizing": noises.Depolarizing,
+    "braket_noise_two_qubit_depolarizing": noises.TwoQubitDepolarizing,
+    "braket_noise_two_qubit_dephasing": noises.TwoQubitDephasing,
+    "braket_noise_amplitude_damping": noises.AmplitudeDamping,
+    "braket_noise_generalized_amplitude_damping": noises.GeneralizedAmplitudeDamping,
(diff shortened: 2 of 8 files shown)


speller26 (commented):
This is a really cool idea! Why not also add this to `AwsDevice` (checking if any noise pragmas are supported, of course) to allow us to attach noise models to DM1 too?

src/braket/devices/local_simulator.py
```diff
@@ -326,3 +341,19 @@ def _(
             )
         results = simulator.run(program, shots, *args, **kwargs)
         return AnalogHamiltonianSimulationQuantumTaskResult.from_object(results)
+
+    def _validate_noise_model_support(self) -> None:
```
krneta, line 345: String parsing for validation is always tricky. How much can we guarantee that the names will match?

yitchen-tim: It will work as long as the class name of the noise instruction matches the pragma name defined in `_to_openqasm()`.

yitchen-tim: I agree that string parsing is tricky. But there is also no other attribute to tell whether a device support noise or not. Alternatives that I have considered are
1. Do not validate. The noise model can be added to SV1 or QPU (which do not support noise). BDK does not throw error. The circuit with noise model applied is only validated at the service. 
2. Add a `openqasm_pragma_name` attribute to connect the name used in `_to_openqasm()` and the name in `supportedPragma`. This seems to make the Noise class clunkier.

yitchen-tim: @speller26 Do you have any thoughts on validation? Adding `openqasm_pragma_name` may remove string parsing. But it still need to match the pragma name in service.

yitchen-tim:
> This is a really cool idea! Why not also add this to AwsDevice (checking if any noise pragmas are supported, of course) to allow us to attach noise models to DM1 too?
Thanks! I wanted to iterate through the proposal first before working on the AwsDevice. Once we agree on the validation method, I'll move forward to AwsDevice.

src/braket/devices/local_simulator.py
krneta: This is only for local simulator, right? So noise model can't be added to SV1 or QPU?

speller26: Unfortunately, getting any information about supported operations relies on some form of string matching; I do like the idea of adding a [translation](https://github.com/amazon-braket/amazon-braket-sdk-python/blob/main/src/braket/circuits/translations.py) for pragma names.

As for device support, SV1 and QPUs don't apply, but noise models do make sense for DM1.

krneta: > As for device support, SV1 and QPUs don't apply, but noise models do make sense for DM1.

I meant this class only applies to local sim, so how we validate doesn't need to be tied to functionality of adding a noise model to things that don't support noise, or am I missing something?

yitchen-tim: I have updated the PR to 
1. Use the circuit.translation to match pragma names
2. Add support to AwsDevice

yitchen-tim:
`black` update causes format check to fail. The tests otherwise would pass.

src/braket/aws/aws_device.py
```diff
@@ -189,6 +202,8 @@ def run(
         See Also:
             `braket.aws.aws_quantum_task.AwsQuantumTask.create()`
         """
+        if isinstance(task_specification, Circuit) and self._noise_model:
+            task_specification = self._noise_model.apply(task_specification)
```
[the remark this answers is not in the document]
yitchen-tim: good idea. And also send warning when task_specification is not circuit where noise model does not apply

src/braket/aws/aws_device.py
```diff
@@ -78,11 +79,19 @@ class AwsDevice(Device):
         "Xy": "XY",
     }
 
-    def __init__(self, arn: str, aws_session: Optional[AwsSession] = None):
+    def __init__(
+        self,
+        arn: str,
+        aws_session: Optional[AwsSession] = None,
+        noise_model: Optional[NoiseModel] = None,
```
kshitijc, line 86: * How does this extend to the emulating the device? If we wanted to pass in a set of device capabilities, would that be a separate attribute?
* Did we consider alternatives like having a method instead (something like `.assume_device_profile()`), having a separate class etc?

yitchen-tim: I view the noise model as the interface for the device capabilities/properties. There are two user stories: (1) I want to emulate a device noise, and (2) I want to study how algorithm result decays with noise. For US#2 it's easier to use noise model directly. US#1 can be solved by implement a `NoiseModel.assume_device` method.

yitchen-tim:
The checks will pass after https://github.com/amazon-braket/amazon-braket-pennylane-plugin-python/pull/229 is merged.

src/braket/devices/local_simulator.py
```diff
@@ -326,3 +322,37 @@ def _(
             )
         results = simulator.run(program, shots, *args, **kwargs)
         return AnalogHamiltonianSimulationQuantumTaskResult.from_object(results)
+
+    def _prepare_batch_tasks_and_inputs(
```
yitchen-tim, line 326: Move this part out of run_batch because flake8 detects this logic as "too complex"

yitchen-tim: okay, thanks, I think flake8 flags is better too. This PR only adds two lines to this function. Moving code shows too many diff that is distracting.

src/braket/aws/aws_device.py
speller26: Agree with Tim; the way I see it, both are good to have: a way to easily emulate a device, as well as to attach an arbitrary noise model to emulate a hypothetical device.

src/braket/devices/device.py
```diff
@@ -104,3 +110,43 @@ def status(self) -> str:
...
+        noise_pragma_name_to_class = copy(one_prob_noise_map)
+        noise_pragma_name_to_class.update({"kraus": Noise.Kraus})
+
+        supported_noises = [
+            noise_pragma_name_to_class[pragma.split("braket_noise_")[-1]]
+            for pragma in (self.properties.action[DeviceActionType.OPENQASM].supportedPragmas)
+            if "braket_noise_" in pragma
+        ]
+
+        noise_operators = [noise_instr.noise for noise_instr in self._noise_model._instructions]
```
speller26, line 124: How about instead of using `supportedPragmas`, using `supportedOperations`? The names are normalized there. Then you can do something like
```python
supported_noise = set(action.supportedOperations).intersection(set(noise_pragma_name_to_class.keys()))
```

src/braket/devices/device.py
```diff
@@ -104,3 +110,43 @@ def status(self) -> str:
...
+        if not all([isinstance(noise, tuple(supported_noises)) for noise in noise_operators]):
```
speller26, line 125: Nor a big deal, I'll say it anyway: instead of iterating over the supported noises for each noise instruction, how about adding the name of each noise instruction to a set, and then raising the exception if the set difference between this set and the supported noise set is nonempty?

speller26:
> The checks will pass after [amazon-braket/amazon-braket-pennylane-plugin-python#229](https://github.com/amazon-braket/amazon-braket-pennylane-plugin-python/pull/229) is merged.

This is really weird; the test is failing due to the `_noise_model` attribute not existing, but yet is able to reach a branch that uses it. This seems to be more of a problem with the the SDK's test infrastructure than the PL plugin tests.

src/braket/devices/local_simulator.py
yitchen-tim: Added the flag to remove the need of this private method. This reduces the diff that is not relevant to this PR.

src/braket/devices/device.py
yitchen-tim: I've considered that but the noise is in supportedOperations for local simulator, but in supportedPragmas in AwsDevice.

src/braket/devices/device.py
yitchen-tim: Thanks for the suggestion! That would be ideal if a device object has a native attribute of "supported noises". Currently it doesn't. To do it in the way you described, I would need to a way to distinguish noise operations/pragmas from other operations/pragmas. To achieve this, I might need string parsing or add more attributes about noise to the device class.

src/braket/aws/aws_device.py
```diff
@@ -104,6 +113,9 @@ def __init__(self, arn: str, aws_session: Optional[AwsSession] = None):
         self._aws_session = self._get_session_and_initialize(aws_session or AwsSession())
         self._ports = None
         self._frames = None
+        self._noise_model = noise_model
+        if noise_model:
+            self._validate_device_noise_model_support()
```
speller26, line 118: nit: reverse
```suggestion
        if noise_model:
            self._validate_device_noise_model_support()
        self._noise_model = noise_model
```

src/braket/devices/device.py
```diff
@@ -104,3 +110,43 @@ def status(self) -> str:
...
+            if any(
+                [
+                    isinstance(instruction.operator, Noise)
+                    for instruction in task_specification.instructions
+                ]
+            ):
+                warnings.warn(
+                    "The noise model of the device is applied to a circuit that already has noise"
+                    " instructions."
+                )
```
speller26, line 144: No need to go through all instructions in the circuit; you can stop searching the moment you find a noise instruction.

src/braket/devices/device.py
speller26: Oof, in that case, as verbose as it is, just create the map from the properties pragma name "braket_pragma_..." to the actual name. It's ugly, but at least it's explicit instead of relying on string parsing.

src/braket/devices/device.py
speller26: You don't have to use a supported noise attribute; you can build the set directly from the `supported_noises` variable that you already have. Something like this:
```suggestion
        supported_noise_set = set(supported_noise)
        noise_model_operator_set = {noise.name for noise in noise_operators}
        if noise_model_operator_set - supported_noise_set:
```

src/braket/devices/device.py
yitchen-tim: I see you you meant now. I added translation from pragma to noise class. Even though the validation only compares the sets of names. I set translation to map to noise class because I think it's more robust than mapping to names.

speller26:
Looks like we just need some tests to complete coverage

src/braket/circuits/translations.py
```diff
@@ -84,6 +84,19 @@
     "phase_damping": noises.PhaseDamping,
 }
 
+supported_noise_pragma_to_noise = {
```
speller26, line 87: nit: Since this is a constant, make it ALL_CAPS

src/braket/devices/device.py
```diff
@@ -104,3 +110,36 @@ def status(self) -> str:
...
         return self._status
+
+    def _validate_device_noise_model_support(self, noise_model: NoiseModel) -> None:
+        supported_noises = set(
+            supported_noise_pragma_to_noise[pragma].__name__
+            for pragma in self.properties.action[DeviceActionType.OPENQASM].supportedPragmas
+            if pragma in supported_noise_pragma_to_noise
+        )
+        noise_operators = set(noise_instr.noise.name for noise_instr in noise_model._instructions)
+        if noise_operators > supported_noises:
```
speller26, line 121: Rather than doing a `>` comparison, you should explicitly check whether there is anything in `noise_operators` that isn't in `supported_noises`; it's conceivable that `noise_operators` contains something not in `supported_noises` while not being a strict superset:
```python
# This would pass, when it should actually faily
noise_operators = {1, 2, 3}
supported_noises = {2, 3, 4, 5}
```

speller26 (approved):

PR #855, opened 2024-01-19
feat: support Hybrid Job decorator with AutoQASM

yitchen-tim:
*Issue #, if available:*
closes https://github.com/amazon-braket/autoqasm/issues/6

*Description of changes:*
Draft PR for testing the proposed solution: I made the following change to enable the job decorator + AutoQASM. Still need to add new tests and adapt existing tests.
- Persist source code of all inner functions of the job decorated function, as py files
- Upload the py file to the job as input data
- Replace the source location of all inner functions to a path in the job
This change strictly only applies to remote job decorator. When using regular job (ie. without decorator) or no job, there is no change in user code, and there is no artifact about source code. 

working example:
```
import json
from braket.jobs import hybrid_job
from braket.devices import Devices
import braket.experimental.autoqasm as aq
from braket.experimental.autoqasm import instructions as ops

@hybrid_job(
    device=Devices.Amazon.SV1,
    dependencies=["git+https://github.com/amazon-braket/amazon-braket-sdk-python.git@feature/autoqasm#egg=amazon-braket-sdk"],
    local=True
) 
def bell_circuit_job():
    @aq.main
    def bell_circuit():
        ops.h(0)
        ops.cnot(0, 1)

    print(bell_circuit.to_ir())

bell_circuit_job()
```

*Testing done:*
More complete tests are needed. I tested few examples of `aq.main`, `aq.subroutine`, `aq.gate`, `aq.gate_calibration` and they all work. Also tested with non-decorator UX, also worked. 

## Merge Checklist

#### General

#### Tests


src/braket/experimental/autoqasm/autograph/pyct/parser.py
```diff
@@ -128,6 +130,21 @@ def dedent_block(code_string):
   return new_code
 
 
+def in_hybrid_job():
+  if "AMZN_BRAKET_INPUT_DIR" in os.environ:
```
rmshaffer, line 134: minor comment: since this is in `autograph` source, we'll have to eventually contribute this to the standalone `autograph` repo when it becomes available. In that case, we might want to consider one more layer of abstraction here so that we can keep "Braket" and "hybrid job" concepts out of `autograph` code.

src/braket/jobs/hybrid_job.py
```diff
@@ -170,6 +171,8 @@ def job_wrapper(*args, **kwargs) -> Callable:
             Returns:
                 Callable: the callable for creating a Hybrid Job.
             """
+            inner_source_temp_dir, inner_source_file_path = persist_inner_function_source(entry_point)
```
rmshaffer, line 174: curious, why do you use a separate temp dir for this file? could you just use the `temp_dir` created in the `with` statement below? (side note, the `with tempfile.TemporaryDirectory()` pattern automatically deletes the dir when leaving the scope, so you wouldn't have to explicitly call `rmtree`)

src/braket/jobs/hybrid_job.py
```diff
@@ -217,13 +221,49 @@ def job_wrapper(*args, **kwargs) -> Callable:
...
+    inside the job decorated function.
+
+    Args:
+        outer_function (callable): The outer function.
+
+    Returns:
+        dict[str, str]: Mapping between name and source code of each inner functions.
+    """
+    inner_function_source = {}
+    for const in outer_function.__code__.co_consts:
```
rmshaffer, line 261: 1. Does this work recursively? e.g. what if my job-decorated function calls another function, and the `@aq.main` is defined there?
2. Does this work for `@aq.main`-decorated functions that are defined outside the scope of the job-decorated function (or imported from another module)?

src/braket/experimental/autoqasm/autograph/pyct/parser.py
```diff
@@ -145,7 +162,10 @@ def parse_entity(entity, future_features):
     return _parse_lambda(entity)
 
   try:
-    original_source = inspect_utils.getimmediatesource(entity)
+    if in_hybrid_job():
+      original_source = get_source_from_job(entity)
+    else:  
+      original_source = inspect_utils.getimmediatesource(entity)
```
rmshaffer, line 168: This is very easy to understand - nice place to fork the behavior!

src/braket/jobs/hybrid_job.py
yitchen-tim: Both do not work.
For (1), I think if `my_func` is in  job-decorated function, there is a way to work. 
For (2), function defined outside is tricky. Need to play with it to see.

src/braket/jobs/hybrid_job.py
yitchen-tim: Actually, there is already two context manager in that code block, I think I can just add the third one for `inner_source_file_path`. I'll update it.

src/braket/experimental/autoqasm/autograph/pyct/parser.py
yitchen-tim: It also makes it easier to maintain and easier to write unit tests when source code is injected outside of autograph.

yitchen-tim: The source code injecting is now separated from autograph!

src/braket/jobs/hybrid_job.py
yitchen-tim: Recursive function is now supported. The solution for function defined outside of job decorator remains elusive. It likely requires a remake of how job decorator works, removing dependency of cloudpickle.

rmshaffer (approved):
I like how general this solution is! I left a handful of very minor comments. It would also be great to get @ajberdy's review.

rmshaffer:
As we discussed offline with @yitchen-tim, we should modify this PR to create a new AutoQASM-specific decorator like `@aq.hybrid_job` which implements this functionality. This will eliminate any impact to the behavior of the core SDK code, and will also make it simpler to decouple AutoQASM when we are ready to move it to a standalone repo.

rmshaffer:
Closing this PR, since AutoQASM code has been moved to a separate repo. The fix from this PR can be largely ported over to the new repo (but see my comment here: https://github.com/amazon-braket/amazon-braket-sdk-python/pull/855#issuecomment-2077834673).

Now being tracked by this issue in the new repo: https://github.com/amazon-braket/autoqasm/issues/6

PR #857, opened 2024-01-23
feat: update S3 locations for jobs

ajberdy:
*Issue #, if available:*

*Description of changes:*

Add a timestamp to the S3 path of various job locations, such as source, input/output, checkpoints.

Should make sure to update defaults in console along with this PR

*Testing done:*

Unit tests and jobs integ tests, plus manual verification

## Merge Checklist

#### General

#### Tests

### DIFF
--- a/src/braket/jobs/quantum_job_creation.py
+++ b/src/braket/jobs/quantum_job_creation.py
@@ -405,7 +405,7 @@ def _process_input_data(
     input_data: str | dict | S3DataSourceConfig,
     job_name: str,
     aws_session: AwsSession,
-    timestamp: str,
+    subdirectory: str,
 ) -> list[dict[str, Any]]:
     """
     Convert input data into a list of dicts compatible with the Braket API.
@@ -415,7 +415,7 @@ def _process_input_data(
             can be an S3DataSourceConfig or a str corresponding to a local prefix or S3 prefix.
         job_name (str): Hybrid job name.
         aws_session (AwsSession): AwsSession for possibly uploading local data.
-        timestamp (str): Timestamp for job.
+        subdirectory (str): Subdirectory within job name for S3 locations.
 
     Returns:
         list[dict[str, Any]]: A list of channel configs.
@@ -425,7 +425,7 @@ def _process_input_data(
     for channel_name, data in input_data.items():
         if not isinstance(data, S3DataSourceConfig):
             input_data[channel_name] = _process_channel(
-                data, job_name, aws_session, channel_name, timestamp
+                data, job_name, aws_session, channel_name, subdirectory
             )
     return _convert_input_to_config(input_data)
 
@@ -435,7 +435,7 @@ def _process_channel(
     job_name: str,
     aws_session: AwsSession,
     channel_name: str,
-    timestamp: str,
+    subdirectory: str,
 ) -> S3DataSourceConfig:
     """
     Convert a location to an S3DataSourceConfig, uploading local data to S3, if necessary.
@@ -444,7 +444,7 @@ def _process_channel(
         job_name (str): Hybrid job name.
         aws_session (AwsSession): AwsSession to be used for uploading local data.
         channel_name (str): Name of the channel.
-        timestamp (str): Timestamp for job.
+        subdirectory (str): Subdirectory within job name for S3 locations.
 
     Returns:
         S3DataSourceConfig: S3DataSourceConfig for the channel.
@@ -459,7 +459,7 @@ def _process_channel(
             aws_session.default_bucket(),
             "jobs",
             job_name,
-            timestamp,
+            subdirectory,
             "data",
             channel_name,
             location_name,
--- a/src/braket/jobs/quantum_job_creation.py
+++ b/src/braket/jobs/quantum_job_creation.py
@@ -453,7 +453,7 @@ def _process_channel(
         return S3DataSourceConfig(location)
     else:
         # local prefix "path/to/prefix" will be mapped to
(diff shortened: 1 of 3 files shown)


src/braket/jobs/quantum_job_creation.py
```diff
@@ -251,19 +255,22 @@ def prepare_quantum_job(
     return create_job_kwargs
 
 
-def _generate_default_job_name(image_uri: str | None = None, func: Callable | None = None) -> str:
+def _generate_default_job_name(
+    image_uri: str | None = None, func: Callable | None = None, timestamp: int | str | None = None
```
krneta, line 259: nit: consider calling it "subdirectory" or "file_path" or something similar so that we can change from timestamp to something else (if necessary) without refactoring. It would also make it more apparent that this value probably shouldn't be an int (this goes for other function declarations as well in this PR as well).

src/braket/jobs/quantum_job_creation.py
```diff
@@ -161,14 +161,15 @@ def prepare_quantum_job(
     _validate_params(param_datatype_map)
     aws_session = aws_session or AwsSession()
     device_config = DeviceConfig(device)
-    job_name = job_name or _generate_default_job_name(image_uri=image_uri)
+    timestamp = str(int(time.time() * 1000))
```
krneta, line 164: Have we finalized if we want the actual timestamp, or a formatted date/time that the customer would be able to read more easily?

src/braket/jobs/quantum_job_creation.py
ajberdy: That's a great call. I'll update

src/braket/jobs/quantum_job_creation.py
ajberdy: Haven't confirmed. Personally I find UTC time not that much more helpful than a timestamp, but happy to to pivot on this. Easy one-line change in the implementation

src/braket/jobs/quantum_job_creation.py
ajberdy: Note that this line in specific is related to the job name, where it *is* a timestamp, separate from the subdirectory we are adding. We don't actually need this kwarg if we are okay with having the two timestamps in `.../job-name-ts1/ts2/...` be different. Thoughts?

src/braket/jobs/quantum_job_creation.py
krneta: Yes, but I'd rather we settle on it as part of this discussion than kick the conversation down the road. I think it should be a fairly quick decision to make.

src/braket/jobs/quantum_job_creation.py
krneta: I think that having ../job-name-ts1/ts1/ is also silly. I thought we were getting rid of the first ts1?

src/braket/jobs/quantum_job_creation.py
```diff
@@ -435,10 +453,16 @@ def _process_channel(
         return S3DataSourceConfig(location)
     else:
         # local prefix "path/to/prefix" will be mapped to
-        # s3://bucket/jobs/job-name/data/input/prefix
+        # s3://bucket/jobs/job-name/timestamp/data/input/prefix
```
krneta, line 456: subdirectory

test/integ_tests/test_create_quantum_job.py
```diff
@@ -58,14 +58,16 @@ def test_failed_quantum_job(aws_session, capsys):
     # Check job is in failed state.
     assert job.state() == "FAILED"
 
+    timestamp = job.name.split("-")[-1]
```
krneta, line 61: How will you get ts2, if we change that here?

test/integ_tests/test_create_quantum_job.py
```diff
@@ -114,27 +116,36 @@ def test_completed_quantum_job(aws_session, capsys):
...
+
     # Check whether the respective folder with files are created for script,
     # output, tasks and checkpoints.
     job_name = job.name
     s3_bucket = f"amazon-braket-{aws_session.region}-{aws_session.account_id}"
     keys = aws_session.list_keys(
         bucket=s3_bucket,
-        prefix=f"jobs/{job_name}",
+        prefix=f"jobs/{job_name}/{timestamp}/",
```
krneta, line 127: rename these timestamps to subdirectory as well?

src/braket/jobs/quantum_job_creation.py
ajberdy: Yes, we will get rid of the first ts1 as long as the uniqueness constraint for job names is dropped

test/integ_tests/test_create_quantum_job.py
ajberdy: Ah fair enough. In that case, I'll just list keys for f"jobs/{job_name}/" and match against `\d+` for the timestamp in the asserts. Might as well make that change now instead of later

test/integ_tests/test_create_quantum_job.py
ajberdy: As per above comment, will remove the timestamp variable here

test/integ_tests/test_create_quantum_job.py
ajberdy: Actually, that seems to lower the test coverage quite a bit since we could be validating assets from old jobs. I should be able to parse from metadata instead

ajberdy:
Getting a hash error when installing packages for dependencies tests. Doesn't appear to be related to the code change, may be a transient error

krneta (approved):

PR #873, opened 2024-02-12
feature: add partial measurement method to circuit class

ashlhans:
*Description of changes:*
Added methods to return the measurements and measurement counts of a specified list of qubits.

Example:
```
bell = Circuit().h(0).cnot(0, 1)
task = device.run(bell, shots=100)
result = task.result()

print(result.partial_measurements([0, 2])) 

[[1 1] 
 [1 0] 
 [1 1] 
 [0 1] 
 [1 0]]

print(result.partial_measurement_counts([0]))

Counter({'1': 5, '0': 5})
```

*Testing done:*
tox
## Merge Checklist

#### General

#### Tests

### DIFF
--- a/src/braket/tasks/gate_model_quantum_task_result.py
+++ b/src/braket/tasks/gate_model_quantum_task_result.py
@@ -153,13 +153,13 @@ class GateModelQuantumTaskResult:
 
     def partial_measurements(self, qubits: Iterable[int]) -> np.ndarray:
         """
-        Gets the partial measurements from the qubits given
+        Gets a filtered list of partial measurements over the qubits provided.
 
         Args:
             qubits (Iterable [int]): an array of qubits
 
         Returns:
-            ndarray: an array of the measurements of the given qubits
+            ndarray: a list of the measurements of the given qubits
         """
         return GateModelQuantumTaskResult._selected_measurements(
             self.measurements, range(len(self.measurements)), qubits
@@ -167,13 +167,13 @@ class GateModelQuantumTaskResult:
 
     def partial_measurement_counts(self, qubits: Iterable[int]) -> np.ndarray:
         """
-        Gets the counts of the partial measurements from the qubits given
+        Gets the counts of the filtered list of measurements over the qubits provided.
 
         Args:
             qubits (Iterable [int]): an array of qubits
 
         Returns:
-            ndarray: an array of the measurement counts of the given qubits
+            ndarray: a list of the measurement counts of the given qubits
         """
         return GateModelQuantumTaskResult.measurement_counts_from_measurements(
             GateModelQuantumTaskResult._selected_measurements(
@@ -452,7 +452,7 @@ class GateModelQuantumTaskResult:
     def _selected_measurements(
         measurements: np.ndarray, measured_qubits: list[int], targets: Optional[list[int]]
     ) -> np.ndarray:
-        if targets is not None and targets != measured_qubits:
+        if targets is not None and not np.array_equal(targets, measured_qubits):
             # Only some qubits targeted
             columns = [measured_qubits.index(t) for t in targets]
             measurements = measurements[:, columns]
--- a/src/braket/tasks/gate_model_quantum_task_result.py
+++ b/src/braket/tasks/gate_model_quantum_task_result.py
@@ -17,7 +17,7 @@ import json
 from collections import Counter
 from collections.abc import Callable
 from dataclasses import dataclass
-from typing import Any, Optional, TypeVar, Union
+from typing import Any, Iterable, Optional, TypeVar, Union
 
 import numpy as np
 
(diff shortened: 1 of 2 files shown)


src/braket/tasks/gate_model_quantum_task_result.py
```diff
@@ -151,6 +151,36 @@ def get_compiled_circuit(self) -> Optional[str]:
...
+    def partial_measurements(self, qubits: Iterable[int]) -> np.ndarray:
+        """
+        Gets the partial measurements from the qubits given
+
+        Args:
+            qubits (Iterable [int]): an array of qubits
+
+        Returns:
+            ndarray: an array of the measurements of the given qubits
+        """
```
ajberdy, line 163: Let's make sure to clarify that this is a post-processing filter, rather than a hardware partial measurement. Maybe something like "Gets a filtered list of partial measurements over the qubits provided."

and nit: I would opt for "list" over "array" in a Python context

src/braket/tasks/gate_model_quantum_task_result.py
```diff
@@ -151,6 +151,36 @@ def get_compiled_circuit(self) -> Optional[str]:
...
+    def partial_measurement_counts(self, qubits: Iterable[int]) -> np.ndarray:
+        """
+        Gets the counts of the partial measurements from the qubits given
+
+        Args:
+            qubits (Iterable [int]): an array of qubits
+
+        Returns:
+            ndarray: an array of the measurement counts of the given qubits
+        """
```
ajberdy, line 177: Same comments as above

ajberdy:
Nice! Can we add some tests?

src/braket/tasks/gate_model_quantum_task_result.py
kshitijc: Sorry, I'm confused. 
* Why are we introducing a post-processing filter? Don't we already have a way to do this using the `sample` result type? 
* I'd have expected a `measure` instruction to be introduced instead. What prevents us from doing that?

src/braket/tasks/gate_model_quantum_task_result.py
```diff
@@ -151,6 +151,36 @@ def get_compiled_circuit(self) -> Optional[str]:
         else:
             return None
 
+    def partial_measurements(self, qubits: Iterable[int]) -> np.ndarray:
```
rmshaffer, line 154: General comment: I don't think "partial measurement" is a term we would want to expose in the API. Just "measure", along with the list of qubits to measure.

src/braket/tasks/gate_model_quantum_task_result.py
ajberdy: Correct. That is still the plan, but will require service support. This is meant to be a convenience method for the result type class. We can rename/hold off if you think it will cause confusion.

Sample result type does give us the same information, but the results are given as [-1, +1] along a given observable, rather than the the familiar 0/1.

src/braket/tasks/gate_model_quantum_task_result.py
ajberdy: Got it. So do you think it would be better to add an optional filter kwarg (`qubits` maybe?) to the existing functions, instead of introducing new ones?

src/braket/tasks/gate_model_quantum_task_result.py
kshitijc: I believe service support is already present. You should be able to run a circuit which measures a subset of the qubits:
```
oq = """
OPENQASM 3.0;
bit[1] b;
qubit[2] q;
h q[0];
cnot q[0], q[1];
b[0] = measure q[0];
"""

from braket.ir.openqasm import Program as OpenQASMProgram
p = OpenQASMProgram(source=oq)

oqc = AwsDevice("arn:aws:braket:eu-west-2::device/qpu/oqc/Lucy")
oqc.run(p)
```


The device property `requiresAllQubitsMeasurement` specifies whether you can measure a subset of the qubits or not.

speller26:
The PR title (and the commit) should be about extracting partial measurements from results rather than adding partial measurements to circuits, which would look more like
```python
circ = Circuit().h(0).cnot(0, 1).measure([1])
```

src/braket/tasks/gate_model_quantum_task_result.py
rmshaffer: That would make more sense to me, yes. But I agree with Kshitij that a measure instruction in the circuit itself is likely to be a more useful way to expose this. The post-processing step doesn't do much that the user can't already do today fairly easily (in fact, these new methods have one-line implementations).

ashlhans:
We are not moving forward with this PR at the moment. Another PR for partial measurement can be found [here](https://github.com/amazon-braket/amazon-braket-sdk-python/pull/904). A PR for improved post-processing with measurements can be found [here](https://github.com/amazon-braket/amazon-braket-sdk-python/pull/894).

PR #887, opened 2024-02-21
feat: update log stream prefix for new jobs

ajberdy:
*Issue #, if available:*

*Description of changes:*

Search for new logs prefix `job-name/uuid` when interacting with jobs with the new ARN format

*Testing done:*

Regression testing and added a unit test. Will run integ tests once new ARN format is deployed

## Merge Checklist

#### General

#### Tests

### DIFF
--- a/src/braket/aws/aws_quantum_job.py
+++ b/src/braket/aws/aws_quantum_job.py
@@ -292,9 +292,9 @@ class AwsQuantumJob(QuantumJob):
         # jobs ARNs used to contain the job name and use a log prefix of `job-name/`
         # now job ARNs use a UUID and a log prefix of `job-name/UUID/`
         return (
-            f"{self.name}/"
+            f"{self.name}"
             if self.arn.endswith(self.name)
-            else f"{self.name}/{self.arn.split('/')[-1]}/"
+            else f"{self.name}/{self.arn.split('/')[-1]}"
         )
 
     def state(self, use_cached_value: bool = False) -> str:
@@ -472,7 +472,7 @@ class AwsQuantumJob(QuantumJob):
         if self.state() in AwsQuantumJob.TERMINAL_STATES and "endedAt" in metadata:
             job_end = int(math.ceil(metadata["endedAt"].timestamp()))
         return fetcher.get_metrics_for_job(
-            self.name, metric_type, statistic, job_start, job_end, self.arn
+            self.name, metric_type, statistic, job_start, job_end, self._logs_prefix
         )
 
     def cancel(self) -> str:
--- a/src/braket/aws/aws_quantum_job.py
+++ b/src/braket/aws/aws_quantum_job.py
@@ -289,6 +289,8 @@ class AwsQuantumJob(QuantumJob):
     @property
     def _logs_prefix(self) -> str:
         """str: the prefix for the job logs."""
+        # jobs ARNs used to contain the job name and use a log prefix of `job-name/`
+        # now job ARNs use a UUID and a log prefix of `job-name/UUID/`
         return (
             f"{self.name}/"
             if self.arn.endswith(self.name)
--- a/src/braket/aws/aws_quantum_job.py
+++ b/src/braket/aws/aws_quantum_job.py
@@ -289,8 +289,8 @@ class AwsQuantumJob(QuantumJob):
     @property
     def _logs_prefix(self) -> str:
         """str: the prefix for the job logs."""
-        # jobs ARNs used to contain the job name and use a log prefix of `job-name/`
-        # now job ARNs use a UUID and a log prefix of `job-name/UUID/`
+        # jobs ARNs used to contain the job name and use a log prefix of `job-name`
+        # now job ARNs use a UUID and a log prefix of `job-name/UUID`
         return (
             f"{self.name}"
             if self.arn.endswith(self.name)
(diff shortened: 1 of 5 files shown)


src/braket/aws/aws_quantum_job.py
```diff
@@ -286,6 +286,15 @@ def name(self) -> str:
         """str: The name of the quantum job."""
         return self.metadata(use_cached_value=True).get("jobName")
 
+    @property
+    def _logs_prefix(self) -> str:
```
krneta, line 290: For my information, is it typical to create properties for private functions? Why not just use the private function instead?

src/braket/aws/aws_quantum_job.py
```diff
@@ -286,6 +286,15 @@ def name(self) -> str:
...
+        """str: the prefix for the job logs."""
+        return (
+            f"{self.name}/"
+            if self.arn.endswith(self.name)
```
krneta, line 294: When can we get rid of this? Should we leave a comment (that wouldn't go into our API docs) explaining why this is here, and when it would be safe to remove?

src/braket/jobs/metrics_data/cwl_insights_metrics_fetcher.py
```diff
@@ -166,11 +169,15 @@ def get_metrics_for_job(
         query_end_time = job_end_time or int(time.time())
         query_start_time = job_start_time or query_end_time - self.QUERY_DEFAULT_JOB_DURATION
 
-        # The hybrid job name needs to be unique to prevent jobs with similar names from being
-        # conflated.
+        stream_prefix = (
```
krneta, line 172: It kind of sucks that we have to duplicate this logic. Should we just pass in the stream prefix?

test/unit_tests/braket/aws/test_aws_quantum_job.py
```diff
@@ -1104,3 +1104,13 @@ def test_bad_device_arn_format(aws_session):
 
     with pytest.raises(ValueError, match=device_not_found):
         AwsQuantumJob._initialize_session(aws_session, "bad-arn-format", logger)
+
+
+def test_logs_prefix(quantum_job_arn, quantum_job_name, aws_session, generate_get_job_response):
+    quantum_job = AwsQuantumJob(quantum_job_arn, aws_session)
+    aws_session.get_job.return_value = generate_get_job_response(jobName=quantum_job_name)
+    assert quantum_job._logs_prefix == f"{quantum_job_name}/"
+    uuid = "UUID-123456789"
```
krneta, line 1113: Maybe add some comments explaining the test? I also think it's important that we test that two jobs with the same name (but different arns) would have different log locations.

src/braket/aws/aws_quantum_job.py
ajberdy: We can't get rid of this, unless we have a limited retention period for old jobs, since the associated log streams for historic jobs will not be updated.

src/braket/jobs/metrics_data/cwl_insights_metrics_fetcher.py
ajberdy: Perhaps, but this is technically a public interface, so I wanted to avoid breaking it

test/unit_tests/braket/aws/test_aws_quantum_job.py
ajberdy: Sure, I'll add some comments. Though if the log location includes the UUID, then it's a precise conclusion that jobs with distinct UUIDs will have distinct log locations; I'm not sure that we would get any value of doing this test twice with different UUIDs and showing that they're not equal. Happy to include it regardless if you disagree

src/braket/aws/aws_quantum_job.py
krneta: It's up to you

src/braket/aws/aws_quantum_job.py
krneta: Hummm... I would hope that after a while (eg. a year of jobs being created using ARN), we'd be able to remove the logic, no?

src/braket/jobs/metrics_data/cwl_insights_metrics_fetcher.py
krneta: I don't understand - if we're not breaking it by adding arn, why would we be breaking it by adding the stream prefix?

test/unit_tests/braket/aws/test_aws_quantum_job.py
krneta: While it's implicitly true, the invariant that we actually care about is that logs with the same job name have different log paths. Generally, we should be testing the invariants so that they remain true even if there are different reasons the code changes the log locations.

I don't feel strongly about it, so it's fine if you leave it as-is.

src/braket/jobs/metrics_data/cwl_insights_metrics_fetcher.py
ajberdy: Oh I misunderstood. I thought you meant replace the `job_name` parameter with `stream_prefix`, which would be the most direct way of updating the code, but would be a breaking change. We could add an optional `stream_prefix` parameter at the end instead of arn to avoid duplicating logic. I like that

src/braket/aws/aws_quantum_job.py
ajberdy: That would just mean we're breaking `job.logs()` for `job`s a year old. Perhaps the simplification is worth that regression if we feel the blast radius is small

krneta (approved):

PR #892, opened 2024-02-22
documentation: Update example notebooks to use AutoQASM return statements

### LINKED ISSUE #790: FreeParameter documentation updates

laurencap:
*Issue #, if available:* Closes #790

*Description of changes:*
Update example notebooks and change default return value name from `"retval_"` to `"return_value"`

*Testing done:*

## Merge Checklist

#### General

#### Tests


examples/autoqasm/4_Native_programming.ipynb
```diff
@@ -46,12 +46,14 @@
      "output_type": "stream",
      "text": [
       "OPENQASM 3.0;\n",
+      "output bit[2] retval_;\n",
```
rmshaffer, line 49: I see `retval_` in the output cells for this notebook - does it need to be re-run with the latest changes?

examples/autoqasm/1_Getting_started_with_AutoQASM.ipynb
```diff
@@ -108,7 +108,7 @@
    "source": [
     "device = LocalSimulator()\n",
     "result = device.run(bell_state, shots=100).result()\n",
-    "counts = Counter(result.measurements[\"c\"])\n",
+    "counts = Counter(result.measurements[\"return_value\"])\n",
```
rmshaffer, line 111: Once we implement outputs in the result object, will this help the user avoid having to use magic string literals like `"return_value"` (or `"return_value0"`, etc. in the case of tuples)?

src/braket/experimental/autoqasm/constants.py
```diff
@@ -35,3 +35,6 @@
 
 RETVAL_VARIABLE_NAME = "retval_"
 """A special name for variables assigned to the return values of function calls."""
+
+MAIN_RETURN_VAL_NAME = "return_value"
+"""The default name for the value returned from autoqasm.main functions."""
```
rmshaffer, line 40: What happens if the user names one of their variables `"return_value"`?

rmshaffer (approved):

examples/autoqasm/1_Getting_started_with_AutoQASM.ipynb
laurencap: Yes, exactly. That's the piece that we can't do without service changes. I can hack in something that allows us to use `result.output` instead of `result.measurements["return_value"]`, but it breaks down when the user returns a named variable, like we have in the IPE demo. I think it's probably best to wait until we can do the complete solution, rather than a hack, but that could be a discussion question for Monday

examples/autoqasm/4_Native_programming.ipynb
laurencap: Ahh, I thought I'd ran them again, thank you for catching this

src/braket/experimental/autoqasm/constants.py
laurencap: Ah yes, good point.

If there's no return statement, no problem. If the local variable is itself returned, with or without other values, no problem. If it's the same type as the returned value, it just gets assigned at the end to the returned value, no problem.

If the input is (strangely) named `return_value`, then we get `NameConflict: Your output parameter has the same name as an input parameter: 'return_value'. Please give them unique names.`. If there's a local var with the same name and different type, we get `RuntimeError: Program has conflicting variables with name return_value`.

Similar to my POV from the other discussion thread, I find it preferable to wait to fix it the correct way through the service. I don't think any of these behaviors are too annoying -- yeah, it's unexpected that it would cause failures in some cases, but there's an error message that tells you how to fix it.

laurencap: Whether or not `return_value` is actually better than `retval_`, though, I don't know.

examples/autoqasm/1_Getting_started_with_AutoQASM.ipynb
laurencap: For multiple returned values, it would be `outputs[0]`, `outputs[1]`

PR #904, opened 2024-03-06
feature: add support for OpenQASM measure on a subset of qubits

ashlhans:
*Issue #, if available:*

*Description of changes:*

- Add a `Measure` operator that measures the qubit targets specified.
- Removed the Validation error in GateModelQuantumTaskResult to check that measured qubits match the qubit count
- Added `Measure` to the circuit diagram

*Testing done:*

## Merge Checklist

#### General

#### Tests

```
>>> circ = Circuit().h(0).cnot(0, 1).measure([0])
>>> print(circ)

T  : │  0  │  1  │  2  │
      ┌───┐       ┌───┐ 
q0 : ─┤ H ├───●───┤ M ├─
      └───┘   │   └───┘ 
            ┌─┴─┐       
q1 : ───────┤ X ├───────
            └───┘       
T  : │  0  │  1  │  2  │

>>> print(circ.to_ir("OPENQASM"))

OPENQASM 3.0;
bit[1] b;
qubit[2] q;
h q[0];
cnot q[0], q[1];
b[0] = measure q[0];
```


ajberdy:
I think to add a feature like this, we should have some examples/tests of it working end to end. Can we add integ tests with the local simulator or an on-demand simulator that supports partial measurement?

src/braket/circuits/circuit.py
```diff
@@ -643,6 +651,61 @@ def add_verbatim_box(
             self._has_compiler_directives = True
         return self
 
+    def measure(self, target_qubits: np.ndarray | int = []) -> Circuit:
```
ajberdy, line 654: Should `circuit.measure()` do measure all qubits?

Also Python has some [weird behavior](https://www.codiga.io/blog/python-avoid-empty-array-parameter/) when using mutable objects as default args, so we should update this signature to use None

src/braket/circuits/measure.py
```diff
@@ -0,0 +1,109 @@
...
+        Raises:
+            ValueError: `qubit_count` is less than 1, `ascii_symbols` are `None`, or
+                `ascii_symbols` length != `qubit_count`
+        """
+        super().__init__(qubit_count=qubit_count, ascii_symbols=ascii_symbols)
+        self._target_index = kwargs.get("index")
+
+    @property
+    def name(self) -> str:
+        return self.__class__.__name__
```
ajberdy, line 51: I'm not sure I follow why we need this

src/braket/circuits/measure.py
```diff
@@ -0,0 +1,109 @@
...
+        elif ir_type == IRType.OPENQASM:
+            return self._to_openqasm(
+                target, serialization_properties or OpenQASMSerializationProperties() ** kwargs
+            )
+        else:
+            raise ValueError(f"Supplied ir_type {ir_type} is not supported.")
+
+    def _to_jaqcd(self) -> Any:
+        """Returns the JAQCD representation of the measure."""
+        raise NotImplementedError("to_jaqcd has not been implemented yet.")
```
ajberdy, line 92: "Measure instructions are not supported with JAQCD" perhaps?

src/braket/circuits/measure.py
```diff
@@ -0,0 +1,109 @@
...
+class Measure(QuantumOperator):
+    """Class `Measure` represents a measure operation on targeted qubits"""
+
+    def __init__(
+        self, qubit_count: Optional[int] = 1, ascii_symbols: Sequence[str] = ["M"], **kwargs
+    ):
+        """Inits a `Measure`.
+
+        Args:
+            qubit_count (Optional[int]): Number of qubits this measure operation interacts with.
```
ajberdy, line 37: I'm not sure about the qubit count for this op. Seems like it might just always be one, unless we want to  represent measuring multiple qubits at once as physically different than one at a time.

If we do have this arg, then it should likely be part of the `__eq__` comparison

test/integ_tests/test_measure.py
```diff
@@ -0,0 +1,87 @@
...
+@pytest.mark.parametrize(
+    "circuit, expected_measured_qubits",
+    [
+        (Circuit().h(0).cnot(0, 1).cnot(1, 2).cnot(2, 3).measure([0, 1, 3]), [0, 1, 3]),
+        (Circuit().h(0).measure(0).h(1), [0]),
+    ],
+)
+def test_measure_targets(circuit, expected_measured_qubits):
+    result = DEVICE.run(circuit, SHOTS).result()
+    print(result)
```
ajberdy, line 77: this and the test above have hanging prints

test/unit_tests/braket/circuits/test_ascii_circuit_diagram.py
```diff
@@ -872,3 +872,17 @@ def __init__(self):
...
+    expected = (
+        "T  : |0|1|2|",
+        "            ",
+        "q0 : -H-C-M-",
+        "        |   ",
+        "q1 : ---X---",
+        "",
+        "T  : |0|1|2|",
+    )
+    _assert_correct_diagram(circ, expected)
```
ajberdy, line 888: can we also test measure for multiple qubits? does it have a line connecting the Ms? do we want this? I'm thinking maybe, otherwise they may be staggered to different time steps

test/unit_tests/braket/circuits/test_circuit.py
```diff
@@ -570,6 +572,169 @@ def test_add_verbatim_box_result_types():
...
+    circ = Circuit().h(0).cnot(0, 1).cnot(1, 2).cnot(2, 3).measure([0, 1, 3])
+    expected = (
+        Circuit()
+        .add_instruction(Instruction(Gate.H(), 0))
+        .add_instruction(Instruction(Gate.CNot(), [0, 1]))
+        .add_instruction(Instruction(Gate.CNot(), [1, 2]))
+        .add_instruction(Instruction(Gate.CNot(), [2, 3]))
+        .add_instruction(Instruction(Measure(), 0))
+        .add_instruction(Instruction(Measure(), 1))
+        .add_instruction(Instruction(Measure(), 3))
```
ajberdy, line 607: these tests make me think that as implemented, measure is a 1-qubit instruction. Maybe we want to update this so you could pass in a list of targets

test/unit_tests/braket/circuits/test_circuit.py
```diff
@@ -570,6 +572,169 @@ def test_add_verbatim_box_result_types():
...
+        Circuit()
+        .add_instruction(Instruction(Gate.X(), 0))
+        .add_instruction(Instruction(Gate.X(), 1))
+        .add_instruction(Instruction(BitFlip(probability=0.1), 0))
+        .add_instruction(Instruction(Measure(), 0))
+    )
+    assert circ == expected
+
+
+def test_measure_verbatim_box():
```
ajberdy, line 625: I know some devices support verbatim but not partial verbatim (ie. the whole circuit needs to be verbatim if at any). I would assume this excludes the measure, but it would be great if we could confirm that

src/braket/circuits/circuit.py
```diff
@@ -635,6 +640,9 @@ def add_verbatim_box(
         if verbatim_circuit.result_types:
             raise ValueError("Verbatim subcircuit is not measured and cannot have result types")
 
+        if verbatim_circuit._measure_targets:
+            raise ValueError("Verbatim subcircuit is not measured.")
```
ajberdy, line 644: I see that this error is derived from the one above, but it's a bit confusing to me personally

test/unit_tests/braket/circuits/test_circuit.py
```diff
@@ -570,6 +572,169 @@ def test_add_verbatim_box_result_types():
...
+
+
+def test_measure_no_target():
+    message = "Measure must include one or more target qubits."
+    with pytest.raises(ValueError, match=message):
+        Circuit().h(0).cnot(0, 1).measure()
+
+
+def test_measure_with_result_types():
+    message = "Cannot perform a measure instruction with a result type."
```
ajberdy, line 678: I wonder if this should be updated to something like "a circuit cannot contain both measure instructions and result types". Otherwise, users may assume only that specific qubits can't have both

test/unit_tests/braket/circuits/test_circuit.py
```diff
@@ -570,6 +572,169 @@ def test_add_verbatim_box_result_types():
...
+
+
+def test_result_type_with_measure():
+    message = "Cannot add a result type with a measure instruction."
+    with pytest.raises(ValueError, match=message):
+        Circuit().h(0).measure(0).sample(observable=Observable.Z(), target=0)
+
+
+def test_measure_with_multiple_measures():
+    message = "Cannot perform more than one measure instruction."
```
ajberdy, line 690: this makes sense, but I'm not sure if there should be an issue or it just combines the targets. Another consideration, should we think about not allowing gates after measure instructions?

For both of these considerations, I think it's important to think about subroutines. For example what if a circuit subroutine uses 5 qubits, but only needs to measure 2, and you want to make a circuit that runs this subroutine on 3 sets of 5 qubits and get a total of 6 qubits measured

test/unit_tests/braket/circuits/test_circuit.py
```diff
@@ -570,6 +572,169 @@ def test_add_verbatim_box_result_types():
...
+    with pytest.raises(ValueError, match=message):
+        Circuit().h(0).cnot(0, 1).measure(0).h(2).measure(1)
+
+
+def test_measure_gate_after():
```
ajberdy, line 695: I don't think this should be valid at our current level of implementation. The local sim won't differentiate between this and having both h's before the measure. Can we update this test to assert correctness (ie. apply an x before and after measure to a qubit and assert the measurement state is 1). I think then we would see this test isn't passing

test/unit_tests/braket/circuits/test_circuit.py
```diff
@@ -570,6 +572,169 @@ def test_add_verbatim_box_result_types():
...
+                "h q[1];",
+            ]
+        ),
+        inputs={},
+    )
+    assert circ == expected
+    assert circ.to_ir("OPENQASM") == expected_ir
+
+
+def test_to_ir_with_measure():
```
ajberdy, line 720: Can we also add a test `from_ir` with measure? I believe you'll need to implement add_measure in BraketProgramContext

src/braket/circuits/circuit.py
ashlhans: Would something like `"Cannot measure a subcircuit inside a verbatim box."` be better?

src/braket/circuits/circuit.py
ashlhans: If there is no target specified, it would make sense to measure all the qubits on the circuit before the measure instruction. I was raising an error if there was no target specified but I think measuring all the qubits makes more sense.

test/integ_tests/test_measure.py
ashlhans: Good catch

src/braket/circuits/circuit.py
ashlhans: Updated the signature to use `None` instead of `[]` and updated to measure all qubits if no targets are provided.

src/braket/circuits/measure.py
ashlhans: You are right. This is not used. Removed it.

src/braket/circuits/measure.py
ashlhans: Updated message to `"Measure instructions are not supported with JAQCD"`

src/braket/circuits/measure.py
ashlhans: Added `qubit_count` to the `__eq__` comparison.

The `qubit_count` needs to match the number of targets and if the `measure` instruction takes in a list of targets like: `Instruction(Measure(qubit_count=3), [0, ,1 , 2])` then Measure would need a qubit count of 3

test/unit_tests/braket/circuits/test_ascii_circuit_diagram.py
ashlhans: Added a test for multiple qubit measurements here and in test_unicode_circuit_diagram

test/unit_tests/braket/circuits/test_circuit.py
ashlhans: Updated to `a circuit cannot contain both measure instructions and result types`

test/unit_tests/braket/circuits/test_circuit.py
ashlhans: I checked with Cody and the measure instruction should be fine _outside_ the verbatim box for devices that support both verbatim and partial verbatim, so long as there is not a measure _inside_ the box (ie. mid-circuit measurement).

test/unit_tests/braket/circuits/test_circuit.py
ashlhans: Added a test for `from_ir` with measure and updated the existing tests.

test/unit_tests/braket/circuits/test_circuit.py
ashlhans: I updated this to raise an error if there is a gate after the measure instruction to avoid any mid-circuit measurement confusion.

test/unit_tests/braket/circuits/test_circuit.py
ashlhans: I think we should not allow gates after measure instructions. That would likely cause confusion as it would not perform mid-circuit measurement. I have updated `circuit.py` to raise an error if a gate is added after a measure instruction.

We should allow multiple measure instructions so long as they are at the end of the circuit. The `braket_program_context` will treat each individual OpenQASM `measure` as a separate instruction that will lead to multiple measure instructions that we should combine at the end of the circuit.

This circuit `Circuit().h(0).cnot(0, 1).cnot(1, 2).cnot(2, 3).measure([0, 2]).measure(3).measure(1)` would measure the targets at the end but the diagram should probably be cleaned up to have all the measurements lined up regardless of each one's target(s).
```
T  : │  0  │  1  │  2  │  3  │     4     │
      ┌───┐                   ┌───┐       
q0 : ─┤ H ├───●───────────────┤ M ├───────
      └───┘   │               └─┬─┘       
            ┌─┴─┐               │   ┌───┐ 
q1 : ───────┤ X ├───●───────────┼───┤ M ├─
            └───┘   │           │   └───┘ 
                  ┌─┴─┐       ┌─┴─┐       
q2 : ─────────────┤ X ├───●───┤ M ├───────
                  └───┘   │   └───┘       
                        ┌─┴─┐ ┌───┐       
q3 : ───────────────────┤ X ├─┤ M ├───────
                        └───┘ └───┘       
T  : │  0  │  1  │  2  │  3  │     4     │
```

src/braket/circuits/circuit.py
```diff
@@ -298,6 +300,9 @@ def add_result_type(
         if target_mapping and target is not None:
             raise TypeError("Only one of 'target_mapping' or 'target' can be supplied.")
 
+        if self._measure_targets:
+            raise ValueError("Cannot add a result type with a measure instruction.")
```
rmshaffer, line 304:
```suggestion
            raise ValueError("Cannot add a result type to a circuit which already contains a measure instruction.")
```

src/braket/circuits/circuit.py
```diff
@@ -458,6 +463,12 @@ def add_instruction(
         if target_mapping and target is not None:
             raise TypeError("Only one of 'target_mapping' or 'target' can be supplied.")
 
+        # Check if there is a measure instruction on the circuit
+        if not isinstance(instruction.operator, Measure) and any(
+            isinstance(instruction.operator, Measure) for instruction in self.instructions
+        ):
+            raise ValueError("Cannot add a gate or noise after a measure instruction.")
```
rmshaffer, line 470: Technically, we only need to prevent this in the case that `_measure_targets` intersects with `target` or `target_mapping.values()`. So this is more restrictive than it needs to be. However, if this an intentional choice for some reason (for simplicity of the code, etc.), I think that's fine - just wanted to call it out.

src/braket/circuits/circuit.py
```diff
@@ -643,6 +657,69 @@ def add_verbatim_box(
...
+
+        Args:
+            target_qubits (Optional[ndarray] | Optional[int]): target qubits to measure.
+                Default=None
+
+        Returns:
+            Circuit: self
+
+        Raises:
+            IndexError: If `circuit` has no qubits.
```
rmshaffer, line 672: Should this be `self`?
```suggestion
            IndexError: If `self` has no qubits.
```

src/braket/circuits/circuit.py
```diff
@@ -643,6 +657,69 @@ def add_verbatim_box(
             self._has_compiler_directives = True
         return self
 
+    def measure(self, target_qubits: Optional[np.ndarray] | Optional[int] = None) -> Circuit:
```
rmshaffer, line 660: Could `target_qubits` be of type `QubitSetInput | None`, similar to `add_instruction`? Or is there a reason it needs to be this way?

src/braket/circuits/circuit.py
```diff
@@ -643,6 +657,69 @@ def add_verbatim_box(
...
+            Instruction('operator': CNot('qubit_count': 2), 'target': QubitSet([Qubit(0),
+                Qubit(1)]),
+            Instruction('operator': H('qubit_count': 1), 'target': QubitSet([Qubit(2)]),
+            Instruction('operator': Measure, 'target': QubitSet([Qubit(0)])]
+        """
+        # check whether measuring an empty circuit
+        if not self.qubits:
+            raise IndexError("Cannot measure an empty circuit.")
+
+        if type(target_qubits) is int:
```
rmshaffer, line 688: To pick up subclasses of `int`: 
```suggestion
        if isinstance(target_qubits, int):
```
(but this is likely irrelevant if using `QubitSetInput` for `target_qubits`)

src/braket/circuits/circuit.py
```diff
@@ -643,6 +657,69 @@ def add_verbatim_box(
...
+
+        # Check that the target qubits are on the circuit
+        if target_qubits and not all(qubit in self.qubits for qubit in target_qubits):
+            raise IndexError("Target qubits must be within the range of the current circuit.")
+
+        # Check if result types are added on the circuit
+        if self.result_types:
+            raise ValueError("a circuit cannot contain both measure instructions and result types.")
+
+        if target_qubits:
```
rmshaffer, line 699: I think we should throw an error if `target_qubits` intersects with `self._measure_targets`. i.e., the user should not be able to measure the same qubit more than once.

src/braket/circuits/circuit.py
```diff
@@ -643,6 +657,69 @@ def add_verbatim_box(
...
+            # Measure all the qubits
+            self.add_instruction(
+                Instruction(
+                    operator=Measure(
+                        qubit_count=self.qubit_count, ascii_symbols=["M"] * self.qubit_count
+                    ),
+                    target=self.qubits,
+                )
+            )
+            self._measure_targets = self.qubits
```
rmshaffer, line 719: Similarly, I think we should throw an error here if `self._measure_targets` already contains any of the qubits. Given that we don't allow gates after measurement, it doesn't make sense to allow multiple measurements of the same qubit.

src/braket/circuits/measure.py
rmshaffer: I also am not sure whether `qubit_count` should be here. It looks like we always serialize into single-qubit measure instructions in the OpenQASM anyway. (Note that we also do the same today with the implicit all-qubit measurement, so this is consistent - which is good.)

So it's just a question of whether we want the Circuit object to retain a difference between `circ.measure([0, 1])` vs. `circ.measure(0).measure(1)`. Since there's no difference in the resulting program, it feels like these should be treated the same. So personally I think it makes the most sense to treat this as a single-qubit instruction and have `circ.measure([0, 1])` be equivalent to `circ.measure(0).measure(1)` -- just like `circ.h([0, 1])` is equivalent to `circ.h(0).h(1)`.

But I don't think this is a very critical decision.

src/braket/circuits/measure.py
```diff
@@ -0,0 +1,102 @@
...
+        return self._ascii_symbols
+
+    def to_ir(
+        self,
+        target: QubitSet | None = None,
+        ir_type: IRType = IRType.OPENQASM,
+        serialization_properties: SerializationProperties | None = None,
+        **kwargs,
+    ) -> Any:
+        """Returns IR object of the meaure.
```
rmshaffer, line 58:
```suggestion
        """Returns IR object of the measure operator.
```

src/braket/circuits/circuit.py
```diff
@@ -643,6 +657,69 @@ def add_verbatim_box(
...
+
+        if target_qubits:
+            self.add_instruction(
+                Instruction(
+                    operator=Measure(
+                        qubit_count=len(target_qubits), ascii_symbols=["M"] * len(target_qubits)
+                    ),
+                    target=target_qubits,
+                )
+            )
```
rmshaffer, line 707: If `Measure` becomes a single-qubit operation (per the other comment thread), then I think this would change to be adding a list of instructions, one for each qubit in `target_qubits`.

test/unit_tests/braket/circuits/test_circuit.py
```diff
@@ -570,6 +572,184 @@ def test_add_verbatim_box_result_types():
...
+        inputs={},
+    )
+    assert circ == expected
+    assert circ.to_ir("OPENQASM") == expected_ir
+
+
+def test_measure_in_verbatim_subcircuit():
+    message = "Cannot measure a subcircuit inside a verbatim box."
+    with pytest.raises(ValueError, match=message):
+        Circuit().add_verbatim_box(Circuit().x(0).x(1).measure(0))
```
rmshaffer, line 658: Is there a technical reason we can't allow measurement inside a verbatim box? As long as that qubit isn't touched again for the remainder of the circuit, it seems like it should be ok - I imagine this is normally what a user would try to do first.

test/unit_tests/braket/circuits/test_circuit.py
```diff
@@ -1268,7 +1448,7 @@ def foo(
     "expected_circuit, ir",
     [
         (
-            Circuit().h(0, control=1, control_state=0),
+            Circuit().h(0, control=1, control_state=0).measure(0).measure(1),
```
rmshaffer, line 1451: Just a note that by modifying these tests, we lose coverage on the case where the `measure` is implicit. Maybe that's fine if there's enough coverage of that case elsewhere - just wanted to mention it.

rmshaffer (commented):
I'm excited for this feature - thanks for putting this together! I know this PR isn't marked ready for review yet, but I had a chance to look through and wanted to leave some thoughts/comments for consideration.

test/unit_tests/braket/circuits/test_circuit.py
ashlhans: Currently, a measurement inside the verbatim box ends up being a mid-circuit measurement since all qubits are measured if there is no explicit `measure` added to the circuit. (ie. behind the scenes all circuits would basically be `Circuit.h(0).measure()`)

Measurements inside the verbatim box can definitely be added if you see this as a common use case. We can not add the measurements to the outer circuit if there is a measurement inside the verbatim circuit.

src/braket/circuits/circuit.py
ashlhans: Added an error here if there are any `self._measure_targets` specified to prevent cases like `Circuit.h(0).measure(0).measure()`

src/braket/circuits/circuit.py
ashlhans: Added an error if the `target_qubits` is already in `self.measure_targets`

src/braket/circuits/circuit.py
ashlhans: Yes, that seems like a cleaner way. Changed the type to QubitSetInput

test/unit_tests/braket/circuits/test_circuit.py
rmshaffer: Oh, I see - so the measured qubits inside the verbatim box don't currently get added to _measure_targets of the outer circuit?

Are there other similar cases where we would need to detect measurements that might be hidden in some nested subcircuit? Like a subroutine?

(By the way, this is something that could likely be addressed in a later PR.)

test/unit_tests/braket/circuits/test_circuit.py
ashlhans: These tests only cover `Circuit.from_ir` which now adds any use of measure in the OpenQASM as a `Measure` instruction on the circuit.

So `b[0] = measure q[0];` will now always be added to the circuit as `.measure(0)` when converting the OpenQASM ir to a circuit. 

The implicit measure is still covered in the `Circuit.to_ir` tests (ie. `Circuit.h(0)` still assumes qubit 0 is measured)

test/unit_tests/braket/circuits/test_circuit.py
ashlhans: Correct, the measured qubits inside the verbatim box don't currently get added to the _measure_targets of the outer circuit. 

It is definitely worth testing any similar cases of measurements inside a nested subcircuit. Do you have any cases in mind that I can test? Otherwise, I can try to get a list together to make sure these get tested in a follow-up PR.

src/braket/circuits/measure.py
ashlhans: Those are good points. I removed `qubit_count` from here to be more consistent.

test/unit_tests/braket/circuits/test_circuit.py
ashlhans: Now that measure is treated as a single qubit instruction, this diagram for `Circuit().h(0).cnot(0, 1).cnot(1, 2).cnot(2, 3).measure([0, 2]).measure(3).measure(1)` looks cleaner as:

```
        T  : │  0  │  1  │  2  │  3  │  4  │
              ┌───┐                   ┌───┐ 
        q0 : ─┤ H ├───●───────────────┤ M ├─
              └───┘   │               └───┘ 
                    ┌─┴─┐             ┌───┐ 
        q1 : ───────┤ X ├───●─────────┤ M ├─
                    └───┘   │         └───┘ 
                          ┌─┴─┐       ┌───┐ 
        q2 : ─────────────┤ X ├───●───┤ M ├─
                          └───┘   │   └───┘ 
                                ┌─┴─┐ ┌───┐ 
        q3 : ───────────────────┤ X ├─┤ M ├─
                                └───┘ └───┘ 
        T  : │  0  │  1  │  2  │  3  │  4  │
```

similar to how `Circuit.h([0, 1])` is formatted
```
T  : │  0  │
      ┌───┐ 
q0 : ─┤ H ├─
      └───┘ 
      ┌───┐ 
q1 : ─┤ H ├─
      └───┘ 
T  : │  0  │
```

src/braket/circuits/circuit.py
ashlhans: Updated this to add a list of instruction since measure is now a single-qubit operation.

src/braket/circuits/circuit.py
ashlhans: That's a good call out and it is possible to make this less restrictive. One point for keeping the current error is that `add_instruction` can have no `target` and no `target_mapping` which would be a little more complex code-wise.

But, If you think it is more important to keep a customer use case, I can definitely change this.

src/braket/circuits/circuit.py
```diff
@@ -708,32 +735,13 @@ def measure(self, target_qubits: QubitSetInput | None = None) -> Circuit:
...
-            self.add_instruction(
-                Instruction(
-                    operator=Measure(
-                        qubit_count=self.qubit_count, ascii_symbols=["M"] * self.qubit_count
-                    ),
-                    target=self.qubits,
-                )
-            )
-            self._measure_targets = self.qubits
+            Circuit._add_measure(self, target_qubits=self.qubits)
```
rmshaffer, line 744:
```suggestion
            self._add_measure(target_qubits=self.qubits)
```

test/unit_tests/braket/circuits/test_circuit.py
rmshaffer: Oh, sorry I completely misunderstood that. Thanks for clearing up my confusion.

test/unit_tests/braket/circuits/test_circuit.py
rmshaffer: I don't know of any other ways to nest a subcircuit inside a circuit like this - there may not be any, which would be great.

src/braket/circuits/circuit.py
rmshaffer: Awesome!

src/braket/circuits/circuit.py
rmshaffer: I don't think it's important - in most cases users would naturally place all the measurements at the end of the circuit. If we find that it is a problem, we can always tweak it later.

rmshaffer (commented):
Thanks for the revisions - having all of the `M` gates drawn as single-qubit operations looks very nice!

test/unit_tests/braket/circuits/test_circuit.py
ajberdy: Ya I agree with your call here, looks good to me

src/braket/circuits/circuit.py
ajberdy: > Technically, we only need to prevent this in the case that _measure_targets intersects with target or target_mapping.values().

Curious about how this works. For example:
```
with_measure = Circuit().h(0).cnot(0, 1).measure(1).h(0)
without_measure = Circuit().h(0).cnot(0, 1).h(0)
```
with_measure seems to put q0 in the |+> or |-> state with 50/50 odds, depending on the outcome of the measurement, while without the measure keeps q0 entangled with q1 in a more complex state. Are these two states indistinguishable, assuming we don't touch q1 anymore?

src/braket/circuits/circuit.py
```diff
@@ -643,6 +662,89 @@ def add_verbatim_box(
...
+        # check whether measuring an empty circuit
+        if not self.qubits:
+            raise IndexError("Cannot measure an empty circuit.")
+
+        if isinstance(target_qubits, int):
+            target_qubits = [target_qubits]
+
+        # Check that the target qubits are on the circuit
+        if target_qubits and not all(qubit in self.qubits for qubit in target_qubits):
+            raise IndexError("Target qubits must be within the range of the current circuit.")
```
ajberdy, line 723: Ambiguous whether this should be the behavior or just add the qubit to the circuit as with all the other operations.

src/braket/circuits/circuit.py
```diff
@@ -643,6 +662,89 @@ def add_verbatim_box(
...
+
+        # Check if result types are added on the circuit
+        if self.result_types:
+            raise ValueError("a circuit cannot contain both measure instructions and result types.")
```
ajberdy, line 727: noticing some casing discrepancies for the errors. this is probably bigger than just this PR

src/braket/circuits/circuit.py
ashlhans: That's a fair point. Are there any specific use cases where a customer would want to measure a qubit not on the circuit?

src/braket/circuits/circuit.py
ashlhans: Yeah, it would probably be worth doing a round of testing at some point with stakeholders to figure out what cases should raise errors.

src/braket/circuits/circuit.py
rmshaffer: Yeah, it's possible for example that a user just wants to test readout error of the device. In which case they may run some gates on a small subset of qubits (or just add a time delay, without running any gates at all) and then measure all of the qubits on the device.

src/braket/circuits/circuit.py
rmshaffer: Yes, they are indistinguishable in the sense that the measurement results will be unaffected. You're right that `measure(1)` causes the qubits to immediately become unentangled - but this only really matters if we're going to do subsequent operations which depend on the state of both qubits.

Another way of looking at it is that any single-qubit operation on qubit `i`, including measurement, commutes with any single-qubit operation on qubit `j`.

rmshaffer: (Note this is very similar to the reason that "faster-than-light communication" is impossible. There's no way for qubit 0 to "know" that a measurement was performed on qubit 1, unless some information is then explicitly transferred between them, e.g., a subsequent 2-qubit gate, or some kind of classical feed-forward that depends on the measurement result.)

src/braket/circuits/circuit.py
ashlhans: In that case, this error is probably not needed as it prevents a use case. I can remove it.

src/braket/circuits/measure.py
```diff
@@ -0,0 +1,99 @@
...
+
+        Raises:
+            ValueError: `qubit_count` is less than 1, `ascii_symbols` are `None`, or
+                `ascii_symbols` length != `qubit_count`
+        """
+        super().__init__(qubit_count=1, ascii_symbols=["M"])
+        self._target_index = kwargs.get("index")
+
+    @property
+    def ascii_symbols(self) -> tuple[str, ...]:
```
ajberdy, line 41: Since this is a 1 qubit operation, wouldn't `ascii_symbols` be just a `tuple[str]`?

test/integ_tests/test_measure.py
```diff
@@ -0,0 +1,85 @@
...
+def test_unsupported_devices(arn):
+    device = AwsDevice(arn)
+    if device.status == "OFFLINE":
+        pytest.skip("Device offline")
+
+    circ = Circuit().h(0).cnot(0, 1).h(2).measure([0, 1])
+    error_string = re.escape(
+        "An error occurred (ValidationException) when calling the "
+        "CreateQuantumTask operation: Device requires all qubits in the program to be measured. "
+        "This may be caused by declaring non-contiguous qubits or measuring partial qubits"
```
ajberdy, line 41: Nit: "measuring partial qubits" sounds like you're measuring parts of qubits. The "partial" of "partial measurement" refers to part of the qubit set, not parts of qubits. Maybe "or not measuring all qubits"? @rmshaffer feel free to weigh in here as a scientist, maybe I'm overthinking and this language is common

src/braket/circuits/circuit.py
```diff
@@ -643,6 +662,85 @@ def add_verbatim_box(
...
+            ):
+                raise ValueError(
+                    f"cannot measure the same qubit(s) {', '.join(map(str, target_qubits))} "
+                    "more than once."
+                )
+            self._add_measure(target_qubits=target_qubits)
+        else:
+            # Check if any qubits are already measured
+            if self._measure_targets:
+                raise ValueError("cannot perform multiple measurements of the same qubits.")
```
ajberdy, line 738: Why not use the same error message as above?

test/unit_tests/braket/circuits/test_ascii_circuit_diagram.py
```diff
@@ -872,3 +872,35 @@ def __init__(self):
...
+        "        |       ",
+        "q1 : ---X-C-----",
+        "          |     ",
+        "q2 : -----X-C-M-",
+        "            |   ",
+        "q3 : -------X-M-",
+        "",
+        "T  : |0|1|2|3|4|",
+    )
+    _assert_correct_diagram(circ, expected)
```
ajberdy, line 906: maybe we can also add a test that's a bit of an edge case? Like having a measure on q0 then a few more ops on q1 then a measure on q1?

test/unit_tests/braket/circuits/test_circuit.py
```diff
@@ -570,6 +572,193 @@ def test_add_verbatim_box_result_types():
...
+
+
+def test_measure_empty_measure_after_measure_with_targets():
+    message = "cannot perform multiple measurements of the same qubits."
+    with pytest.raises(ValueError, match=message):
+        Circuit().h(0).measure(0).measure()
+
+
+def test_measure_gate_after():
+    message = "Cannot add a gate or noise after a measure instruction."
```
ajberdy, line 717: It sounds like we could reduce this constraint to only prevent ops on measured qubits. I think that would be valuable if you feel like the implementation won't be too complicated

src/braket/circuits/measure.py
ashlhans: Yes, good catch!

test/integ_tests/test_measure.py
rmshaffer: I agree - "or not measuring all qubits" is clearer.
```suggestion
        "This may be caused by declaring non-contiguous qubits or not measuring all qubits."
```

src/braket/circuits/circuit.py
rmshaffer: Makes sense, and having a unit test for that case would also be helpful if there isn't one already

src/braket/circuits/circuit.py
ashlhans: I was originally thinking that adding an empty `.measure(0).measure()` after other measure instructions might be a different error case than measuring a target twice `measure(0).measure(1).measure(0)`

Now that I'm looking at it again, I think these are probably similar enough to use the same error message.

test/integ_tests/test_measure.py
ashlhans: I also agree that 'not measuring all qubits' is clearer. 

This ValidationException is thrown on the service side so I don't think I can change the wording here in this PR.

src/braket/circuits/circuit.py
ashlhans: I went through and made sure every error message is lowercase to be consistent here.

src/braket/circuits/circuit.py
ashlhans: Added a unit test for `Circuit().h(0).cnot(0, 1).measure(4)`

src/braket/circuits/circuit.py
ashlhans: Updated this to use the same error message

test/unit_tests/braket/circuits/test_ascii_circuit_diagram.py
ashlhans: Added a test here and in `test_unicode_circuit_diagram` similar to what you suggested.

test/unit_tests/braket/circuits/test_circuit.py
ashlhans: Added an error when the `target`, `target_mapping` or `instruction.target` overlap with `self._measured_targets`

src/braket/circuits/circuit.py
```diff
@@ -407,6 +416,42 @@ def _add_to_qubit_observable_set(self, result_type: ResultType) -> None:
...
+        """
+        if (
+            target
+            and target in self._measure_targets
+            or (target_mapping and all(targ in self._measure_targets for targ in target_mapping))
+            or (
+                instruction.target
+                and all(targ in self._measure_targets for targ in instruction.target)
+            )
+        ):
```
ajberdy, line 450: nit: some comments on the different cases might be helpful here

src/braket/circuits/circuit.py
```diff
@@ -407,6 +416,42 @@ def _add_to_qubit_observable_set(self, result_type: ResultType) -> None:
...
+        Raises:
+            ValueError: If adding a gate or noise operation after a measure instruction.
+        """
+        if (
+            target
+            and target in self._measure_targets
+            or (target_mapping and all(targ in self._measure_targets for targ in target_mapping))
+            or (
+                instruction.target
+                and all(targ in self._measure_targets for targ in instruction.target)
```
ajberdy, line 448: should this be all or any? Hopefully we're confident in the logic here through unit tests

test/integ_tests/test_measure.py
ajberdy: Oh right, good point

ajberdy (dismissed):
Nice!

src/braket/circuits/circuit.py
ashlhans: Based on the logic, this should be `any`. Updated it and updated a test to check for targets both in `self._measure_targets` and not.

ajberdy (approved):

PR #910, opened 2024-03-12
perf: use dicts instead of ordered dicts

math411:
*Issue #, if available:*

*Description of changes:*
As of [Python 3.7](https://mail.python.org/pipermail/python-dev/2017-December/151283.html), `dict`s are now ordered. This removes the need for another import.

*Testing done:*
```
import timeit

print(timeit.timeit("from collections import OrderedDict"))
print(timeit.timeit("OrderedDict()", setup="from collections import OrderedDict"))
print(timeit.timeit("{}"))


>>>0.4242547479807399
>>>0.06917781697120517
>>>0.02174670697422698
```

## Merge Checklist

#### General

#### Tests


src/braket/circuits/moments.py
```diff
@@ -106,7 +105,7 @@ class Moments(Mapping[MomentsKey, Instruction]):
     """
 
     def __init__(self, instructions: Iterable[Instruction] | None = None):
-        self._moments: OrderedDict[MomentsKey, Instruction] = OrderedDict()
+        self._moments: dict[MomentsKey, Instruction] = {}
```
rmshaffer, line 108: Just a note that `dict` and `OrderedDict` are not equivalent. Notably, they have different logic for comparing equality - two `dict` objects are considered equal if they contain the same set of items regardless of order, whereas two `OrderedDict` objects are equal only if the items are in the same order. See: https://github.com/python/cpython/issues/87857

By changing this to `dict`, the `__eq__` method here would now return `True` even if the contents are in different order. So the `__eq__` method may need to be updated so that the comparison takes order into account.

PR #934, opened 2024-04-02
fix: Support single-register measurements in `from_ir`

ashlhans:
*Issue #, if available:*
Currently, the statement `b = measure q` in an OpenQASM program results in the following error:

`ValueError: Operator qubit count 1 must be equal to size of target qubit set QubitSet([Qubit(0), Qubit(1)])`

This is because a `measure` instruction is always a 1 qubit instruction but right now, [all the targets get added](https://github.com/amazon-braket/amazon-braket-sdk-python/blob/main/src/braket/circuits/braket_program_context.py#L170) to the measure instruction even if there are more than 1 qubit measured. 

To reproduce this issue:
```
bell_qasm = """
OPENQASM 3;

qubit[2] q;
bit[2] c;

h q[0];
cnot q[0], q[1];

c = measure q;
"""

from braket.circuits import Circuit
print(Circuit.from_ir(bell_qasm))
```
*Description of changes:*
- Add a loop to add a single measure instruction for each qubit target. 
- Add a test to prevent this error in the future

*Testing done:*

## Merge Checklist

#### General

#### Tests

### DIFF
--- a/test/unit_tests/braket/circuits/test_circuit.py
+++ b/test/unit_tests/braket/circuits/test_circuit.py
@@ -799,6 +799,24 @@ def test_from_ir_with_measure():
     assert Circuit.from_ir(source=ir.source, inputs=ir.inputs) == expected_circ
 
 
+def test_from_ir_with_single_measure():
+    ir = OpenQasmProgram(
+        source="\n".join(
+            [
+                "OPENQASM 3.0;",
+                "bit[2] b;",
+                "qubit[2] q;",
+                "h q[0];",
+                "cnot q[0], q[1];",
+                "b = measure q;",
+            ]
+        ),
+        inputs={},
+    )
+    expected_circ = Circuit().h(0).cnot(0, 1).measure(0).measure(1)
+    assert Circuit.from_ir(source=ir.source, inputs=ir.inputs) == expected_circ
+
+
 def test_add_with_instruction_with_default(cnot_instr):
     circ = Circuit().add(cnot_instr)
     assert circ == Circuit().add_instruction(cnot_instr)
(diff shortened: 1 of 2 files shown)


test/unit_tests/braket/circuits/test_circuit.py
```diff
@@ -799,6 +799,24 @@ def test_from_ir_with_measure():
     assert Circuit.from_ir(source=ir.source, inputs=ir.inputs) == expected_circ
 
 
+def test_from_ir_with_single_measure():
```
kshitijc, line 802: Could we add tests to do a round trip transformation? `Circuit -> OQ3 -> Circuit.from_ir` and assert that the original and the parsed circuit are the same

ashlhans: I will add a round trip test.

This `test_from_ir_with_single_measure` test would not have the same OpenQASM output though because `Circuit.from_ir` would not output `b = measure q` since the BDK does not handle the measure assignment. 

So  instead it would output:
```
b[0] = measure q[0];
b[1] = measure q[1];
```

mbeach-aws (approved):

test/unit_tests/braket/circuits/test_circuit.py
```diff
@@ -811,6 +811,48 @@ def test_from_ir_with_measure():
...
+                "b[1] = measure q[1];",
+            ]
+        ),
+        inputs={},
+    )
+    new_ir = circuit.to_ir("OPENQASM")
+    new_circuit = Circuit.from_ir(new_ir)
+
+    assert new_ir == ir
+    assert Circuit.from_ir(source=ir.source, inputs=ir.inputs) == circuit
```
speller26, line 852: This is already covered by other test cases, right? Not that I'm complaining.

speller26 (approved):

test/unit_tests/braket/circuits/test_circuit.py
ashlhans: Yes, it is covered by other test cases

PR #940, opened 2024-04-04
fix: measure target qubits are required

ashlhans:
*Issue #, if available:*

*Description of changes:*

The Circuit `measure` method `target_qubits` are not required.

### Changes 
- measure `target_qubits` are now required and there is no `measure_all()` equivalent
- Removed the error if the circuit is empty
- FreeParameter and float target qubits are not allowed. This will result in an error message
- Noise instructions cannot be added after a measure instruction
- `apply_readout_noise ` gets added right after the last measure instruction instead of the very end of the circuit.

*Testing done:*

## Merge Checklist

#### General

#### Tests


src/braket/circuits/circuit.py
```diff
@@ -436,26 +438,36 @@ def _check_if_qubit_measured(
...
+                    # check if there is a measure instruction on the target qubit
+                    target
+                    and target in self._measure_targets
+                )
+                # check if there is a measure instruction on any qubits in the target_mapping
+                or (
+                    target_mapping and any(targ in self._measure_targets for targ in target_mapping)
+                )
+                # If no target or target_mapping is supplied, check if there is a measure
+                # instruction on the current instructions target qubit
```
mbeach-aws, line 462: these conditions are extremely difficult to read. Could you refactor this to something simpler? Perhaps 
```
cond_1 = A and B
cond_2 = C  and D
...
if cond_1 or cond_2: ... 
```
could work

src/braket/circuits/circuit.py
```diff
@@ -750,13 +761,13 @@ def measure(self, target_qubits: QubitSetInput | None = None) -> Circuit:
...
-        if not self.qubits:
-            raise IndexError("cannot measure an empty circuit.")
-
         if isinstance(target_qubits, int):
             target_qubits = [target_qubits]
 
+        # Check if target is a FreeParameter or a float
+        if isinstance(target_qubits, FreeParameter) or isinstance(target_qubits, float):
+            raise ValueError(f"target qubit index '{target_qubits}' must be an integer.")
+
```
mbeach-aws, line 770: I think lines 767 to 770 are avoidable if line 765 casts the target qubits into a QubitSet instead of `[target_qubits]`. The QubitSet object should already contain validation logic for floats/FreeParameters

src/braket/circuits/moments.py
```diff
@@ -257,13 +257,17 @@ def sort_moments(self) -> None:
...
 
         for key, instruction in self._moments.items():
             moment_copy[key] = instruction
             if key.moment_type == MomentType.READOUT_NOISE:
                 key_readout_noise.append(key)
             elif key.moment_type == MomentType.INITIALIZATION_NOISE:
                 key_initialization_noise.append(key)
+            elif key.moment_type == MomentType.MEASURE:
+                last_measure = key.time
+                key_noise.append(key)
```
mbeach-aws, line 270: interesting, I wonder if this could have unindented consequences if we introduce mid-circuit measurement, because of `last_measure = key.time` and `max(last_measure)`.

ashlhans: Good point, this likely will need to change for mid-circuit measurement because this is assuming there is no measure _before_ the readout noise is added. 

Maybe there is a way to make sure the `last_measure` is actually the last measure on the circuit?

src/braket/circuits/circuit.py
ashlhans: Refactored these as you suggested.

src/braket/circuits/circuit.py
ashlhans: I updated line 764 - 766 to 

```
if not isinstance(target_qubits, Iterable):
       target_qubits = QubitSet(target_qubits)
```
and removed this check.

src/braket/circuits/moments.py
ashlhans: Similar to the mid-circuit measurement consequences, should the following circuit be allowed:

```
circ = Circuit().h(0).cnot(0, 1)
circ.apply_readout_noise(Noise.BitFlip(probability=0.1), target_qubits=1)
circ.measure([0])
circ.cnot(1, 2)
circ.measure(3)
```
```
T  : │  0  │  1  │        2        │  3  │
      ┌───┐       ┌───┐                   
q0 : ─┤ H ├───●───┤ M ├───────────────────
      └───┘   │   └───┘                   
            ┌─┴─┐       ┌─────────┐       
q1 : ───────┤ X ├───●───┤ BF(0.1) ├───────
            └───┘   │   └─────────┘       
                  ┌─┴─┐                   
q2 : ─────────────┤ X ├───────────────────
                  └───┘                   
                                    ┌───┐ 
q3 : ───────────────────────────────┤ M ├─
                                    └───┘ 
T  : │  0  │  1  │        2        │  3  │
```

mbeach-aws: this is interesting, and maybe beyond the scope of this PR. 

The circuit you've shown applies readout noise to qubit 1, even through the measurement is on qubit 3. That's strange and should not work. The apply readout (i.e. measurement noise), should only apply _before_ a measurement operation. The reason it currently applies at the end is because we _implicitly_ added measurements to the end of a circuit. 

So 
circ.apply_readout_noise(..., target_qubits=1)
circ.measure([0])
should have no noise
But
circ.apply_readout_noise(..., target_qubits=0)
circ.measure([0])
should have the noise before the measure operation.

ashlhans: You are right, I think this problem is probably worth it's own PR.

Maybe there should be some validation to check that a measure instruction exists on the readout target and we can keep track of all the measure locations to add the noise right before it.

src/braket/circuits/circuit.py
```diff
@@ -436,26 +438,31 @@ def _check_if_qubit_measured(
                 qubit mappings to apply to the `instruction.target`. Key is the qubit in
                 `instruction.target` and the value is what the key will be changed to.
                 Default = `None`.
+            target_qubits (QubitSet | None): the target qubits for a Noise `instruction`.
+                Default = `None`.
```
speller26, line 442: Why do we need this if `target` is already a `QubitSetInput`, which includes `QubitSet`?

src/braket/circuits/circuit.py
```diff
@@ -436,26 +438,31 @@ def _check_if_qubit_measured(
...
-            # check if there is a measure instruction on any qubits in the target_mapping
-            or (target_mapping and any(targ in self._measure_targets for targ in target_mapping))
-            # If no target or target_mapping is supplied, check if there is a measure
-            # instruction on the current instructions target qubit
-            or (
-                instruction.target
-                and any(targ in self._measure_targets for targ in instruction.target)
+        if self._measure_targets:
+            measure_on_target_qubits = target_qubits and any(target_qubits in self._measure_targets)
+            measure_on_target = target and target in self._measure_targets
```
speller26, line 449: If we use only `target`, then this can modified for both types included in `QubitSetInput`

src/braket/circuits/circuit.py
```diff
@@ -419,15 +419,17 @@ def _add_to_qubit_observable_set(self, result_type: ResultType) -> None:
...
         target_mapping: dict[QubitInput, QubitInput] | None = None,
+        target_qubits: QubitSet | None = None,
     ) -> None:
         """Checks if the target qubits are measured. If the qubit is already measured
         the instruction will not be added to the Circuit.
 
         Args:
-            instruction (Instruction): `Instruction` to add into `self`.
+            instruction (Instruction | None): `Instruction` to add into `self`.
+                Default = `None`.
```
speller26, line 432: Instead of allowing `instruction` to be `None`, why not just treat measurement as another instruction and have the check from
https://github.com/amazon-braket/amazon-braket-sdk-python/blob/e93b8a7c26d1fb9b0b8dc855895ee9b963ec7897/src/braket/circuits/circuit.py#L765-L773
just be done here? After all, we already disallow measurements on previously-measured qubits.

speller26 (commented):
Why remove `measure_all` functionality? If it's already released, wouldn't that be a breaking change?

src/braket/circuits/circuit.py
ashlhans: Since the name of this method is `_check_if_qubit_measured` it would make sense to check if the target qubit is already measured here. Would it also make sense to add the check for repeated qubits as well? I'm tempted to treat that as more of a syntax error and keep that where it is.

https://github.com/amazon-braket/amazon-braket-sdk-python/blob/e93b8a7c26d1fb9b0b8dc855895ee9b963ec7897/src/braket/circuits/circuit.py#L774-L782

src/braket/circuits/circuit.py
ashlhans: Good point. I removed `target_qubits` and adjusted the check for `target` to account for both types.

src/braket/circuits/circuit.py
ashlhans: Adjusted this check to account for both types in `QubitSetInput`

src/braket/circuits/circuit.py
speller26: It would actually make sense to check for repeated qubit indices for all instructions too, since they don't make sense; however, I think that's already done separately, and combining that logic with the corresponding check in `measure` can be done in a different PR.

ashlhans: Okay, I'll look into where the repeated qubit indices are checked and plan to combine the measure logic there in a follow-up PR

src/braket/circuits/circuit.py
```diff
@@ -440,22 +441,38 @@ def _check_if_qubit_measured(
...
-            # If no target or target_mapping is supplied, check if there is a measure
-            # instruction on the current instructions target qubit
-            or (
-                instruction.target
-                and any(targ in self._measure_targets for targ in instruction.target)
+        if self._measure_targets:
+            measure_on_target = (
+                target and any(tar in self._measure_targets for tar in target)
+                if isinstance(target, Iterable)
+                else target in self._measure_targets
```
speller26, line 448: Can this be combined into lines 453-457? An `Instruction` containing a `Measure` operator is still an instruction.

src/braket/circuits/circuit.py
```diff
@@ -440,22 +441,38 @@ def _check_if_qubit_measured(
...
+                    "cannot add a gate or noise operation on a qubit after a measure instruction."
+                )
+
+            # Check if the target_qubits are already measured
+            if isinstance(instruction.operator, Measure) and measure_on_instruction_target:
+                intersection = set(instruction.target) & set(self._measure_targets)
+                raise ValueError(
+                    f"cannot measure the same qubit(s) {', '.join(map(str, intersection))} "
+                    "more than once."
+                )
```
speller26, line 475: Can this be combine as well, and with the error message "Cannot apply instruction to measured qubits"

src/braket/circuits/circuit.py
```diff
@@ -419,15 +419,16 @@ def _add_to_qubit_observable_set(self, result_type: ResultType) -> None:
 
     def _check_if_qubit_measured(
         self,
-        instruction: Instruction,
+        instruction: Instruction | None = None,
```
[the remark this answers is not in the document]
ashlhans: Yep, `instruction` will never be `none`. Good catch! I thought I had already removed that.

src/braket/circuits/circuit.py
ashlhans: Yes, it can. I will combine these two errors

src/braket/circuits/circuit.py
speller26: Never mind, `instruction` and `target` are decoupled, so this still makes sense.

src/braket/circuits/circuit.py
```diff
@@ -440,22 +440,27 @@ def _check_if_qubit_measured(
...
-            # instruction on the current instructions target qubit
-            or (
-                instruction.target
-                and any(targ in self._measure_targets for targ in instruction.target)
+        if self._measure_targets:
+            measure_on_target = (
+                target and any(tar in self._measure_targets for tar in target)
+                if isinstance(target, Iterable)
+                else target in self._measure_targets
             )
```
speller26, line 448:
```suggestion
            measure_on_target = any(tar in self._measure_targets for tar in QubitSet(target))
```

src/braket/circuits/circuit.py
```diff
@@ -440,22 +440,27 @@ def _check_if_qubit_measured(
...
-            raise ValueError(
-                "cannot add a gate or noise operation on a qubit after a measure instruction."
+            measure_on_target_mapping = target_mapping and any(
+                targ in self._measure_targets for targ in target_mapping.values()
             )
+            measure_on_instruction_target = (
+                instruction
+                and instruction.target
+                and any(targ in self._measure_targets for targ in instruction.target)
+            )
```
speller26, line 456: `instruction` is required now
```suggestion
            measure_on_instruction_target = any(
                targ in self._measure_targets for targ in QubitSet(instruction.target)
            )
```

src/braket/circuits/circuit.py
```diff
@@ -440,22 +440,27 @@ def _check_if_qubit_measured(
...
+            if (
+                # check if there is a measure instruction on the targeted qubit(s)
+                measure_on_target
+                or measure_on_target_mapping
+                or measure_on_instruction_target
```
speller26, line 461: Alternatively, you can inline some variable definitions:
```suggestion
                measure_on_target_mapping
                or any(tar in self._measure_targets for tar in QubitSet(target))
                or any(tar in self._measure_targets for tar in QubitSet(instruction.target))
```

speller26: This would be even better if we could add `QubitSet`s:
```python
                measure_on_target_mapping
                or any(tar in self._measure_targets for tar in QubitSet(target) + QubitSet(instruction.target))
```

speller26: This would be even better if we could add `QubitSets`:
```suggestion
                measure_on_target_mapping
                or any(tar in self._measure_targets for tar in QubitSet(target) + QubitSet(instruction.target))
```

src/braket/circuits/circuit.py
```diff
@@ -440,22 +440,20 @@ def _check_if_qubit_measured(
...
-        ):
-            raise ValueError(
-                "cannot add a gate or noise operation on a qubit after a measure instruction."
+            measure_on_instruction_target = any(
+                targ in self._measure_targets for targ in QubitSet(instruction.target)
             )
+            if (
+                # check if there is a measure instruction on the targeted qubit(s)
+                measure_on_target_mapping
+                or any(tar in self._measure_targets for tar in QubitSet(target) + QubitSet(instruction.target))
```
speller26, line 454: This doesn't actually work; we'd need to support addition on `QubitSet`s
```suggestion
                or any(tar in self._measure_targets for tar in QubitSet(target))
                or any(tar in self._measure_targets for tar in QubitSet(instruction.target))
```

src/braket/circuits/circuit.py
```diff
@@ -440,22 +440,21 @@ def _check_if_qubit_measured(
...
-            and target in self._measure_targets
-            # check if there is a measure instruction on any qubits in the target_mapping
-            or (target_mapping and any(targ in self._measure_targets for targ in target_mapping))
-            # If no target or target_mapping is supplied, check if there is a measure
-            # instruction on the current instructions target qubit
-            or (
-                instruction.target
-                and any(targ in self._measure_targets for targ in instruction.target)
+        if self._measure_targets:
+            measure_on_target = any(tar in self._measure_targets for tar in QubitSet(target))
```
speller26, line 444: No longer necessary
```suggestion
```

src/braket/circuits/circuit.py
```diff
@@ -440,22 +440,21 @@ def _check_if_qubit_measured(
...
+            measure_on_target_mapping = target_mapping and any(
+                targ in self._measure_targets for targ in target_mapping.values()
             )
-        ):
-            raise ValueError(
-                "cannot add a gate or noise operation on a qubit after a measure instruction."
+            measure_on_instruction_target = any(
+                targ in self._measure_targets for targ in QubitSet(instruction.target)
             )
```
speller26, line 450: No longer necessary
```suggestion
```

speller26 (approved):

PR #955, opened 2024-04-16
feature: Direct Reservation context manager

mbeach-aws:
*Issue #, if available:*

*Description of changes:*
a new context manger to modify AwsQuantumTasks created within this context to run on a device with a reservation
    ARN.This is useful for ensuring that all quantum task
```
    Examples:
        As a context manager
        >>> with DirectReservation(device_arn, reservation_arn="<my_reservation_arn>"):
        ...     task1 = device.run(circuit, shots)
        ...     task2 = device.run(circuit, shots)

        or start the reservation
        >>> DirectReservation(device_arn, reservation_arn="<my_reservation_arn>").start()
        ... task1 = device.run(circuit, shots)
        ... task2 = device.run(circuit, shots)
```
*Testing done:*

## Merge Checklist

#### General

#### Tests


examples/reservation.py
```diff
@@ -14,11 +14,18 @@
...
 
 # To run a task in a device reservation, change the device to the one you reserved
-# and fill in your reservation ARN
+# and fill in your reservation ARN.
+with DirectReservation(device, reservation_arn="reservation ARN"):
+    task = device.run(bell, shots=100)
+print(task.result().measurement_counts)
+
+# Alternatively, you may start the reservation globally
+DirectReservation(device, reservation_arn="reservation ARN").start()
```
speller26, line 29: Does this need a corresponding `stop` call?

examples/reservation.py
```diff
@@ -14,11 +14,18 @@
...
 task = device.run(bell, shots=100, reservation_arn="reservation ARN")
```
speller26, line 30: Reservation ARN technically isn't needed for this use case, but we should still have an example for creating tasks with a reservation ARN

src/braket/aws/aws_session.py
```diff
@@ -235,6 +235,12 @@ def create_quantum_task(self, **boto3_kwargs) -> str:
         Returns:
             str: The ARN of the quantum task.
         """
+        # Add reservation arn if available and device is correct.
+        device_arn = os.getenv("AMZN_BRAKET_DEVICE_ARN_TEMP")
+        reservation_arn = os.getenv("AMZN_BRAKET_RESERVATION_ARN_TEMP")
```
speller26, line 240: Probably don't need `_TEMP`

examples/reservation.py
mbeach-aws: good catch! will fix this

src/braket/aws/aws_session.py
mbeach-aws: ok I will remove it

mbeach-aws: actually, `"AMZN_BRAKET_DEVICE_ARN"` might need the prefix to not conflict with the env var used in hybrid jobs.

src/braket/reservations/reservations.py
```diff
@@ -0,0 +1,90 @@
...
+    References:
+
+    [1] https://docs.aws.amazon.com/braket/latest/developerguide/braket-reservations.html
+    """
+
+    def __init__(self, device: Device | str, reservation_arn: str | None):
+        if isinstance(device, AwsDevice):
+            self.device_arn = device.arn
+        elif isinstance(device, str):
+            self.device_arn = device
```
peterkomar-aws, line 58: Checking the validity of the ARN like this at the start of the context would save the developer from some surprises later inside the context.
```suggestion
            self.device_arn = AwsDevice(device).arn
```

src/braket/reservations/reservations.py
```diff
@@ -0,0 +1,90 @@
...
+        elif isinstance(device, Device):  # LocalSimulator
+            self.device_arn = ""  # instead of None, use empty string
+        else:
+            raise ValueError("device must be an AwsDevice or its ARN.")
```
peterkomar-aws, line 62:
```suggestion
            raise ValueError("device must be an AwsDevice or its ARN, or a local simulator Device.")
```

src/braket/reservations/reservations.py
```diff
@@ -0,0 +1,90 @@
...
+    def __enter__(self):
+        self.start()
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb) -> None:
+        self.stop()
+
+    def start(self) -> None:
+        """Start the reservation context."""
+        if self.context_active:
```
peterkomar-aws, line 76: This does not check if the user is trying to nest DirectReservation contexts. We should detect and throw and error for
```python
with DirectReservation(device1, reservation_arn=arn1):
    with DirectReservation(device1, reservation_arn=arn2):
        pass
```
because the environmental variables would be overwritten in an unexpected way.

src/braket/reservations/reservations.py
```diff
@@ -0,0 +1,90 @@
...
+            raise RuntimeError("Reservation context is already active.")
+
+        os.environ["AMZN_BRAKET_RESERVATION_DEVICE_ARN"] = self.device_arn
```
peterkomar-aws, line 79: Using a single envionmental variable with a fixed name to refer to a single reservation excludes the possibility of managing multiple active reservations in parallel. Let's be very clear that this is an edge case, and if a developer wishes to build a server using BDK, they should not rely on this context manager.

src/braket/reservations/reservations.py
```diff
@@ -0,0 +1,90 @@
...
+        if self.reservation_arn:
+            os.environ["AMZN_BRAKET_RESERVATION_TIME_WINDOW_ARN"] = self.reservation_arn
+        self.context_active = True
+
+    def stop(self) -> None:
+        """Stop the reservation context."""
+        if not self.context_active:
```
peterkomar-aws, line 86: I'd raise a warning at most, not an error. (Raising an error here conditions the developer to be think twice before closing the context. This increases the chance that they accidentally leave it open compared to where the same context can be closed multiple times.)

src/braket/reservations/reservations.py
```diff
@@ -0,0 +1,90 @@
...
+        >>> DirectReservation(device_arn, reservation_arn="<my_reservation_arn>").start()
+        ... task1 = device.run(circuit, shots)
+        ... task2 = device.run(circuit, shots)
+
+    References:
+
+    [1] https://docs.aws.amazon.com/braket/latest/developerguide/braket-reservations.html
+    """
+
+    def __init__(self, device: Device | str, reservation_arn: str | None):
```
peterkomar-aws, line 54: I'm not following what checks we are doing in this constructor. The error message in L62 suggests that we wish to be stringent (only AwsDevice, or their ARN, and local simulator Device (?)). Is this true? If so, what is the purpose of allowing arbitrary string? 

Also, why do we fill in `""` for when device is a `Device`, what is the use case for passing a local sim device to a reservation? It's likely a mistake by the user, and we should help them catch it, no?

peterkomar-aws (commented):
Awesome feature. Let's be 100% clear in documentation that this is intended for high-level modules, and not suitable in a server-like environment, where multiple reservations may be active at the same time.

src/braket/reservations/reservations.py
mbeach-aws: good point, should be fixed now!

src/braket/reservations/reservations.py
mbeach-aws: fair enough, added a warning instead.

src/braket/aws/aws_session.py
```diff
@@ -235,6 +236,33 @@ def create_quantum_task(self, **boto3_kwargs) -> str:
         Returns:
             str: The ARN of the quantum task.
         """
+        # Add reservation arn if available and device is correct.
+        device_arn = os.getenv("AMZN_BRAKET_RESERVATION_DEVICE_ARN")
+        reservation_arn = os.getenv("AMZN_BRAKET_RESERVATION_TIME_WINDOW_ARN")
+
+        # if the task has a reservation_arn and also context does, raise a warning
```
speller26, line 243: Raise warning if reservation ARN is found in both context and task parameters

src/braket/reservations/reservations.py
```diff
@@ -0,0 +1,97 @@
...
+
+import os
+import warnings
+from contextlib import AbstractContextManager
+
+from braket.aws import AwsDevice
+from braket.devices import Device
+
+
+class DirectReservation(AbstractContextManager):
```
speller26, line 24: Why not name this `AwsReservation` or `AwsDirectReservation` and put it in `braket.aws`?

src/braket/reservations/reservations.py
speller26: Yeah, this is worth adding to the class docstring:
> Note: this context manager only works for one reservation at a time

src/braket/reservations/reservations.py
zmohammad01: lets go with `from braket.aws import DirectReservation`

mbeach-aws: okay, after chatting w @zmohammad01, sounds like 
```
from braket.aws import DirectReservation
```
is the preferred naming. So I moved it to the braket.aws folder.

speller26 (dismissed):

speller26 (approved):

kshitijc:
Context managers are typically used to allocate and release resources. If in the future we allow programmatic creation of reservations, will the proposed interface here make it hard to do that? Do we need to add in the context manager interface right now?

speller26:
The use case for reservations is booking a slot ahead of time, with enough room for cancellation if needed; creation of a reservation upon entering the context manager could be a very long blocking call. As well, the current implementation doesn't rule out the world where reservation creation gets fast and fine-grained enough that one could reserve a device for immediate use; the `reservation_arn` field could become optional, with extra optional parameters added for creation.

PR #976, opened 2024-05-13
feature: Add support for SerializableProgram abstraction to Device interface

rmshaffer:
*Issue #, if available:*

*Description of changes:*
The scenario is that an external library like [AutoQASM](https://github.com/amazon-braket/autoqasm) wants to create program objects which can be executed on Braket devices, including local simulators and AWS devices, but also wants the freedom to return (simulated) results in a format that does not necessarily match the Braket schema for gate model task results.

This change adds a minimal abstraction, `SerializableProgram`, which represents a program that can be serialized to OpenQASM. It adds support for this in the Device class so that these programs can be run on Braket, while also supporting a local simulator which wants to return a `GateModelQuantumTaskResult` directly rather than being forced to match the schema of `GateModelTaskResult`.

For example implementations of this abstraction, see:
- https://github.com/amazon-braket/autoqasm/blob/7e4441dd6cc42c1ee01b52d86e1ee7a04fa15b40/src/autoqasm/program/program.py#L121-L154
- https://github.com/amazon-braket/autoqasm/blob/7e4441dd6cc42c1ee01b52d86e1ee7a04fa15b40/src/autoqasm/program/program.py#L227-L261

*Testing done:*
Tests added to verify new code and maintain 100% code coverage.

## Merge Checklist

#### General

#### Tests


laurencap (dismissed):

yitchen-tim:
>  It adds support for this in the Device class so that these programs can be run on Braket, while also supporting a local simulator which wants to return a GateModelQuantumTaskResult directly rather than being forced to match the schema of GateModelTaskResult.

If the vision is to support any simulator that can execute OpenQASM, do you think it's worth creating a new result class instead of GateModelQuantumTaskResult? Currently, the MCM simulator writes results in the `measurements` field of `GateModelQuantumTaskResult`. Although `GateModelQuantumTaskResult` does not validate the field type, it does have type hint and doc that does not agree with the value written into the `measurements` field.

rmshaffer:
> If the vision is to support any simulator that can execute OpenQASM, do you think it's worth creating a new result class instead of GateModelQuantumTaskResult? Currently, the MCM simulator writes results in the `measurements` field of `GateModelQuantumTaskResult`. Although `GateModelQuantumTaskResult` does not validate the field type, it does have type hint and doc that does not agree with the value written into the `measurements` field.

You're right, using the `measurements` field for arbitrary `output` values is not really intended. To support arbitrary OpenQASM programs with `output` variables, eventually we would need a task result object which is more generic than the current `GateModelTaskResult` and `GateModelQuantumTaskResult`, both of which assume that the result of a task is a single bitstring. However, doing this correctly would involve defining a new task result schema (in the [schemas repo](https://github.com/amazon-braket/amazon-braket-schemas-python)) which I believe is beyond the scope of this PR.

One possible option for now would be to define some kind of generic `QuantumTaskResult` object which does not have any particular schema. But this feels like it would be a temporary solution and I'd be hesitant to add that now. Curious to get other thoughts on this.

krneta (approved):

src/braket/circuits/serialization.py
```diff
@@ -32,6 +33,31 @@ class QubitReferenceType(str, Enum):
...
+        self,
+        ir_type: IRType = IRType.OPENQASM,
+        build_if_necessary: bool = False,
+    ) -> str:
+        """Serializes the program into an intermediate representation.
+
+        Args:
+            ir_type (IRType): The IRType to use for converting the program to its
+                IR representation. Defaults to IRType.OPENQASM.
+            build_if_necessary (bool): Whether to allow the program to be implicitly
```
krneta, line 48: This is a little weird. Why do we need it?

rmshaffer: Agreed, this arg doesn't belong here. Removing it.

krneta (approved):

PR #1017, opened 2024-07-10
feature: Emulators with circuit validation and noise models

Altanali:
*Issue #, if available:*

*Description of changes:*

This PR introduces the notion of device emulators and emulation passes to the BDK; device emulators provide some form of program compilation and validation and may apply a noise model to the program before simulating the program on the BDK LocalSimulator. 

AwsDevice specific emulators are created based on the device properties and capabilities; noise models for gate-based QPUs are created based on calibration data retrieved from Braket service when creating an AwsDevice.

*Testing done:*

Unit testing covers 99% of emulator passes and noise model generation based on AwsDevice calibration data.  (TODO: Complete unit tests for 100% coverage). 

## Merge Checklist

#### General

#### Tests


src/braket/emulators/emulator_interface.py
```diff
@@ -0,0 +1,65 @@
...
+        """
+        This method passes the input program through EmulatorPasses that perform
+        only validation, without modifying the input program.
+
+        Args:
+            task_specification (ProgramType): The program to validate with this
+                emulator's validation passes.
+        """
+        for emulator_pass in self._emulator_passes:
+            if isinstance(emulator_pass, EmulatorCriterion):
```
maolinml, line 40: Why do we need this if-statement?

src/braket/emulators/emulator_passes/criteria/gate_criterion.py
```diff
@@ -0,0 +1,68 @@
...
+        """
+        Checks that all non-verbatim gates used in the circuit are in this criterion's
+        supported gate set and that all verbatim gates used in the circuit are in this
+        criterion's native gate set.
+
+        Args:
+            circuit (Circuit): The Braket circuit whose gates to validate.
+        """
+        idx = 0
+        while idx < len(circuit.instructions):
```
maolinml, line 38: High level questions. I thought if there is a verbatim box in the circuit, then all the gates have to be native gates. 
1. Does the following validation satisfies this requirement? I wasn't able to tell yet.
2. Can we simply check if there is a verbatim box in the circuit, and check against either the supported gate set and native gate set accordingly for all the gates?

src/braket/aws/aws_device.py
```diff
@@ -855,3 +866,116 @@ def _parse_calibration_json(
...
+        Examples:
+            >>> device = AwsDevice(Devices.IQM.Garnet)
+            >>> circuit = Circuit().cnot(0, 1).h(2).cz(2, 3)
+            >>> device.validate(circuit)
+            >>> # validates, compiles and runs on the local simulator.
+            >>> result = device.emulator(circuit, shots=100)
+            >>> print(result.result().measurement_counts)
+
+        Returns:
+            Emulator: An emulator for this device, if this is not a simulator device.
```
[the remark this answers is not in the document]
Altanali: Agreed - added!

src/braket/aws/aws_device.py
```diff
@@ -855,3 +866,116 @@ def _parse_calibration_json(
                     parsed_calibration_data[gate_qubit_key] = gate_qubit_pulse
 
         return parsed_calibration_data
+
+    @property
+    def emulator(self) -> Emulator:
+        """
+        A device emulator mimics the restrictions and noise of an AWS QPU by validating and
+        compiling programs before running them on a simulated backend. An emulator can be used
+        as a soft check that a program can run on an AwsDevice.
```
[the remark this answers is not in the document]
Altanali: This is a good point, I've made the proposed changes to clarify that the emulator's are specific to the exact AwsDevice being used.

src/braket/aws/aws_device.py
```diff
@@ -855,3 +866,116 @@ def _parse_calibration_json(
...
+                "Creating an emulator from a Braket managed simulator is not supported."
+            )
+        if not hasattr(self, "_emulator"):
+            self._emulator = self._setup_emulator()
+        return self._emulator
+
+    def _setup_emulator(self) -> Emulator:
+        """
+        Sets up an Emulator object whose properties mimic that of this AwsDevice, if the device is a
+        real QPU (not simulated).
```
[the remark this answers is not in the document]
Altanali: Fixed!

src/braket/aws/aws_device.py
```diff
@@ -855,3 +866,116 @@ def _parse_calibration_json(
...
+        )
+        return self._emulator
+
+    def validate(
+        self,
+        task_specification: Circuit,
+    ) -> None:
+        """
+        Runs all non-modifying emulator passes on the input program and raises an
+        error if any device-specific criterion are not met by the program. If the
```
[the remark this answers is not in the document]
Altanali: Your call out is correct, this was a mistake on my part! "Criteria" should have been used here (and in line 926) where we are referencing **multiple** device constraints/criteria.

src/braket/aws/aws_device.py
```diff
@@ -855,3 +866,116 @@ def _parse_calibration_json(
...
+        return self.emulator.run_program_passes(task_specification, apply_noise_model)
+
+    def emulate(
+        self,
+        task_specification: Circuit,
+        shots: Optional[int] = None,
+        inputs: Optional[dict[str, float]] = None,
+    ) -> QuantumTask:
+        """Emulate a quantum task specification on this quantum device emulator.
+        A quantum task can be a circuit or an annealing problem. Emulation
```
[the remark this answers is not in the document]
Altanali: The task _cannot_ be an annealing problem, this was a typo, thank you!

src/braket/aws/aws_device.py
```diff
@@ -855,3 +866,116 @@ def _parse_calibration_json(
...
+        involves running all emulator passes on the input program before running
+        the program on the emulator's backend.
+
+        Args:
+            task_specification (Circuit): Specification of a quantum task
+                to run on device.
+
+            shots (Optional[int]): The number of times to run the quantum task on the device.
```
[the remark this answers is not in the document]
Altanali: The arguments and results of Emulator.emulate() follows that of `LocalSimulator.run`; if no result types are provided (and if shots=0), `emulate()` raises an exception (`ValueError: No result types specified for circuit and shots=0. See `braket.circuits.result_types`) just as the LocalSimulator would.

src/braket/aws/aws_emulator_helpers.py
```diff
@@ -0,0 +1,201 @@
...
+    """
+    Create a QubitCountCriterion pass which checks that the number of qubits used in a program does
+    not exceed the number of qubits allowed by a QPU, as defined in the device properties.
+
+    Args:
+        properties (DeviceCapabilities): QPU Device Capabilities object with a
+            QHP-specific schema.
+
+    Returns:
+        QubitCountCriterion: An eulator pass that checks that the number of qubits used in a program
```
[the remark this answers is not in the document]
Altanali: Fixed!

src/braket/aws/aws_emulator_helpers.py
```diff
@@ -0,0 +1,201 @@
...
+        does not exceed that of the max qubit count on the device.
+    """
+    qubit_count = properties.paradigm.qubitCount
+    return QubitCountCriterion(qubit_count)
+
+
+def create_gate_criterion(properties: DeviceCapabilities) -> GateCriterion:
+    supported_gates = properties.action[DeviceActionType.OPENQASM].supportedOperations
```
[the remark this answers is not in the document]
Altanali: Fixed!

src/braket/aws/aws_emulator_helpers.py
```diff
@@ -0,0 +1,201 @@
...
+    Args:
+        properties (DeviceCapabilities): QPU Device Capabilities object with a
+            QHP-specific schema.
+
+    Returns:
+        GateCriterion: An emulator pass that checks that a circuit only uses supported gates and
+        verbatim circuits only use native gates.
+    """
+
+    if isinstance(properties, IqmDeviceCapabilities):
```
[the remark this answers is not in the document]
Altanali: Added a comment about this! To make note of the logic here: IqmDeviceCapabilities include the `start/end_verbatim_box` instructions under its device capabilities, which is a mistake. These are Braket compiler directives and we do not want to consider them when validating a program's gate operations.

src/braket/aws/aws_emulator_helpers.py
```diff
@@ -0,0 +1,201 @@
...
+
+    return GateCriterion(supported_gates=supported_gates, native_gates=native_gates)
+
+
+@singledispatch
+def create_connectivity_criterion(
+    properties: DeviceCapabilities, connectivity_graph: DiGraph
+) -> ConnectivityCriterion:
+    """
+    Creates a ConnectivityCriterion pass which validates that multi-qubit gates are applied to
```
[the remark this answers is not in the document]
Altanali: Yes, that's correct! Updated the docstring!

src/braket/aws/aws_emulator_helpers.py
```diff
@@ -0,0 +1,201 @@
...
+    provide back edges, but the calibration data only provides edges in one direction.
+    Additionally, IQM does not provide back edges in its connectivity_graph (nor is this
+    resolved manually by AwsDevice at the moment).
+    """
+    gate_connectivity_graph = connectivity_graph.copy()
+    edge_properties = properties.standardized.twoQubitProperties
+    for u, v in gate_connectivity_graph.edges:
+        edge_key = "-".join([str(qubit) for qubit in (u, v)])
+        edge_property = edge_properties.get(edge_key)
+        if not edge_property:
```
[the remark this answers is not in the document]
Altanali: Added a comment regarding this! This check is in case there is an edge between qubits in their connectivity graph, but the QHP has not provided calibration data about the gates that can be applied to the qubit-pair; the per-edge/per-two-qubit-gate calibration data is used to determine what gates can be applied to a qubit-pair.

src/braket/aws/aws_emulator_helpers.py
```diff
@@ -0,0 +1,201 @@
...
+            gate_connectivity_graph[u][v]["supported_gates"] = set()
+            continue
+        edge_supported_gates = get_qpu_gate_translation(
+            properties, [property.gateName for property in edge_property.twoQubitGateFidelity]
+        )
+        gate_connectivity_graph[u][v]["supported_gates"] = set(edge_supported_gates)
+
+    for u, v in gate_connectivity_graph.edges:
```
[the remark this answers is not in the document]
Altanali: Added! To note the reason here: because Rigetti/IQM have undirected topologies but their topology_graph is directed (and does not include reverse edges), we have to manually add reverse edges so that during validation, a circuit is not marked as invalid for using using two-qubit gates in a direction not provided explicitly by the QHP.

src/braket/aws/aws_emulator_helpers.py
```diff
@@ -0,0 +1,201 @@
...
+    """
+    Qubits in IonQ's trapped ion devices are all fully connected with identical
+    gate-pair capabilities. IonQ does not expliclty provide a set of edges for
+    gate connectivity between qubit pairs in their trapped ion QPUs.
+    We extrapolate gate connectivity across all possible qubit edge pairs.
+    """
+    gate_connectivity_graph = connectivity_graph.copy()
+    native_gates = get_qpu_gate_translation(properties, properties.paradigm.nativeGateSet)
+
+    for edge in gate_connectivity_graph.edges:
```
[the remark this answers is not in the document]
Altanali: The connectivity_graph for IonQ actually already has reverse edges as it's created using `nx.complete_graph` which includes edges in both directions!

src/braket/aws/aws_emulator_helpers.py
```diff
@@ -0,0 +1,201 @@
...
+@create_gate_connectivity_criterion.register(IonqDeviceCapabilities)
+def _(properties: IonqDeviceCapabilities, connectivity_graph: DiGraph) -> GateConnectivityCriterion:
+    """
+    Qubits in IonQ's trapped ion devices are all fully connected with identical
+    gate-pair capabilities. IonQ does not expliclty provide a set of edges for
+    gate connectivity between qubit pairs in their trapped ion QPUs.
+    We extrapolate gate connectivity across all possible qubit edge pairs.
+    """
+    gate_connectivity_graph = connectivity_graph.copy()
+    native_gates = get_qpu_gate_translation(properties, properties.paradigm.nativeGateSet)
```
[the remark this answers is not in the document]
Altanali: `properties.paradigm.nativeGateSet` is a list of strings - `get_qpu_gate_translations` can either translate a string gate name or a list of string gate names by checking if the argument `gate_name` is a string or is `Iterable`. The function `_get_qpu_gate_translations` actually performs the translations for individual gate strings.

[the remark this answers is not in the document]
Altanali: Argument docstring is updated to more explicitly describe this behavior!

src/braket/emulators/emulator_passes/criteria/emulator_criterion.py
```diff
@@ -0,0 +1,28 @@
...
+        An emulator criterion is used to perform some non-modifying validation
+        pass on an input program. Implementations of validate should return
+        nothing if the input program passes validation and raise an error otherwise.
+
+        Args:
+            circuit (Circuit): circuit to be evaluated against this criteria.
+        """
+        raise NotImplementedError
+
+    def run(self, program: ProgramType) -> ProgramType:
```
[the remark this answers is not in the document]
Altanali: The `run` method is a result of inheriting from the `EmulatorPass` class which requires all instances of `EmulatorPass` implement `run` - this was done so `EmulatorCriterion` (which just validate a program but don't change it) and other types of EmulatorPasses (e.g. a mapping/routing pass) can be linked together without having to check the type of the "EmulatorPass".

src/braket/emulators/emulator_passes/criteria/gate_criterion.py
```diff
@@ -0,0 +1,68 @@
...
+from braket.circuits.gate import Gate
+from braket.circuits.translations import BRAKET_GATES
+from braket.emulators.emulator_passes.criteria.emulator_criterion import EmulatorCriterion
+
+
+class GateCriterion(EmulatorCriterion):
+    def __init__(self, supported_gates: Iterator[str] = [], native_gates: Iterator[str] = []):
+        """
+        args:
+            native_gates (Iterator[str]): A list of gates supported inside of verbatim mode by
```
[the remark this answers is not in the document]
Altanali: Good catch! Updated.

src/braket/emulators/emulator_passes/criteria/emulator_criterion.py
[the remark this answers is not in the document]
Altanali: Doc string added to describe this behavior!

src/braket/emulators/emulator_passes/criteria/gate_criterion.py
```diff
@@ -0,0 +1,68 @@
...
+            the emulator.
+            supported_gates (Iterator[str]): A list of gates supported outside of verbatim mode
+            by the emulator. A gate is a Braket gate name.
+        """
+        if len(supported_gates) == 0 and len(native_gates) == 0:
+            raise ValueError("Supported gate set or native gate set must be provided.")
+
+        try:
```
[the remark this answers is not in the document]
Altanali: I agree with this, splitting it out into two try-except statements.

src/braket/emulators/emulator_passes/criteria/gate_criterion.py
```diff
@@ -0,0 +1,68 @@
...
+                ):
+                    instruction = circuit.instructions[idx]
+                    if isinstance(instruction.operator, Gate):
+                        gate = instruction.operator
+                        if not type(gate) in self._native_gates:
+                            raise ValueError(
+                                f"Gate {gate.name} is not a native gate supported by this device."
+                            )
+                    idx += 1
+                if idx == len(circuit.instructions) or not isinstance(
```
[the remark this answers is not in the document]
Altanali: That's fair, this if-statement should realistically never be triggered unless a user is intentionally adding standalone StartVerbatimBox instructions for some reason.

src/braket/emulators/emulator_passes/criteria/gate_connectivity_criterion.py
```diff
@@ -0,0 +1,121 @@
...
+
+from braket.circuits.circuit import Circuit
+from braket.circuits.compiler_directives import EndVerbatimBox, StartVerbatimBox
+from braket.circuits.gate import Gate
+from braket.emulators.emulator_passes.criteria.emulator_criterion import EmulatorCriterion
+from braket.registers.qubit_set import QubitSet
+
+
+class GateConnectivityCriterion(EmulatorCriterion):
+    def __init__(
```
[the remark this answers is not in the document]
Altanali: 1. The scenario you've described should be an error case, if `directed=False` reverse edges should have the same supported gates or should not be provided (and later filled in). 

2. While no current gate-based QPUs have directed edges with asymmetric supported gate sets, supporting directed graphs was done in consideration of QPUs like OQC's Lucy, which has a directed topology - because QPUs with directed topologies _have_ been supported by Braket, I felt it necessary to support this for future cases. Additionally, I had considered a case where we might _remove_ supported gates in certain directions if, for instance, the fidelity of a gate applied in one direction was far worse than in the other direction.

src/braket/emulators/emulator.py
```diff
@@ -0,0 +1,188 @@
...
+        """Emulate a quantum task specification on this quantum device emulator.
+        A quantum task can be a circuit or an annealing problem. Emulation
+        involves running all emulator passes on the input program before running
+        the program on the emulator's backend.
+
+        Args:
+            task_specification (Union[Circuit, OpenQasmProgram]): Specification of a quantum task
+                to run on device.
+            shots (Optional[int]): The number of times to run the quantum task on the device.
+                Default is `None`.
```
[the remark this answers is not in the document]
Altanali: Shots default=None follows the default value for the shots argument in the run methods for `aws_device` and `local_simulator`.

src/braket/emulators/emulator.py
```diff
@@ -0,0 +1,188 @@
...
+        self._noise_model = noise_model
+        self._backend = LocalSimulator(backend="braket_dm", noise_model=noise_model)
+
+    def run_program_passes(
+        self, task_specification: ProgramType, apply_noise_model: bool = True
+    ) -> ProgramType:
+        """
+        Passes the input program through all EmulatorPass objects contained in this
+        emulator and applies the emulator's noise model, if it exists, before
+        retruning the compiled program.
```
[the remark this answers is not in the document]
Altanali: Good catch! Fixed!

src/braket/emulators/emulator_interface.py
Altanali: In `validate()`, we only want to run the `EmulatorCriterion` emulator passes that are only used for validation, and are non-modifying; this check is used to avoid modifying `EmulatorPass` instances (currently, there are none, but if we have mapping/routing or nativization passes, we will want to avoid applying those to the Program when calling `validate()`).

src/braket/emulators/emulator_passes/criteria/gate_criterion.py
Altanali: 1. If there is a verbatim box in the circuit, all the gates _don't_ have to be native gates - just the ones in the verbatim box. Supported gates outside the verbatim box can still be nativized without any issue!

**_However_**, if there is a verbatim box, then _all qubit references must be to hardware qubits_ (this is checked in `ConnectivityCriterion`). 

2. Because of the comment in response to (1), we need to use both the supported gate set and the native gate set when validating a circuit.

src/braket/aws/aws_emulator_helpers.py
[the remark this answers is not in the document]
Altanali: Update: start/end_verbatim_box have been **removed** from the IQM Device Capabilities struct so this check can be removed as well.

src/braket/aws/aws_emulator_helpers.py
```diff
@@ -0,0 +1,200 @@
...
+        verbatim circuits only use native gates.
+    """
+
+    supported_gates = properties.action[DeviceActionType.OPENQASM].supportedOperations
+    native_gates = properties.paradigm.nativeGateSet
+
+    return GateCriterion(supported_gates=supported_gates, native_gates=native_gates)
+
+
+@singledispatch
```
speller26, line 56: `singledispatch` produces really ugly documentation when used on public functions; only use it on private methods, like so:
```python
def connectivity_criterion(...):
    return _connectivity_criterion(...)

@singledispatch
def _connectivity_criterion(...):
    ...
```

src/braket/aws/aws_emulator_helpers.py
```diff
@@ -0,0 +1,200 @@
...
+    gate_connectivity_graph = connectivity_graph.copy()
+    native_gates = get_qpu_gate_translation(properties, properties.paradigm.nativeGateSet)
+
+    for edge in gate_connectivity_graph.edges:
+        gate_connectivity_graph[edge[0]][edge[1]]["supported_gates"] = set(native_gates)
+
+    return GateConnectivityCriterion(gate_connectivity_graph)
+
+
+def get_qpu_gate_translation(
```
speller26, line 154: Are the names not already standardized across QPUs? This shouldn't be necessary.

src/braket/aws/aws_noise_models.py
```diff
@@ -0,0 +1,275 @@
...
+    OneQubitProperties,
+    StandardizedGateModelQpuDeviceProperties,
+)
+from braket.devices import Devices
+
+"""
+ The following gate duration values are not available through Braket device
+ calibration data and must be hardcoded.
+"""
+QPU_GATE_DURATIONS = {
```
speller26, line 32: Make this private; in fact, make things private whenever possible.

src/braket/aws/aws_noise_models.py
```diff
@@ -0,0 +1,275 @@
...
+        for edge in self.two_qubit_edge_specs.keys():
+            if edge[0] not in self.qubit_labels or edge[1] not in self.qubit_labels:
+                raise ValueError(f"Invalid qubit pair {edge}")
+
+    def __post_init__(self):
+        self._validate_single_qubit_specs()
+        self._validate_two_qubit_specs()
+
+
+def create_device_noise_model(properties: DeviceCapabilities, arn: str) -> NoiseModel:
```
speller26, line 78: `device_noise_model` is sufficient

src/braket/emulators/emulator_passes/emulator_pass.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+class EmulatorPass(ABC):
+    @abstractmethod
+    def run(self, program: ProgramType) -> ProgramType:
+        """Runs the emulator pass on the provided program.
+        Args:
+            program (ProgramType): The program to run the emulator pass on.
+        Returns:
+            ProgramType: The program after the emulator pass has been applied.
```
speller26, line 14: nit
```suggestion
        """Runs the emulator pass on the provided program.

        Args:
            program (ProgramType): The program to run the emulator pass on.

        Returns:
            ProgramType: The program after the emulator pass has been applied.
```

src/braket/emulators/emulator_passes/criteria/connectivity_criterion.py
```diff
@@ -0,0 +1,143 @@
...
+                        mapping integers (nodes) to a list of integers (adjancency lists): {e}"
+                )
+        else:
+            self._connectivity_graph = connectivity_graph
+
+        if not directed:
+            for edge in self._connectivity_graph.edges:
+                self._connectivity_graph.add_edge(edge[1], edge[0])
+
+    def validate(self, circuit: Circuit) -> None:
```
speller26, line 75: Should this also have an abstract parent method?

src/braket/emulators/emulator_passes/criteria/connectivity_criterion.py
```diff
@@ -0,0 +1,143 @@
...
+        """
+        Verifies that any verbatim box in a circuit is runnable with respect to the
+        device connectivity definied by this criterion. If any sub-circuit of the
+        input circuit is verbatim, we validate the connectivity of all gate operations
+        in the circuit.
+
+        Args:
+            circuit (Circuit): The Braket circuit whose gate operations to
+                validate.
+        """
```
speller26, line 85: Should also add a `Raises` section

speller26 (commented):
Still reviewing; great work though!

src/braket/emulators/emulator_passes/criteria/connectivity_criterion.py
Altanali: Are you referring to `validate()` for the `ConnectivityCriterion` class? If so, `ConnectivityCriterion.validate()` implements the abstractmethod, `EmulatorCriterion.validate` from the `EmulatorCriterion` class that `ConnectivityCriterion` inherits from. 

I'm not sure if this answers your question?

src/braket/aws/aws_emulator_helpers.py
Altanali: There are a few places where translation to a Braket gate name was necessary, i.e. gate names in Rigetti calibration data ("CPHASE" is used instead of the standard "cphaseshift") or in IonQs native gate set (translating "GPI"/"GPI2" to "GPi"/"GPi2"). I felt this was a little cleaner than doing translations in each place required.

src/braket/aws/aws_emulator_helpers.py
Altanali: Updated! Thank you for this callout, I wasn't aware of that interaction!

src/braket/aws/aws_emulator_helpers.py
```diff
@@ -0,0 +1,200 @@
...
+from braket.device_schema.rigetti import RigettiDeviceCapabilities
+from braket.emulators.emulator_passes import (
+    ConnectivityCriterion,
+    GateConnectivityCriterion,
+    GateCriterion,
+    QubitCountCriterion,
+)
+
+
+def create_qubit_count_criterion(properties: DeviceCapabilities) -> QubitCountCriterion:
```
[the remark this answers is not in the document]
Altanali: Updated!

src/braket/aws/aws_noise_models.py
Altanali: Updated!

src/braket/emulators/emulator_passes/emulator_pass.py
Altanali: Updated!

src/braket/emulators/emulator_passes/criteria/connectivity_criterion.py
Altanali: Added a Raises section to all methods where an exception is raised!

src/braket/emulators/emulator.py
```diff
@@ -0,0 +1,188 @@
...
+from braket.emulators.emulator_passes import EmulatorPass, ProgramType
+from braket.ir.openqasm import Program as OpenQasmProgram
+from braket.tasks import QuantumTask
+from braket.tasks.quantum_task_batch import QuantumTaskBatch
+
+
+class Emulator(Device, EmulatorInterface):
+
+    DEFAULT_SIMULATOR_BACKEND = "default"
+    DEFAULT_NOISY_BACKEND = "braket_dm"
```
speller26, line 20: private:
```
    _DEFAULT_SIMULATOR_BACKEND = "default"
    _DEFAULT_NOISY_BACKEND = "braket_dm"
```

src/braket/emulators/emulator_interface.py
```diff
@@ -0,0 +1,69 @@
+from __future__ import annotations
+
+from abc import ABC
+from typing import Iterable, Union
+
+from braket.emulators.emulator_passes import EmulatorPass, ProgramType
+from braket.emulators.emulator_passes.criteria import EmulatorCriterion
+
+
+class EmulatorInterface(ABC):
```
speller26, line 10: Is this really an `ABC` (let alone an interface) if there are no abstract methods? Maybe just call it `BaseEmulator`?

src/braket/aws/aws_emulator_helpers.py
speller26, line 1: Rename to `aws_emulation`?

src/braket/emulators/emulator_passes/criteria/emulator_criterion.py
```diff
@@ -0,0 +1,48 @@
...
+            program (ProgramType): The program to validate.
+
+        Returns:
+            ProgramType: The unmodified progam passed in as input.
+        """
+        self.validate(program)
+        return program
+
+    @abstractmethod
+    def __eq__(self, other: EmulatorCriterion) -> bool:
```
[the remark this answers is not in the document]
Altanali: Not making the `__eq__` method implementation required makes sense, it's definitely not necessary; it was originally there because I was interested in some notion of "emulator equality", but that wasn't really fleshed out. The method header/body has been removed from the ValidationPass class definition!

src/braket/emulators/emulator_passes/criteria/gate_criterion.py
```diff
@@ -0,0 +1,80 @@
...
+
+from braket.circuits import Circuit
+from braket.circuits.compiler_directives import EndVerbatimBox, StartVerbatimBox
+from braket.circuits.gate import Gate
+from braket.circuits.translations import BRAKET_GATES
+from braket.emulators.emulator_passes.criteria.emulator_criterion import EmulatorCriterion
+
+
+class GateCriterion(EmulatorCriterion):
+    def __init__(self, supported_gates: Iterator[str] = [], native_gates: Iterator[str] = []):
```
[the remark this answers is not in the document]
Altanali: Woah, thanks for sharing the article on mutable defaults - that's some weird behavior! Updated to use `None` as the default!

src/braket/emulators/emulator_passes/emulator_pass.py
```diff
@@ -0,0 +1,25 @@
+from abc import ABC, abstractmethod
+from typing import TypeVar
+
+ProgramType = TypeVar("ProgramType")
+
+
+class EmulatorPass(ABC):
```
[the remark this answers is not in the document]
Altanali: Updated in ea85675fcf3cdf0c516adef64f3f72307dae0ff5!

src/braket/aws/aws_emulator_helpers.py
Altanali: Updated!

src/braket/emulators/emulator_interface.py
Altanali: This makes sense, `EmulatorInterface` is definitely not an appropriate name for this class - changed to `BaseEmulator`.

src/braket/emulators/__init__.py
```diff
@@ -0,0 +1 @@
+from braket.emulators.emulator import Emulator  # noqa: F40
```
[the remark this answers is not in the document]
Altanali: Changed!

src/braket/emulation/emulation_passes/emulation_pass.py
speller26, line 1: This is general enough that it perhaps belongs in its own module (maybe `passes` or `transpilation`)? The class itself might even be named just `Pass`; at the very least, `EmulationPass` is to specific for what it's actually capable of.

speller26:
Looking good to me, let's see if @krneta has any feedback!

src/braket/emulation/emulation_passes/emulation_pass.py
Altanali: I've moved it to a module called "passes" under the name "BasePass" (to try and avoid collisions with the `pass` keyword even if the casing is different). Now, ValidationPass inherits directly from BasePass!

src/braket/emulation/emulation_passes/gate_device_passes/gate_validator.py
```diff
@@ -0,0 +1,86 @@
...
+        if not len(supported_gates) and not len(native_gates):
+            raise ValueError("Supported gate set or native gate set must be provided.")
+
+        try:
+            self._supported_gates = set(BRAKET_GATES[gate.lower()] for gate in supported_gates)
+        except KeyError as e:
+            raise ValueError(f"Input {str(e)} in supported_gates is not a valid Braket gate name.")
+
+        try:
+            self._native_gates = set(BRAKET_GATES[gate.lower()] for gate in native_gates)
```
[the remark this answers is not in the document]
Altanali: Changed to `frozenset`!

src/braket/emulation/emulation_passes/gate_device_passes/gate_validator.py
```diff
@@ -0,0 +1,86 @@
...
+from braket.circuits.gate import Gate
+from braket.circuits.translations import BRAKET_GATES
+from braket.emulation.emulation_passes import ValidationPass
+
+
+class GateValidator(ValidationPass[Circuit]):
+    def __init__(
+        self,
+        supported_gates: Optional[Iterator[str]] = None,
+        native_gates: Optional[Iterator[str]] = None,
```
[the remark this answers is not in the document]
Altanali: 🤦‍♂️ Thanks for catching this, fixed!

src/braket/aws/aws_device.py
```diff
@@ -856,3 +868,114 @@ def _parse_calibration_json(
...
+        A device emulator mimics the restrictions and noise of the AWS QPU by validating and
+        compiling programs before running them on a simulated backend. An emulator can be used
+        as a soft check that a program can run the target AwsDevice.
+
+        Examples:
+            >>> device = AwsDevice(Devices.IQM.Garnet)
+            >>> circuit = Circuit().cnot(0, 1).h(2).cz(2, 3)
+            >>> device.validate(circuit)
+            >>> # validates, compiles and runs on the local simulator.
+            >>> result = device.emulator(circuit, shots=100)
```
maolinml: There is a typo here. should be "device.emulate(...)".

src/braket/aws/aws_device.py
```diff
@@ -582,6 +586,26 @@ def ports(self) -> dict[str, Port]:
         self._update_pulse_properties()
         return self._ports or {}
 
+    def emulator(self, local: bool = True) -> Emulator:
```
speller26: I think we can probably remove the `local` parameter for now; if we introduce non-local later, the default will still be `True`

src/braket/aws/aws_device.py
```diff
@@ -582,6 +586,26 @@ def ports(self) -> dict[str, Port]:
...
+        """
+        A device emulator mimics the restrictions and noise of the AWS QPU by validating and
+        compiling programs before running them on a simulated backend. An emulator can be used
+        as a soft check that a program can run the target AwsDevice.
+
+        Returns:
+            Emulator: An emulator for this device, if this is not a simulator device. Raises an
+            exception if an emulator is requested for a simulator device.
+        """
+        if self._arn in [simulator_enum.value for simulator_enum in Devices.Amazon]:
```
speller26:
```suggestion
        if self._type == AwsDeviceType.SIMULATOR:
```

src/braket/emulation/emulator.py
```diff
@@ -0,0 +1,236 @@
...
+        **kwargs,
+    ):
+        Device.__init__(self, name=kwargs.get("name", "DeviceEmulator"), status="AVAILABLE")
+        BaseEmulator.__init__(self, emulator_passes)
+        self._noise_model = noise_model
+
+        backend_name = self._get_local_simulator_backend(backend, noise_model)
+        self._backend = LocalSimulator(backend=backend_name, noise_model=noise_model)
+
+    def _get_local_simulator_backend(
```
speller26: Should this belong to `LocalEmulator` instead? This class should have any notion of local vs managed backends

src/braket/emulation/device_emulator_properties.py
```diff
@@ -0,0 +1,232 @@
...
+        cls, device_properties: DeviceCapabilities
+    ) -> "DeviceEmulatorProperties":
+        if isinstance(device_properties, IonqDeviceCapabilities):
+            device_properties = standardize_ionq_device_properties(device_properties)
+        if isinstance(device_properties, DeviceCapabilities):
+            return cls._from_json(device_properties.json())
+        raise ValueError("device_properties has to be an instance of DeviceCapabilities.")
+
+    @classmethod
+    def _from_json(cls, device_properties_json: str) -> "DeviceEmulatorProperties":
```
speller26: Doesn't need to be private; allowing loading from a json string helps with flexibility

src/braket/emulation/base_emulator.py
```diff
@@ -0,0 +1,79 @@
...
+
+from __future__ import annotations
+
+from collections.abc import Iterable
+from typing import Optional, Union
+
+from braket.passes import BasePass, ProgramType, ValidationPass
+
+
+class BaseEmulator:
```
speller26: Thinking about it more, this class should be renamed; it's not doing any emulation at all, but transforming or validating the payload. Maybe `ProgramTransformer` or `PassManager`?

src/braket/aws/aws_device.py
```diff
@@ -582,6 +587,34 @@ def ports(self) -> dict[str, Port]:
...
+            raise ValueError("local can only be True.")
+
+        self._emulator = LocalEmulator.from_device_properties(self.properties)
+
+        # Add device-specific gate value validators
+        if "Ankaa" in self.name:
+            self._emulator.add_pass(AnkaaRxValidator())
+
+        if "Aria" in self.name:
+            self._emulator.add_pass(AriaMSValidator())
```
speller26: More robust to instead have a top-level map
```python
_DEVICE_PASEES = {
    Devices.IonQ.Aria1: [AriaMSValidator()],
    Devices.IonQ.Aria2: [AriaMSValidator()],
    Devices.Rigetti.Ankaa3: [AnkaaRxValidator()]
}
```
You never know, a QHP _could_ introduce different behavior for a new QPU of the same class; as well, it's safer to use defined types rather than strings.

src/braket/aws/aws_device.py
```diff
@@ -582,6 +587,34 @@ def ports(self) -> dict[str, Port]:
...
+
+        return self._emulator
```
speller26, line 594: Do we need a branch to return this emulator immediately if `self._emulator` is already populated? Because as is, this method will reconstruct the emulator each time.

src/braket/emulation/device_emulator_properties.py
speller26: Why use this instead of constructing directly from the object? Relying on JSON kvps isn't as safe.

src/braket/emulation/device_emulator_utils.py
```diff
@@ -0,0 +1,120 @@
...
+    ResultType(name="Sample", observables=["x", "y", "z", "h", "i"], minShots=1, maxShots=20000),
+    ResultType(
+        name="Expectation", observables=["x", "y", "z", "h", "i"], minShots=1, maxShots=20000
+    ),
+    ResultType(name="Variance", observables=["x", "y", "z", "h", "i"], minShots=1, maxShots=20000),
+    ResultType(name="Probability", observables=None, minShots=1, maxShots=20000),
+]
+
+
+def standardize_ionq_device_properties_json(device_properties_json: str) -> str:
```
speller26: Does this need to be public? In fact, is it necessary to have this at all (since it's deep down within capabilities parsing)?

src/braket/emulation/emulator.py
```diff
@@ -0,0 +1,316 @@
...
+
+        Returns:
+            QuantumTask: The QuantumTask tracking task execution on this device emulator.
+        """
+
+        task_specification = self.transform(task_specification, apply_noise_model=False)
+        # Don't apply noise model as the local simulator will automatically apply it.
+
+        # Remove the verbatim box before submitting to the braket density matrix simulator
+        task_specification_v2 = self._remove_verbatim_box(task_specification)
```
speller26, line 86: Is this necessary? I thought boxes were ignored for simulation.

src/braket/emulation/emulator.py
```diff
@@ -0,0 +1,316 @@
...
+
+    def _raise_exception(self, exception: Exception) -> None:
+        """
+        Wrapper for exceptions, appends the emulator's name to the exception
+        note.
+
+        Args:
+            exception (Exception): The exception to modify and raise.
+        """
+        raise EmulatorError(str(exception) + f" ({self._name})") from exception
```
speller26: Any reason these can't live in the base class?

src/braket/emulation/emulator.py
```diff
@@ -0,0 +1,316 @@
...
+        Args:
+            noisy_verbatim_circ (ProgramType): The input verbatim noisy circuit
+
+        Returns:
+            ProgramType: A verbatim noisy circuit without the verbatim boxes
+        """
+        noisy_verbatim_circ_2 = [
+            instruction
+            for instruction in noisy_verbatim_circ.instructions
+            if "VerbatimBox" not in instruction.operator.name
```
speller26: Better to use `isinstance` rather than a string match

src/braket/aws/aws_device.py
maolinml: done.

src/braket/emulation/device_emulator_properties.py
maolinml: done

src/braket/emulation/base_emulator.py
maolinml: Indeed. I think `PassManager` sounds good to me, and i have changed to it.

src/braket/aws/aws_device.py
maolinml: Good point, updated.

src/braket/aws/aws_device.py
maolinml: added a line to return emulator earlier if it is not None.

src/braket/emulation/device_emulator_properties.py
maolinml: We have in mind the use case that customer download the json from console, and instantiate the emulator from that. Also, it is easier to customize the noise model using JSON [converted to dict], right?

src/braket/emulation/device_emulator_utils.py
maolinml: Yes, since there is no "standardized" field in ionq's device properties which breaks the flow of instantiating the emulator. We could make it private, but I think we discuss to add a similar function to the schema to standardize ionq's device properties. No strong opinion here, let me know if you prefer public vs private.

src/braket/emulation/emulator.py
maolinml: no, local dm will give an error if there is a verbatim box. Actually would it worth to add this function to the Circuit class? :)

src/braket/emulation/emulator.py
maolinml: done.

src/braket/emulation/emulator.py
maolinml: Good idea. moved it and remove the validate function from the Emulator class and modify the transform function accordingly.

src/braket/aws/aws_device.py
maolinml: This is not relevant any more...

src/braket/emulation/emulator.py
```diff
@@ -0,0 +1,251 @@
...
+class Emulator(Device):
+    _DEFAULT_SIMULATOR_BACKEND = "default"
+    _DEFAULT_NOISY_BACKEND = "braket_dm"
+
+    """An emulator is a simulation device that more closely resembles
+    the capabilities and constraints of a real device or of a specific device model."""
+
+    def __init__(
+        self,
+        backend: str = "default",
```
speller26: This should match the signature of `LocalSimulator`, and allow either a string or a `BraketSimulator` instance.

src/braket/emulation/emulator.py
```diff
@@ -0,0 +1,251 @@
...
+                return Emulator._DEFAULT_NOISY_BACKEND
+            return Emulator._DEFAULT_SIMULATOR_BACKEND
+        return backend
+
+    def run(
+        self,
+        task_specification: Union[
+            Circuit,
+            OpenQasmProgram,
+        ],
```
speller26: You can use the [`TaskSpecification`](https://github.com/amazon-braket/amazon-braket-sdk-python/blob/31f56261c11796a98ebadf158c5bd56d310d4198/src/braket/tasks/quantum_task.py#L29) union type. Even though it contains more than necessary, the`LocalSimulator` technically supports all of these types, and it's up to the backend to validate the type.

src/braket/emulation/emulator.py
```diff
@@ -0,0 +1,251 @@
...
+        self,
+        task_specifications: Union[
+            Union[Circuit, OpenQasmProgram],
+            list[
+                Union[
+                    Circuit,
+                    OpenQasmProgram,
+                ]
+            ],
+        ],
```
speller26: Same, use
```suggestion
        task_specifications: TaskSpecification | Sequence[TaskSpecification]
```

src/braket/emulation/emulator.py
```diff
@@ -0,0 +1,251 @@
...
+        )
+        if (not has_measurement) and len(task_specification.result_types) == 0:
+            task_specification.measure(target_qubits=task_specification.qubits)
+
+        program = self._pass_manager.transform(task_specification)
+        return (
+            self._noise_model.apply(program) if apply_noise_model and self.noise_model else program
+        )
+
+    def _remove_verbatim_box(self, noisy_verbatim_circ: Circuit) -> Circuit:
```
speller26, line 148: Unnecessary; as of [1.27](https://github.com/amazon-braket/amazon-braket-default-simulator-python/pull/285) the local simulator now works with (ignores) verbatim boxes.

src/braket/emulation/emulator.py
```diff
@@ -0,0 +1,251 @@
...
+        This method passes the input program through Passes that perform
+        only validation, without modifying the input program.
+
+        Args:
+            task_specification (Circuit): The program to validate with this
+                emulator's validation passes.
+        """
+        self._pass_manager.validate(task_specification)
+
+    def add_pass(self, emulator_pass: Union[Iterable[ValidationPass], ValidationPass]) -> Emulator:
```
speller26: Do we need this method? We could take it out and introduce it later if it's wanted.

src/braket/emulation/local_emulator.py
```diff
@@ -0,0 +1,235 @@
...
+        emulator.add_pass(QubitCountValidator(device_em_properties.qubitCount))
+        emulator.add_pass(GateValidator(native_gates=device_em_properties.nativeGateSet))
+        emulator.add_pass(_set_up_connectivity_validator(device_em_properties))
+        emulator.add_pass(_set_up_gate_connectivity_validator(device_em_properties))
+        emulator.add_pass(
+            ResultTypeValidator(
+                device_em_properties.supportedResultTypes,
+                device_em_properties.connectivityGraph,
+            )
+        )
```
speller26: Instead of using `.add_pass`, why not just put these passes in a list and give them to the constructor?

src/braket/emulation/pass_manager.py
```diff
@@ -0,0 +1,99 @@
...
+from braket.emulation.passes import ValidationPass
+
+
+class EmulatorValidationError(Exception):
+    """Custom exception validation errors from emulators."""
+
+
+class PassManager:
+    def __init__(self, emulator_passes: Optional[Iterable[ValidationPass]] = None):
+        self._emulator_passes = emulator_passes if emulator_passes is not None else []
```
speller26: nit (since this is private): rename to `self._passes`

src/braket/emulation/pass_manager.py
```diff
@@ -0,0 +1,99 @@
...
+from braket.circuits import Circuit
+from braket.emulation.passes import ValidationPass
+
+
+class EmulatorValidationError(Exception):
+    """Custom exception validation errors from emulators."""
+
+
+class PassManager:
+    def __init__(self, emulator_passes: Optional[Iterable[ValidationPass]] = None):
```
speller26: `emulator_passes` -> `passes`

src/braket/emulation/pass_manager.py
```diff
@@ -0,0 +1,99 @@
...
+                emulator's validation passes.
+        """
+
+        try:
+            for emulator_pass in self._emulator_passes:
+                emulator_pass(task_specification)
+        except Exception as e:
+            self._raise_exception(e)
+
+    def add_pass(
```
speller26: Like with the emulator itself, we could add this later on

src/braket/emulation/passes/circuit_passes/not_implemented_validator.py
```diff
@@ -0,0 +1,85 @@
...
+
+        # Check if the circuit has any unsupported gates
+        for instruction in program.instructions:
+            if isinstance(instruction.operator, Gate):
+                gate = instruction.operator
+                if gate.name.lower() in [g.lower() for g in self._unsupported_gates]:
+                    raise ValueError(
+                        f"The gate {gate.name} is not supported by this emulator. "
+                        f"Check the device documentation for a list of supported gates."
+                    )
```
speller26: Shouldn't this be the job of `GateEmulator`?

src/braket/emulation/passes/circuit_passes/gate_connectivity_validator.py
```diff
@@ -0,0 +1,171 @@
...
+from networkx import DiGraph
+
+from braket.circuits.circuit import Circuit
+from braket.circuits.compiler_directives import EndVerbatimBox, StartVerbatimBox
+from braket.circuits.gate import Gate
+from braket.emulation.passes import ValidationPass
+from braket.registers.qubit_set import QubitSet
+
+
+class GateConnectivityValidator(ValidationPass):
```
speller26, line 26: Will this be (optionally) updated with supported angles from the native gate calibration file?

src/braket/emulation/passes/circuit_passes/gate_validator.py
```diff
@@ -0,0 +1,96 @@
...
+                verbatim modeby the emulator. A gate is a Braket gate name.
+            native_gates (Optional[Iterable[str]]): A list of gates supported inside of
+                verbatim mode by the emulator.
+
+        Raises:
+            ValueError: If supported_gates and native_gates are empty or any of the provided
+            gate are not supported by the Braket BDK.
+        """
+        supported_gates, native_gates = (supported_gates or []), (native_gates or [])
+        if not len(supported_gates) and not len(native_gates):
```
speller26: Should be fine to just use
```suggestion
        if not supported_gates and not native_gates:
```

src/braket/emulation/emulator.py
maolinml: Done.

src/braket/emulation/pass_manager.py
maolinml: removed

src/braket/emulation/pass_manager.py
maolinml: renamed

src/braket/emulation/emulator.py
maolinml: removed

src/braket/emulation/passes/circuit_passes/gate_connectivity_validator.py
maolinml: Probably no. If we want to validate gate values, we should have a separate validator.

src/braket/emulation/passes/circuit_passes/not_implemented_validator.py
maolinml: UNSUPPORTED_GATES  removed

src/braket/emulation/emulator.py
maolinml: The issue remain, so we will keep this.

src/braket/emulation/passes/circuit_passes/gate_validator.py
```diff
@@ -0,0 +1,98 @@
...
+        Raises:
+            ValueError: If a gate operation or verbatim gate operation is not in this validator's
+            supported or native gate set, respectively.
+        """
+        idx = 0
+        while idx < len(circuit.instructions):
+            instruction = circuit.instructions[idx]
+            if isinstance(instruction.operator, StartVerbatimBox):
+                idx += 1
+                while idx < len(circuit.instructions) and not isinstance(
```
speller26: Instead of nested loops, how about using a `for` loop and just keeping track of an `in_verbatim` flag? I feel like that would be easier to read.

src/braket/emulation/passes/circuit_passes/gate_connectivity_validator.py
```diff
@@ -0,0 +1,187 @@
...
+        Args:
+            circuit (Circuit): The circuit whose gate instructions need to be validated
+                against this validator's gate connectivity graph.
+
+        Raises:
+            ValueError if any of the gate operations use qubits or qubit edges that don't exist
+            in the qubit connectivity graph or the gate operation is not supported by the edge.
+        """
+        idx = 0
+        while idx < len(circuit.instructions):
```
speller26: Likewise, this can be done with a `for` loop and an `in_verbatim` flag

src/braket/emulation/emulator.py
```diff
@@ -0,0 +1,180 @@
...
+    def run_batch(
+        self,
+        task_specifications: TaskSpecification | list[TaskSpecification],
+        shots: int | None,
+        max_parallel: int | None,
+        inputs: dict[str, float] | list[dict[str, float]] | None,
+        *args: Any,
+        **kwargs: Any,
+    ) -> QuantumTaskBatch:
+        raise NotImplementedError("Emulator.run_batch() is not implemented yet.")
```
speller26:
```suggestion
        raise NotImplementedError("Emulator does not support run_batch.")
```

src/braket/emulation/device_emulator_properties.py
```diff
@@ -0,0 +1,216 @@
...
+    @staticmethod
+    def validate_oneQubitProperties(
+        oneQubitProperties: dict[str, OneQubitProperties], qubitCount: int
+    ) -> None:
+        """Validate oneQubitProperties."""
+        if len(oneQubitProperties) != qubitCount:
+            raise ValueError("The length of oneQubitProperties should be the same as qubitCount")
+
+    @staticmethod
+    def node_validator(node: str, qubit_labels: list[int], field_name: str) -> None:
```
speller26: Can these be private?

speller26: Also rename to snake_case

src/braket/emulation/device_emulator_properties.py
```diff
@@ -0,0 +1,216 @@
...
+
+            for neighbor in neighbors:
+                if int(neighbor) not in qubit_labels:
+                    raise ValueError(
+                        f"Neighbor {neighbor} for node {node} must represent a valid qubit index "
+                        f"in `qubit_labels`."
+                    )
+
+    @classmethod
+    def validate_twoQubitProperties(
```
speller26: private, snake_case

src/braket/emulation/device_emulator_properties.py
```diff
@@ -0,0 +1,216 @@
...
+        self.validate_connectivityGraph(connectivityGraph, qubit_labels)
+        self.validate_twoQubitProperties(twoQubitProperties, qubit_labels)
+
+        # Store properties
+        self.qubitCount = qubitCount
+        self.nativeGateSet = nativeGateSet
+        self.connectivityGraph = connectivityGraph
+        self.oneQubitProperties = oneQubitProperties
+        self.twoQubitProperties = twoQubitProperties
+        self.supportedResultTypes = supportedResultTypes
```
speller26: Can you make these snake_case, and make them properties (and the attributes themselves private)?

src/braket/emulation/passes/_device_emulator_validators.py
```diff
@@ -0,0 +1,76 @@
...
+                    for item in edge_property.twoQubitGateFidelity
+                    if item.gateName.lower() in BRAKET_GATES
+                ]
+                gate_connectivity_graph[edge] = set(edge_supported_gates)
+
+        reversed_gate_connectivity_graph = {}
+        for edge, edge_property in gate_connectivity_graph.items():
+            reversed_edge = (edge[1], edge[0])
+            if reversed_edge not in gate_connectivity_graph:
+                reversed_gate_connectivity_graph[reversed_edge] = edge_property
```
speller26: I feel like I'm misunderstanding this, but it looks like any graph passed to GateConnectivityValidator will effectively be undirected?

src/braket/emulation/local_emulator.py
```diff
@@ -0,0 +1,227 @@
...
+            qubits = [int(qubit) for qubit in edge.split("-")]
+            twoQubitGateFidelity = data.twoQubitGateFidelity
+
+            valid_gate_names = {
+                gate_fidelity.gateName.lower(): ind
+                for ind, gate_fidelity in enumerate(twoQubitGateFidelity)
+                if gate_fidelity.gateName.lower() in BRAKET_GATES
+            }
+
+            if len(valid_gate_names) == 0:
```
speller26: nit:
```suggestion
            if not valid_gate_names:
```

src/braket/emulation/passes/circuit_passes/gate_validator.py
maolinml: Good idea. Logic simplified.

src/braket/emulation/passes/circuit_passes/gate_connectivity_validator.py
maolinml: Good idea. Simplified.

src/braket/emulation/passes/circuit_passes/gate_connectivity_validator.py
```diff
@@ -0,0 +1,172 @@
...
+                against this validator's gate connectivity graph.
+
+        Raises:
+            ValueError if any of the gate operations use qubits or qubit edges that don't exist
+            in the qubit connectivity graph or the gate operation is not supported by the edge.
+        """
+
+        in_verbatim = False
+        for instruction in circuit.instructions:
+            if isinstance(instruction.operator, (StartVerbatimBox, EndVerbatimBox)):
```
speller26: What happens if we encounter an `EndVerbatim` first?

src/braket/emulation/passes/circuit_passes/gate_validator.py
```diff
@@ -0,0 +1,89 @@
...
+        Args:
+            circuit (Circuit): The Braket circuit whose gates to validate.
+
+        Raises:
+            ValueError: If a gate operation or verbatim gate operation is not in this validator's
+            supported or native gate set, respectively.
+        """
+        in_verbatim = False
+        for instruction in circuit.instructions:
+            if isinstance(instruction.operator, (StartVerbatimBox, EndVerbatimBox)):
```
speller26: Likewise

src/braket/emulation/passes/circuit_passes/gate_connectivity_validator.py
```diff
@@ -0,0 +1,172 @@
...
+                in_verbatim = not in_verbatim
+                continue
```
speller26:
```suggestion
        operator = instruction.operator
        if isinstance(operator, Gate) and in_verbatim:
            self._validate_gate_in_verbatim(instruction)
        elif isinstance(operator, StartVerbatimBox):
            if in_verbatim:
                raise ValueError("Already in verbatim box")
            verbatim = True
        elif isinstance(operator, EndVerbatimBox):
            if not in_verbatim:
                raise ValueError("Already outside of verbatim box")
            verbatim = False
```

src/braket/emulation/passes/circuit_passes/gate_validator.py
```diff
@@ -0,0 +1,89 @@
...
+                in_verbatim = not in_verbatim
+                continue
+            if isinstance(instruction.operator, Gate):
+                gate = instruction.operator
+                if in_verbatim:
+                    if type(gate) not in self._native_gates:
+                        raise ValueError(f"Gate {gate.name} is not a native gate for this device.")
+                elif type(gate) not in self._supported_gates:
+                    raise ValueError(f"Gate {gate.name} is not a supported gate for this device.")
```
speller26:
```suggestion
            operator = instruction.operator
            if isinstance(operator, Gate):
                if in_verbatim:
                    if type(operator) not in self._native_gates:
                        raise ValueError(f"Gate {operator.name} is not a native gate for this device.")
                elif type(operator) not in self._supported_gates:
                    raise ValueError(f"Gate {operator.name} is not a supported gate for this device.")
            elif isinstance(operator, StartVerbatimBox):
                if in_verbatim:
                    raise ValueError("Already in verbatim box")
                in_verbatim = True
            elif isinstance(operator, EndVerbatimBox):
                if not in_verbatim:
                    raise ValueError("Already outside of verbatim box")
                in_verbatim = False
```

src/braket/emulation/device_emulator_properties.py
maolinml: done.

src/braket/emulation/device_emulator_properties.py
maolinml: done

src/braket/emulation/device_emulator_properties.py
maolinml: done

src/braket/emulation/passes/circuit_passes/gate_connectivity_validator.py
maolinml: Wonderful question, i forgot about that, and I think you have solved the problem. Thanks! meta-question: Can we add this kind of validator to the Circuit class? I believe we are doing some validation in the BDK when a circuit is created, we can check a) if Verbatim box is closed and b) if EndVerbatim comes first, right?

src/braket/emulation/passes/circuit_passes/gate_validator.py
maolinml: Thanks!

src/braket/emulation/passes/circuit_passes/gate_connectivity_validator.py
speller26: Probably would make sense to add that validation, but it's not that big a deal because users already add boxes with `add_verbatim_box`, which puts the delimiters in the right place.

src/braket/emulation/passes/circuit_passes/gate_connectivity_validator.py
```diff
@@ -0,0 +1,180 @@
...
+                raise ValueError("Already in verbatim box")
+            verbatim = True
+        elif isinstance(operator, EndVerbatimBox):
+            if not in_verbatim:
+                raise ValueError("Already outside of verbatim box")
+            verbatim = False
+
+            if isinstance(instruction.operator, Gate) and in_verbatim:
+                self._validate_gate_in_verbatim(instruction)
+
```
speller26: Looks like this can't be reached.
```suggestion
```

maolinml: yeah i fixed that.

src/braket/emulation/passes/_device_emulator_validators.py
maolinml: Fixed!

Issue #1083, opened 2025-05-21
Enable graphical visualizations for circuits


rmshaffer:
**Describe the feature you'd like**
Build functionality to visualize a `Circuit` object using matplotlib and/or LaTeX. The interface should be similar to the Qiskit function [`qiskit.visualization.circuit_drawer`](https://docs.quantum.ibm.com/api/qiskit/qiskit.visualization.circuit_drawer).

> **Note for UnitaryHACK participants:** At a minimum, an implementation of this feature should support _either_ matplotlib _or_ LaTeX (both would be great!), and should support visualizing everything currently supported by the text-based circuit visualization.

**How would this feature be used? Please describe.**
This would allow Braket circuit visualizations to be much more readable, for example, in Jupyter notebooks, or for use on websites or in academic papers.

**Describe alternatives you've considered**
The existing text-based visualization via `print(circuit)` works, but it is not very readable, and cannot be used as a graphical format for exporting a circuit.

skushagra:
Hey @rmshaffer,

Good to see this issue here, I went through the hacker guide on unitaryHack and want to work on this.

I have previously worked on circuit visualizers similar to Qiskit's using Matplotlib, Plotly as well as using latex(quantikz and otherwise). 

Please let me know if this issue is open and if I can work on it.

skushagra:
Hello @rmshaffer 

Sorry for replying a little late. I saw the PR raised with respect to this last week and decided to let them continue with the issue. I'll be happy to help whereever required!

rmshaffer:
@skushagra thanks for checking in! @bryanhzaz, I know there's still a bit of work to do on your PR. If you think any help would be useful, please let us know!

skushagra:
@rmshaffer if it works, I keep sharing my ideas on the PR. I think it will be helpful! Here for any help required. I'll work on some other issues on the repo meanwhile!

Qubit1718:
Hi @rmshaffer, 

The PR uses matplotlib to create the images. Please review at your convenience.

PR #1094, opened 2025-06-12
feat: add circuit drawer module with Matplotlib (#1083)

Qubit1718:
Issue #, if available: #1083

Description of changes: 
1) Added a new file in src/braket/circuits/text_diagram_builders - matplotlib_circuit_diagram.py
2) Added the new function(MatplotlibCircuitDiagram) into __init__.py
3) Added matplotlib as requirement in setup.py
4) Added a new file for testing

Testing done: Yes

## Merge Checklist

#### General

- [ x] I have read the [CONTRIBUTING](https://github.com/amazon-braket/amazon-braket-sdk-python/blob/main/CONTRIBUTING.md) doc
- [ x] I used the PR title format described in [CONTRIBUTING](https://github.com/amazon-braket/amazon-braket-sdk-python/blob/main/CONTRIBUTING.md#PR-title-format)
- [ ]x I have updated any necessary documentation, including [READMEs](https://github.com/amazon-braket/amazon-braket-sdk-python/blob/main/README.md) and [API docs](https://github.com/amazon-braket/amazon-braket-sdk-python/blob/main/CONTRIBUTING.md#documentation-guidelines) (if appropriate)

#### Tests

- [ x] I have added tests that prove my fix is effective or that my feature works (if appropriate)
- [x ] I have checked that my tests are not configured for a specific region or account (if appropriate)

### DIFF
--- a/src/braket/circuits/text_diagram_builders/matplotlib_circuit_diagram.py
+++ b/src/braket/circuits/text_diagram_builders/matplotlib_circuit_diagram.py
@@ -1,0 +1,338 @@
+from __future__ import annotations
+
+from typing import Union
+
+import matplotlib.pyplot as plt
+from matplotlib import patches
+from matplotlib.axes import Axes
+from matplotlib.figure import Figure
+
+import braket.circuits.circuit as cir
+from braket.circuits.circuit_diagram import CircuitDiagram
+from braket.circuits.instruction import Instruction
+from braket.circuits.moments import MomentType
+from braket.circuits.result_type import ResultType
+from braket.circuits.result_types import ObservableResultType
+from braket.circuits.text_diagram_builders.text_circuit_diagram_utils import _group_items
+from braket.registers.qubit_set import QubitSet
+
+
+class MatplotlibCircuitDiagram(CircuitDiagram):
+    """A class that builds circuit diagrams using matplotlib."""
+
+    # Default styling parameters
+    GATE_BOX_HEIGHT = 0.6
+    GATE_BOX_WIDTH = 1.0
+    GATE_BOX_PADDING = 0.2
+    WIRE_SPACING = 1.0
+    COLUMN_SPACING = 1.5
+    GATE_COLOR = "#E0E0E0"
+    WIRE_COLOR = "#000000"
+    TEXT_COLOR = "#000000"
+    GLOBAL_PHASE_COLOR = "#666666"
+
+    @classmethod
+    def build_diagram(cls, circuit: cir.Circuit) -> Figure:
+        """Build a matplotlib figure for the specified circuit.
+
+        Args:
+            circuit (cir.Circuit): The circuit to build a diagram for.
+
+        Returns:
+            Figure: Matplotlib figure containing the circuit diagram.
+        """
+        if not circuit.instructions:
+            return plt.figure()
+
+        if all(m.moment_type == MomentType.GLOBAL_PHASE for m in circuit._moments):
+            fig, ax = plt.subplots(figsize=(4, 1))
+            ax.text(0.5, 0.5, f"Global phase: {circuit.global_phase}", ha="center", va="center")
+            ax.axis("off")
+            return fig
+
+        # Get circuit qubits and sort them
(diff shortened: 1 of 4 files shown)


src/braket/circuits/text_diagram_builders/matplotlib_circuit_diagram.py
```diff
@@ -0,0 +1,338 @@
...
+        fig_width = 10  # Will be adjusted based on circuit size
+        fig_height = max(4, num_qubits * 1.5)
+        fig, ax = plt.subplots(figsize=(fig_width, fig_height))
+
+        # Draw the circuit
+        cls._draw_circuit(ax, circuit, circuit_qubits)
+
+        # Adjust layout and return
+        plt.tight_layout()
+        return fig
```
rmshaffer, line 68: In a Jupyter notebook, when calling `build_diagram()`, it's always displaying the image automatically - even if I capture the return value in a variable. If I capture the return value of `build_diagram()` in a variable, or if this is not the last line in the cell, then the image should not be displayed automatically.

src/braket/circuits/text_diagram_builders/matplotlib_circuit_diagram.py
```diff
@@ -0,0 +1,338 @@
...
+            cls._draw_result_types(ax, circuit.result_types, circuit_qubits, x_pos)
+
+        # Draw global phase if present (after all instructions, so it's not overlapped)
+        if circuit.global_phase != 0:
+            y_center = (
+                0 if len(circuit_qubits) == 1 else -len(circuit_qubits) * cls.WIRE_SPACING / 2
+            )
+            # If global phase is exactly 0.5, display as 0.50π, else as a number with two decimals
+            if abs(circuit.global_phase - 0.5) < 1e-9:
+                phase_str = "0.50π"
```
rmshaffer, line 121: This is cool. Two comments
(1) would be cool to do this for other multiples of `pi`, not just `0.5`
(2) we should share the logic for displaying global phase and for angle parameters to other gates

src/braket/circuits/text_diagram_builders/matplotlib_circuit_diagram.py
```diff
@@ -0,0 +1,338 @@
...
+            str: Text representation of the instruction
+        """
+        if hasattr(instruction, "operator"):
+            operator = instruction.operator
+            # Compiler directive
+            if operator.__class__.__name__ == "CompilerDirective":
+                return str(operator.ascii_symbols[0])
+            # Measurement
+            if operator.__class__.__name__ == "Measure":
+                return "Measure"
```
rmshaffer, line 294: for text diagrams we just use `M` here - could do the same for consistency

rmshaffer (changes requested):
This is a nice start! I left a few specific comments. Overall I think there are still some critical features missing that would be great to add before we merge this PR.

To compare with our existing text-based visualization, I tested with this code snippet:
```
from braket.circuits import Circuit, MatplotlibCircuitDiagram, ResultType

verbatim_circuit = Circuit().h(0).cnot(0,1)
circuit = Circuit().add_verbatim_box(verbatim_circuit).cnot(0,2).ccnot(0,2,3).ms(1,3,1.1,1.2).add_result_type(ResultType.StateVector())

print(circuit)
plot = MatplotlibCircuitDiagram.build_diagram(circuit)
```

which produces this output:
image

We can see a few issues here off the bat:
- the controlled and multiqubit gates (CNOT, CCNOT, MS) don't visualize which specific qubits they act on, and don't indicate which qubit is the control and which is the target
- the StartVerbatim and EndVerbatim markers are not printed
- the gate angle parameters (to `MS` in this case) are not displayed
- the qubit labels could be simplified to just `q0` or `q[0]` instead of `q[Qubit(0)]`
- the result type `StateVector` is shown only on the first qubit, when it actually applies to all qubits (or, we could just omit that - since our text-based visualizer does not print out the result types)

I haven't tested thoroughly - in general, though, the guidance is that the information we show in the matplotlib image should be the same as what we show in the text-based version.

PR #1098, opened 2025-06-24
Validate hybrid jobs name

shalini9894:
Hello,
 
Below are the changes I made,

1.Added _HYBRID_JOB_NAME_PATTERN regex in  hybrid_job.py to enforce AWS job name rules.

2. Added _validate_hybrid_job_name() helper.

3. Created 8 pytest cases in test_hybrid_job_name.py covering valid and invalid names.

4. Also, I specifically ran only the part I worked thus passing 8 tests.

5. Did .venv cleanup.


shalini9894:
> Thank you for the contribution and apologies for the late response. Do you still plan to work on this? If so, it would probably be worth making this a general ARN validator rather than just a job ARN validator.

Sure. I am on it. Can you please say me what exactly I need to do? This would be helpful for me to work clearly! Thank you.

src/braket/jobs/hybrid_job.py
```diff
@@ -45,6 +46,20 @@
...
+    """Raise ValueError if the hybrid-job name is invalid."""
+    if not _HYBRID_JOB_NAME_PATTERN.fullmatch(name):
+        raise ValueError(
+            f"Invalid hybrid job name: '{name}'. "
+            "Use 1–50 characters (letters, digits, hyphens); no underscores or leading/trailing hyphens."
+        )
+
+
+
+
```
speller26:
```suggestion
```

src/braket/jobs/hybrid_job.py
```diff
@@ -45,6 +46,17 @@
 from braket.jobs.quantum_job import QuantumJob
 from braket.jobs.quantum_job_creation import _generate_default_job_name
 
+
```
speller26:
```suggestion
```

PR #1002, opened 2024-06-17
feature: Add delay instruction for circuits

rmshaffer:
*Issue #, if available:*
#974

*Description of changes:*
Starting with the changes from #993 and performing the additional work required to merge to `main`.

The biggest outstanding items are:
- [ ] `Delay` (like the existing `Barrier`) should inherit from `CompilerDirective` rather than `Gate`, since it is an instruction that affects the compilation of a program rather than the quantum state. See [this comment thread](https://github.com/amazon-braket/amazon-braket-sdk-python/pull/993#discussion_r1628291648).
- [ ] Significant duplication between the new `DurationGate` class and the existing `AngledGate` class that should be reduced. See [this comment thread](https://github.com/amazon-braket/amazon-braket-sdk-python/pull/993#discussion_r1638743919).

*Testing done:*
`tox`

## Merge Checklist

_Put an `x` in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your pull request._

#### General

- [x] I have read the [CONTRIBUTING](https://github.com/amazon-braket/amazon-braket-sdk-python/blob/main/CONTRIBUTING.md) doc
- [x] I used the commit message format described in [CONTRIBUTING](https://github.com/amazon-braket/amazon-braket-sdk-python/blob/main/CONTRIBUTING.md#commit-your-change)
- [x] I have updated any necessary documentation, including [READMEs](https://github.com/amazon-braket/amazon-braket-sdk-python/blob/main/README.md) and [API docs](https://github.com/amazon-braket/amazon-braket-sdk-python/blob/main/CONTRIBUTING.md#documentation-guidelines) (if appropriate)

#### Tests

- [ ] I have added tests that prove my fix is effective or that my feature works (if appropriate)
- [ ] I have checked that my tests are not configured for a specific region or account (if appropriate)

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #993, opened 2024-06-05
feature: Add delay and barrier for circuits

Manvi-Agrawal:
## Issue , if available:
Fixes https://github.com/amazon-braket/amazon-braket-sdk-python/issues/974

## Testing done:
- Unit test coverage is 100%.
- Running this code produces this QASM code
```python
from braket.circuits import Circuit
circ = Circuit().barrier([0, 1, 2])
print(circ.to_ir("OPENQASM").source)
```

#### OUTPUT
```text
OPENQASM 3.0;
bit[3] b;
qubit[3] q;
barrier q[0], q[1], q[2];
b[0] = measure q[0];
b[1] = measure q[1];
b[2] = measure q[2];
```

```python
from braket.circuits import Circuit

circ = Circuit().delay([0, 1, 2], 30)

print(circ.to_ir("OPENQASM").source)
```

#### OUTPUT
```text
OPENQASM 3.0;
bit[3] b;
qubit[3] q;
delay[30 s] q[0], q[1], q[2];
b[0] = measure q[0];
b[1] = measure q[1];
b[2] = measure q[2];
```

## Merge Checklist

_Put an `x` in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your pull request._

#### General

- [x] I have read the [CONTRIBUTING](https://github.com/amazon-braket/amazon-braket-sdk-python/blob/main/CONTRIBUTING.md) doc
- [x] I used the commit message format described in [CONTRIBUTING](https://github.com/amazon-braket/amazon-braket-sdk-python/blob/main/CONTRIBUTING.md#commit-your-change)
- [x] I have updated any necessary documentation, including [READMEs](https://github.com/amazon-braket/amazon-braket-sdk-python/blob/main/README.md) and [API docs](https://github.com/amazon-braket/amazon-braket-sdk-python/blob/main/CONTRIBUTING.md#documentation-guidelines) (if appropriate)

#### Tests

- [x] I have added tests that prove my fix is effective or that my feature works (if appropriate)
- [ ] I have checked that my tests are not configured for a specific region or account (if appropriate)

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

Issue #974, opened 2024-05-09
Support for adding barriers and delays to `Circuit`


rmshaffer:
**Describe the feature you'd like**

Users should be able to add barriers and delays to `Circuit` objects using the same syntax as gates.

Implementing barrier and delay instructions should involve minimal complexity. OpenQASM supports both the `barrier` and `delay` instructions at the program level, so we could simply add these instructions in global scope in the resulting OpenQASM program.

For example, the code may look like:
```
circ = Circuit().x(0).barrier([0, 1]).x(0).delay(0, 1e-6)
print(circ)
```

which would output something like:
```
T  : |0|1|2|     3     |
                     
q0 : -X-|-X-delay(1e-6)-
        |            
q1 : ---|---------------

T  : |0|1|2|     3     |
```

**How would this feature be used? Please describe.**

The `barrier` and `delay` instructions are useful when precise timing of quantum operations is needed within a program. One example is in https://github.com/amazon-braket/amazon-braket-sdk-python/issues/417, where the time of measurement needs to be determined precisely.

**Describe alternatives you've considered**

Today, the Braket service supports `barrier` and `delay` as OpenPulse instructions, which is a workaround for achieving this functionality. For example, the program below contains a `barrier` on qubits `$0` and `$1` inside a `cal` block:

```
program = Program(source="""OPENQASM 3.0;
bit[1] b;
x $0;
cal {
    waveform t = {0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0};
    play(q0_drive, t);
    barrier $0, $1;
}
b[0] = measure $0;
""")
```

rmshaffer:
An implementation of `Barrier` would probably look something like this:

```
from braket.circuits import Gate, Instruction, circuit

class Barrier(Gate):
    def __init__(self, qubit_count):
        super().__init__(qubit_count=qubit_count, ascii_symbols=[f"." for i in range(qubit_count)])

    def bind_values(self, **kwargs):
        raise NotImplementedError

    @property
    def _qasm_name(self):
        return f"barrier"

    def __hash__(self):
        return hash((self.name, self.qubit_count, self.qubit_count))

    @staticmethod
    @circuit.subroutine(register=True)
    def barrier(target):
        return Instruction(Barrier(len(target)), target=target)

Gate.register_gate(Barrier)
```

Example usage of `Barrier` would then look like:
```
from braket.circuits import Circuit

circ = Circuit().barrier([0,1,2])
[...]

PR #1118, opened 2025-08-29
feature: add barrier instruction

nilsquet:
*Issue #, if available:*
#974 

*Description of changes:*
Starting with the changes from #993 and #1002, the `barrier` instruction is implemented as a `CompilerDirective`. To this end, changes were necessary in a few files.

*Testing done:*
`tox`
## Merge Checklist

#### General

#### Tests

### DIFF
--- a/src/braket/circuits/circuit.py
+++ b/src/braket/circuits/circuit.py
@@ -733,7 +733,8 @@ class Circuit:  # noqa: PLR0904
         """Add a barrier compiler directive to the circuit.
 
         Args:
-            target (QubitSetInput | None): Target qubits for the barrier. If None, applies to all qubits in the circuit.
+            target (QubitSetInput | None): Target qubits for the barrier.
+            If None, applies to all qubits in the circuit.
 
         Returns:
             Circuit: self
@@ -743,9 +744,11 @@ class Circuit:  # noqa: PLR0904
             >>> circ = Circuit().h(0).h(1).barrier()  # barrier on all qubits
         """
         target_qubits = self.qubits if target is None else QubitSet(target)
-        
+
         if target_qubits:
-            self.add_instruction(Instruction(compiler_directives.Barrier(list(target_qubits)), target=target_qubits))
+            self.add_instruction(
+                Instruction(compiler_directives.Barrier(list(target_qubits)), target=target_qubits)
+            )
             self._has_compiler_directives = True
         return self
 
--- a/src/braket/circuits/circuit.py
+++ b/src/braket/circuits/circuit.py
@@ -729,6 +729,26 @@ class Circuit:  # noqa: PLR0904
             self._has_compiler_directives = True
         return self
 
+    def barrier(self, target: QubitSetInput | None = None) -> Circuit:
+        """Add a barrier compiler directive to the circuit.
+
+        Args:
+            target (QubitSetInput | None): Target qubits for the barrier. If None, applies to all qubits in the circuit.
+
+        Returns:
+            Circuit: self
+
+        Examples:
+            >>> circ = Circuit().h(0).barrier([0, 1]).cnot(0, 1)
+            >>> circ = Circuit().h(0).h(1).barrier()  # barrier on all qubits
+        """
+        target_qubits = self.qubits if target is None else QubitSet(target)
+        
+        if target_qubits:
+            self.add_instruction(Instruction(compiler_directives.Barrier(list(target_qubits)), target=target_qubits))
+            self._has_compiler_directives = True
+        return self
+
     def _add_measure(self, target_qubits: QubitSetInput) -> None:
         """Adds a measure instruction to the the circuit
 
--- a/src/braket/circuits/compiler_directives.py
+++ b/src/braket/circuits/compiler_directives.py
@@ -79,11 +79,6 @@ class Barrier(CompilerDirective):
     def qubit_count(self) -> int:
         return len(self._qubit_indices)
 
-    def __eq__(self, other: CompilerDirective) -> bool:
-        if isinstance(other, Barrier):
-            return self._qubit_indices == other._qubit_indices
-        return super().__eq__(other)
-
     def _to_jaqcd(self) -> Any:
         raise NotImplementedError("Barrier is not supported in JAQCD")
 
--- a/src/braket/circuits/compiler_directives.py
+++ b/src/braket/circuits/compiler_directives.py
@@ -95,6 +95,3 @@ class Barrier(CompilerDirective):
                 return f"barrier {qubits};"
             return "barrier;"
         return super().to_ir(target, ir_type, serialization_properties, **kwargs)
-
-    def _to_openqasm(self) -> str:
-        return "barrier"
--- a/src/braket/circuits/compiler_directives.py
+++ b/src/braket/circuits/compiler_directives.py
@@ -60,3 +60,42 @@ class EndVerbatimBox(CompilerDirective):
 
     def _to_openqasm(self) -> str:
         return "}"
+
+
+class Barrier(CompilerDirective):
+    r"""Barrier compiler directive."""
+
+    def __init__(self, qubit_indices: list[int]):
+        super().__init__(["||"])
+        self._qubit_indices = qubit_indices
+
+    @property
+    def qubit_indices(self) -> list[int]:
+        return self._qubit_indices
+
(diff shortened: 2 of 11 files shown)


src/braket/circuits/compiler_directives.py
```diff
@@ -60,3 +62,39 @@ def _to_jaqcd(self, *args, **kwargs) -> Any:
...
+        **kwargs,
+    ) -> Any:
+        if ir_type.name == "OPENQASM":
+            if target:
+                qubits = ", ".join(serialization_properties.format_target(int(q)) for q in target)
+                return f"barrier {qubits};"
+            return "barrier;"
+        return super().to_ir(target, ir_type, serialization_properties, **kwargs)
+
+    def _to_openqasm(self) -> str:
```
yitchen-tim: This will never be reached unless called directly. But when called directly, it's likely incorrect (should be `barrier;`)

I think what we should do is either
- Change `_to_openqasm` of the parent class to accept `args` and `kwargs`. Since it's a private function, it's okay to change it as long as it's consistent in the code base. Then, overwrite only `_to_openqasm` in this class.
- Overwrite only `to_ir` and don't implement `_to_openqasm`.

src/braket/circuits/moments.py
```diff
@@ -174,8 +174,12 @@ def add(self, instructions: Iterable[Instruction] | Instruction, noise_index: in
     def _add(self, instruction: Instruction, noise_index: int = 0) -> None:
         operator = instruction.operator
         if isinstance(operator, CompilerDirective):
-            time = self._update_qubit_times(self._qubits)
-            self._moments[MomentsKey(time, None, MomentType.COMPILER_DIRECTIVE, 0)] = instruction
+            qubit_range = instruction.target.union(instruction.control or QubitSet())
```
yitchen-tim, line 177: Why `instruction.control` not `instruction.qubit_indices` as defined in the Barrier class?

src/braket/circuits/text_diagram_builders/ascii_circuit_diagram.py
```diff
@@ -70,6 +70,19 @@ def _duplicate_time_at_bottom(cls, lines: str) -> None:
...
+    @classmethod
+    def _get_compiler_directive_symbols(
+        cls, item: Instruction, circuit_qubits: QubitSet
+    ) -> list[str]:
+        """Get symbols for compiler directives, handling barriers specially."""
+        if item.operator.name == "Barrier":
+            return item.ascii_symbols
+        ascii_symbol = item.ascii_symbols[0]
+        marker = "*" * len(ascii_symbol)
+        num_after = len(circuit_qubits) - 1
```
yitchen-tim: This should be incorrect for barrier. This will plot barrier on all qubits, not just the ones barrier is applied to.

src/braket/circuits/text_diagram_builders/ascii_circuit_diagram.py
```diff
@@ -102,11 +115,7 @@ def _create_diagram_column(
                 control_qubits = QubitSet()
                 target_and_control = target_qubits.union(control_qubits)
                 qubits = circuit_qubits
-                ascii_symbol = item.ascii_symbols[0]
-                marker = "*" * len(ascii_symbol)
-                num_after = len(circuit_qubits) - 1
-                after = ["|"] * (num_after - 1) + ([marker] if num_after else [])
-                ascii_symbols = [ascii_symbol, *after]
+                ascii_symbols = cls._get_compiler_directive_symbols(item, circuit_qubits)
```
yitchen-tim: Same comment as the previous one

test/unit_tests/braket/circuits/test_ascii_circuit_diagram.py
```diff
@@ -956,3 +956,33 @@ def test_measure_with_readout_noise():
...
+        "        |  ",
+        "q100 : -||-",
+        "",
+        "T    : |0 |",
+    )
+    _assert_correct_diagram(circ, expected)
+
+
+def test_barrier_circuit_visualization_with_other_gates():
+    circ = Circuit().x(0).barrier(target=[0, 100]).h(3)
```
yitchen-tim, line 1013: It's a bit tricky. How do we plot barrier on q and q100 but not on q3, any thought?

src/braket/circuits/circuit.py
```diff
@@ -729,6 +729,29 @@ def add_verbatim_box(
...
+        Returns:
+            Circuit: self
+
+        Examples:
+            >>> circ = Circuit().h(0).barrier([0, 1]).cnot(0, 1)
+            >>> circ = Circuit().h(0).h(1).barrier()  # barrier on all qubits
+        """
+        target_qubits = self.qubits if target is None else QubitSet(target)
+
+        if target_qubits:
```
laurencap, line 748: because `self.qubits` can be None?

src/braket/circuits/compiler_directives.py
```diff
@@ -60,3 +62,39 @@ def _to_jaqcd(self, *args, **kwargs) -> Any:
 
     def _to_openqasm(self) -> str:
         return "}"
+
+
+class Barrier(CompilerDirective):
+    r"""Barrier compiler directive."""
```
laurencap: Nit: I don't think `r` is needed here

src/braket/circuits/compiler_directives.py
laurencap: +1 I like the second option, but I haven't seen `_to_openqasm` used before so I am missing some context around why it exists

nilsquet: Thanks a lot for the feedback. I implemented the second option now.

test/unit_tests/braket/circuits/test_ascii_circuit_diagram.py
nilsquet: Good catch. I adjusted the implementation now, such that the barrier does not affect `q3`, since it is none of the target qubits. However, this caused quiet some modifications in different places.

src/braket/circuits/compiler_directives.py
nilsquet: I removed it.

src/braket/circuits/circuit.py
nilsquet: The motivation was to not have any effect when adding a barrier to an empty circuit.

src/braket/circuits/moments.py
nilsquet: Because there could be other `CompilerDirective` objects that are not `Barrier` and therefore don't have the `qubit_indices` attribute.

src/braket/circuits/text_diagram_builders/ascii_circuit_diagram.py
nilsquet: Since I re-worked the plotting part, this behavior has been adapted as well.

src/braket/circuits/text_diagram_builders/ascii_circuit_diagram.py
nilsquet: Since I re-worked the plotting part, this behavior has been adapted as well.

nilsquet:
Since the linter was not happy anymore with the added complexity introduced here, I had to refactor the `ascii_circuit_diagram.py` and `text_circuit_diagram.py` it bit.

src/braket/circuits/moments.py
yitchen-tim: okay thanks, a different question:
Based on line 750 of circuit.py (in `add_barrier()` code)
```
Instruction(compiler_directives.Barrier(list(target_qubits)), target=target_qubits)
```
shouldn't it be `instruction.target` here instead of `instruction.control`?

nilsquet: No, the motivation for this statement was to not be too Barrier-specific but cover further CompilerDirectes as well. In the barrier case, all affected qubits are in `instruction.target`, the `instruction.control` will be `None` and, therefore, `qubit_range` is set to `instruction.target`. 

The reason to have `instruction.control or QubitSet()` in there is to cover the cases in which `instruction.control` is None and to avoid any RuntimeError, we just union with an empty qubit set.

yitchen-tim: got it thanks!

yitchen-tim (approved):

PR #1163, opened 2025-11-12
feature: Emulator support for ProgramSets, transformation passes 

sesmart:
**Issue #, if available:**

Emulator currently does not support ProgramSet, despite braket_dm supporting ProgramSet. Complicating this, there also are discrepancies between the Validation passes and the supported SpecificationTypes, as only Circuit is currently supported.

**Description of changes:**

Added support for ProgramSet in the LocalEmulator, and also clarified emulator support for different specifications. 

- Updated EmulatorProperties
- Updated ValidationPasses
- Updated TransformationPasses
- Updated Emulator

**More details:**

Simpler modifications for just ProgramSets would make longer term support and validation of other specifications harder. Specification-specific pass sequences in the LocalEmulator (i.e. for circuit, ProgramSet, etc.) also were considered, but would be more bloated. There also is substantial reuse between certain passes.

EmulatorPropertes
- EmulatorProperties has two more optional attributes, supportedSpecification and supportedActions
  - supportedSpecification defaults to Circuit
  - supportedActions defaults to empty dict, only used by program set validator (could also be used for a shot validator) 
 
ValidationPasses:
- Each ValidationPass also now has a _supported_specifications attribute, which dictated which TaskSpecification it can run. 
  -  PassManager will check if the current instance is supported by the pass, and otherwise ignore it 
- Added two Validations (generic, not just for circuits)
  - SpecificationValidator - checks whether the specification is supported by emulator 
  - ProgramSetValidator - checks whether or not the program set is valid based on MaximumExecutables and TotalShots
  - NotImplementedValidator - program sets supported, but other task types are not
- Relevant circuit_passes validators now support Circuit and ProgramSet specifications

TransformationPasses:
- Analogous to ValidatorPasses, these are modifying passes that return a specification instead of raising exceptions
- Added Transformations for circuit | programset:
  - VerbatimTransformation - Removes Verbatim from the circuit 
  - NoiseModelTransformation - applies a NoiseModel 
  - MeasurementTransformation- appends measurements (to programs) if none detected; leaves programsets alone

Emulator:
- Changed noise model loading from backend to a pass

*Testing done:*

Added relevant tests. 

## Merge Checklist

#### General

#### Tests


src/braket/emulation/passes/modifier_pass.py
```diff
@@ -0,0 +1,65 @@
...
+# language governing permissions and limitations under the License.
+
+from __future__ import annotations
+
+from abc import ABC, abstractmethod
+
+from braket.tasks.quantum_task import TaskSpecification
+
+
+class ModifierPass(ABC):
```
speller26: `TransformationPass`? We already use the term "modifier" to refer to gate modifiers

sesmart: With instances like "VerbatimTransformer" or "VerbatimTransformation" (favored)?

sesmart: okay, updated

src/_test_emulator.py
maolinml, line 1: Why do we have this file here?

src/braket/emulation/device_emulator_properties.py
```diff
@@ -42,6 +55,7 @@ class DeviceEmulatorProperties:
         twoQubitProperties (dict[str, TwoQubitProperties]): Properties of two-qubit calibration
             details
         supportedResultTypes (list[ResultType]): List of supported result types.
+        supportedActions (list[])
```
maolinml: The docstring for the two added items need to be added here

src/braket/emulation/device_emulator_properties.py
```diff
@@ -207,6 +237,13 @@ def from_device_properties(
                 "The action in device_properties_json must have key `braket.ir.openqasm.program`."
             )
 
+        supportedSpecifications = []
+        for action in properties_dict["action"]:
+            supportedSpecifications += ACTION_TO_SPECIFICATION[action]
+        supportedSpecifications = tuple(supportedSpecifications)
+
+        supportedActions = properties_dict["action"]
```
maolinml: Looking at these few lines makes me wonder why we need both "supportedActions" and "supportedSpecifications" in "DeviceEmulatorProperties". It seems like they have the same information but in different format, is it?

src/braket/emulation/device_emulator_properties.py
```diff
@@ -182,6 +204,14 @@ def two_qubit_properties(self) -> dict[str, TwoQubitProperties]:
     def supported_result_types(self) -> list[ResultType]:
         return self._supported_result_types
 
+    @property
+    def supported_specification(self) -> tuple[BraketSchemaBase] | BraketSchemaBase:
```
maolinml: nit: maybe call it "supported_specification**s**" to be consistent?

src/braket/emulation/emulator.py
```diff
@@ -128,23 +129,12 @@ def transform(
...
-
-        return (
-            self._noise_model.apply(program) if apply_noise_model and self.noise_model else program
-        )
+        modifier = PassManager([MeasurementTransformation()])
+        if apply_noise_model:
+            modifier.append(NoiseModelTransformation(noise_model=self.noise_model))
+        return modifier.transform(program)
 
     def _remove_verbatim_box(self, noisy_verbatim_circ: Circuit) -> Circuit:
```
maolinml: This "_remove_verbatim_box" needs to be removed since it is not used anymore

src/braket/emulation/emulator.py
```diff
@@ -79,11 +82,9 @@ def run(
             QuantumTask: The QuantumTask tracking task execution on this device emulator.
         """
 
-        task_specification = self.transform(task_specification, apply_noise_model=False)
-        # Don't apply noise model as the local simulator will automatically apply it.
-
+        task_specification = self.transform(task_specification, apply_noise_model=True)
         # Remove the verbatim box before submitting to the braket density matrix simulator
-        task_specification_v2 = self._remove_verbatim_box(task_specification)
+        task_specification_v2 = VerbatimTransformation().run(task_specification)
```
maolinml: Can "VerbatimTransformation()" be done as part of the "self.transform" above, meaning we append the "VerbatimTransformation()" into the modifier? See below.

src/braket/emulation/emulator.py
```diff
@@ -128,23 +129,12 @@ def transform(
...
-        if (not has_measurement) and len(task_specification.result_types) == 0:
-            task_specification.measure(target_qubits=task_specification.qubits)
-
-        return (
-            self._noise_model.apply(program) if apply_noise_model and self.noise_model else program
-        )
+        modifier = PassManager([MeasurementTransformation()])
+        if apply_noise_model:
+            modifier.append(NoiseModelTransformation(noise_model=self.noise_model))
+        return modifier.transform(program)
```
maolinml: Can we add "VerbatimTransformation()" in front of this line, so that "modifier.transform(program)" removes the verbatim box?

src/braket/emulation/emulator.py
```diff
@@ -128,23 +129,12 @@ def transform(
...
-        has_measurement = any(
-            isinstance(instr.operator, Measure) for instr in task_specification.instructions
-        )
-        if (not has_measurement) and len(task_specification.result_types) == 0:
-            task_specification.measure(target_qubits=task_specification.qubits)
-
-        return (
-            self._noise_model.apply(program) if apply_noise_model and self.noise_model else program
-        )
+        modifier = PassManager([MeasurementTransformation()])
```
maolinml: Can we add "VerbatimTransformation()" here? Note that the order is important [at least in the previous implementation] where we want to add measurement first and then remove the verbatim box.

src/braket/emulation/emulator.py
```diff
@@ -79,11 +82,9 @@ def run(
             QuantumTask: The QuantumTask tracking task execution on this device emulator.
         """
 
-        task_specification = self.transform(task_specification, apply_noise_model=False)
-        # Don't apply noise model as the local simulator will automatically apply it.
-
+        task_specification = self.transform(task_specification, apply_noise_model=True)
```
maolinml, line 88: Just to confirm my understanding: With "apply_noise_model=True", we apply the noise model to the circuits, hence we don't want to add the noise model to the local sim. This is a different implementation can the old one, but it should work for circuit too.

maolinml: I am worrying that if a customer supply a "backend" which is a local dm with noise model, then we are actually applying the noise model twice. Shall we add a warning for that [this question applies to previous implementation too i think]?

src/braket/emulation/local_emulator.py
```diff
@@ -93,9 +95,10 @@ def from_device_properties(
                 device_em_properties.supported_result_types,
                 device_em_properties.connectivity_graph,
             ),
+            ProgramSetValidator(device_em_properties.supported_actions),
         ]
 
-        local_backend = LocalSimulator(backend=backend, noise_model=noise_model)
+        local_backend = LocalSimulator(backend=backend)
```
maolinml, line 101: This is consistent with the fact that now we apply the noise model to the circuits, hence there shouldn't be noise model in the local sim. This is good.

src/braket/emulation/local_emulator.py
```diff
@@ -93,9 +95,10 @@ def from_device_properties(
                 device_em_properties.supported_result_types,
                 device_em_properties.connectivity_graph,
             ),
+            ProgramSetValidator(device_em_properties.supported_actions),
```
maolinml, line 98: Per the comment for DeviceEmulatorProperties, shall we use "device_em_properties. supported_specification" [or device_em_properties.supported_specification] for both SpecificationValidator and ProgramSetValidator?

src/braket/emulation/passes/circuit_passes/result_type_validator.py
```diff
@@ -51,6 +51,7 @@ def __init__(
         }
 
         self._connectivity_graph = connectivity_graph
+        self._supported_specifications = Circuit
```
maolinml, line 54: we don't need to change result type validator for program set?

src/braket/emulation/device_emulator_properties.py
maolinml: One idea: Can we instantiate DeviceEmulatorProperties with just "supportedActions", and define "self. supported_specifications" based on it?

src/braket/emulation/passes/generic/program_set_validator.py
```diff
@@ -0,0 +1,59 @@
...
+
+        Raises:
+            ValueError: The task specification is not supported.
+        """
+        self.device_actions = device_action
+        self._supported_specifications = ProgramSet
+
+    def validate(self, task_specification: ProgramSet) -> None:
+        """
+        Checks that the number of qubits used in this circuit does not exceed this
```
maolinml: The docstring does not seem to agree with that it is actually doing below. So do we want to validate max qubits?

src/braket/emulation/passes/circuit_passes/not_implemented_validator.py
```diff
@@ -37,10 +46,22 @@ def validate(self, program: TaskSpecification) -> None:
...
+
+        # Validate out unsupported specifications
+        if isinstance(program, unsupported_specifications):
+            raise ValueError(f"Specification {type(program)} is not supported yet.")  # noqa: TRY004
 
+        if isinstance(program, ProgramSet):
+            for item in program:
+                self.validate(item)
+            return
         # Check if the program has a verbatim box when required
```
maolinml, line 54: Shall we check that all the circuits in the program set are verbatim circuits?

src/_test_emulator.py
sesmart: Huh. I was pretty sure this file was deleted.

src/braket/emulation/device_emulator_properties.py
sesmart: Good question. The actions that I found supported in the devices.properties.actions were OpenQASM Programs, OpenQASM  ProgramSets, AHS, and JACQD (not a supported task specification). I guess it is only Pulse that is not there - for Rigetti all I found was an attribute in the device.properties.

src/braket/emulation/emulator.py
sesmart: Am okay to remove it, and figure there won't be any backward compatibility issues as it is a private method (one test that directly tests it), but just left it there just in case.

src/braket/emulation/emulator.py
sesmart: Main motivation here is that NoiseModel.transform only supports Circuits, so the local simulator with a noise model is not smart enough to apply it to elements of the program set, even though the simulator can parse program sets.

sesmart: Yes, I think we certainly could raise a warning checking that the device noise model is None.

src/braket/emulation/passes/circuit_passes/result_type_validator.py
sesmart: No, ProgramSets on construction will validate that there are no results type present. Previously I had it but realized you never got to the ResultTypeValidator so you can't test it.

src/braket/emulation/passes/generic/program_set_validator.py
sesmart: Good catch, updated.

src/braket/emulation/passes/circuit_passes/not_implemented_validator.py
sesmart: Yep! I believe that is what is being performed here, though we could actually do check Verbatim as a validation pass as well?

src/braket/emulation/device_emulator_properties.py
sesmart: (see also two other comments I believe)

src/braket/emulation/emulator.py
sesmart: It would be nice, but because if we ran transform alone, we would not need to remove Verbatim, it may make more sense only to have it in the run function.

src/braket/emulation/emulator.py
maolinml: Yeah, let's remove this function and the unit test associated to it.

src/braket/emulation/emulator.py
maolinml: oh now i understand your question if we should have a "remove_verbatim" flag. Well in this case, I would prefer to add the flag in the "transform" function which is default to be False. But I will let you decide if you want to make the change.

src/braket/emulation/device_emulator_properties.py
maolinml: I am not sure what is the best option here: keep both or use one of "supportedActions" and "supportedSpecifications" when defining "DeviceEmulatorProperties". Maybe @speller26 can chime in and give some thought?

src/braket/emulation/device_emulator_properties.py
maolinml: Let's finalize this after we decide if we want both "supportedActions" and "supportedSpecifications" or only one of them.

maolinml: Reminder to a) complete this docstring and b) add the docstring for "supportedSpecifications"

src/braket/emulation/device_emulator_properties.py
maolinml: I am still not sure if it is a good idea to have both "supportedActions" and "supportedSpecifications" as inputs for "DeviceEmulatorProperties". What happen if there is conflict between the inputs? For example, there is AHS in the supported action whereas there is only program set in supported specification?

src/braket/emulation/pass_manager.py
```diff
@@ -68,3 +68,6 @@ def _raise_exception(self, exception: Exception) -> None:
             exception (Exception): The exception to modify and raise.
         """
         raise EmulatorValidationError(str(exception)) from exception
+
+    def append(self, single_pass: TransformationPass | ValidationPass) -> None:
```
speller26: If we have this, how about implementing `__add__` and `__iadd__` too?

src/braket/emulation/passes/circuit_passes/measurement_transformation.py
```diff
@@ -0,0 +1,75 @@
...
+            Modified circuit(s) with measurements added where needed
+        """
+        if isinstance(circuits, ProgramSet):
+            new_programs = []
+            for program in circuits:
+                match program:
+                    case Circuit():
+                        new_programs.append(self.transform(program))
+                    case CircuitBinding() as entry if entry.observables:
+                        new_programs.append(entry)
```
speller26:
```suggestion
                    case CircuitBinding(observables=observables) if observables:
                        new_programs.append(program)
```

src/braket/emulation/passes/circuit_passes/noise_model_transformation.py
```diff
@@ -0,0 +1,61 @@
...
+        """Apply noise model to circuits.
+
+        Args:
+            circuits: Circuit or ProgramSet to add noise to
+
+        Returns:
+            Circuit(s) with noise model applied
+        """
+        if self._noise_model is None:
+            return circuits
```
speller26:
```suggestion
        if not self._noise_model:
            return circuits
```

src/braket/emulation/passes/circuit_passes/not_implemented_validator.py
```diff
@@ -37,10 +46,22 @@ def validate(self, program: TaskSpecification) -> None:
...
+            OpenQASMProgram
+            | SerializableProgram
+            | AnalogHamiltonianSimulation
+            | OpenQASMProgramSet
+            | PulseSequence
+        )
+
+        # Validate out unsupported specifications
+        if isinstance(program, unsupported_specifications):
+            raise ValueError(f"Specification {type(program)} is not supported yet.")  # noqa: TRY004
```
speller26: Wouldn't it be easier to maintain the list of _supported_ specs? New specs are unsupported by default, and we'd have to add to this list every time.

src/braket/emulation/passes/circuit_passes/verbatim_transformation.py
```diff
@@ -0,0 +1,76 @@
...
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+
+from braket.circuits import Circuit
+from braket.circuits.compiler_directives import EndVerbatimBox, StartVerbatimBox
+from braket.emulation.passes import TransformationPass
+from braket.program_sets import ProgramSet
+
+
+class VerbatimTransformation(TransformationPass):
```
speller26: I feel like this name is a misnomer; it suggests that it _makes_ a circuit verbatim.

speller26 (commented):
Still reviewing

src/braket/emulation/pass_manager.py
sesmart: Done.

src/braket/emulation/passes/circuit_passes/not_implemented_validator.py
sesmart: Fair point, yep!

src/braket/emulation/passes/circuit_passes/verbatim_transformation.py
sesmart: changed to RemoveVerbatimTransformation, which is long, but accurate

src/braket/emulation/passes/circuit_passes/measurement_transformation.py
```diff
@@ -0,0 +1,75 @@
...
+                        )
+                    case _:
+                        raise NotImplementedError
+            return ProgramSet(new_programs, shots_per_executable=circuits.shots_per_executable)
+
+        has_measurement = any(
+            isinstance(instr.operator, Measure) for instr in circuits.instructions
+        )
+        if (not has_measurement) and len(circuits.result_types) == 0:
+            circuits.measure(target_qubits=sorted(circuits.qubits))
```
speller26: No longer needs the explicit `sort`, since `Circuit.qubits` is now sorted if qubits are virtual.

src/braket/emulation/device_emulator_properties.py
speller26: Another reminder :)

src/braket/emulation/device_emulator_properties.py
```diff
@@ -75,6 +91,12 @@ def __init__(
         self._one_qubit_properties = oneQubitProperties
         self._two_qubit_properties = twoQubitProperties
         self._supported_result_types = supportedResultTypes
+        if not supportedSpecifications:
+            supportedSpecifications = Circuit
+        self._supported_specifications = supportedSpecifications
+        if not supportedActions:
+            supportedActions = {}
+        self._supported_actions = supportedActions
```
speller26: Why not just ternary
```suggestion
        self._supported_specifications = supportedSpecifications or Circuit
        self._supported_actions = supportedActions or {}
```

src/braket/emulation/device_emulator_properties.py
speller26: supportedSpecifications is probably all we need for emulation, as action type is a schema construct, and the emulators don't modify things that have already been serialized to IR.

src/braket/emulation/emulator.py
```diff
@@ -79,11 +83,13 @@ def run(
...
         # Remove the verbatim box before submitting to the braket density matrix simulator
-        task_specification_v2 = self._remove_verbatim_box(task_specification)
+        task_specification_v2 = RemoveVerbatimTransformation().run(task_specification)
+
+        if hasattr(self._backend, "_noise_model") and self._backend._noise_model is not None:
+            warnings.warn("Backend device already has a noise model defined.", stacklevel=2)
+
         return self._backend.run(task_specification_v2, shots, inputs, *args, **kwargs)
```
speller26: Since this is only used once, how about just passing `RemoveVerbatimTransformation().run(task_specification)` directly?
```suggestion
        if hasattr(self._backend, "_noise_model") and self._backend._noise_model is not None:
            warnings.warn("Backend device already has a noise model defined.", stacklevel=2)

return self._backend.run(RemoveVerbatimTransformation().run(task_specification), shots, inputs, *args, **kwargs)
```
(remember to move the preceding comment to right before the `.run` call; I can't include it in the suggestion for some reason.)

speller26 (commented):
Plus Mao's other unaddressed comments

src/braket/emulation/passes/circuit_passes/measurement_transformation.py
sesmart: ^ exactly related to our conversation actually - it was one of the few places that would need Circuit.qubits - though actually, the inputs should be Verbatim, so I need to take another look at this.

src/braket/emulation/device_emulator_properties.py
sesmart: Updated.

src/braket/emulation/passes/circuit_passes/measurement_transformation.py
sesmart: As emulator only supports verbatim atm, and measure is applied before verbatim is removed, this should work okay - I think it will not sort generally?

src/braket/emulation/emulator.py
sesmart: Done.

src/braket/emulation/passes/validation_pass.py
speller26, line 1: Deleting this is a breaking change; you'll want to at least leave an import here.

src/braket/emulation/passes/passes.py
```diff
@@ -0,0 +1,102 @@
...
+# language governing permissions and limitations under the License.
+
+from __future__ import annotations
+
+from abc import ABC, abstractmethod
+
+from braket.tasks.quantum_task import TaskSpecification
+
+
+class _EmulatorPass(ABC):
```
speller26: What makes this emulator-specific?

src/braket/emulation/passes/passes.py
```diff
@@ -0,0 +1,102 @@
...
+    @abstractmethod
+    def run(self, task_specification: TaskSpecification) -> TaskSpecification:
+        raise NotImplementedError
+
+    def __call__(self, task_specification: TaskSpecification) -> TaskSpecification:
+        return self.run(task_specification)
+
+    @property
+    def supported_specifications(self) -> TaskSpecification:
```
speller26, line 30: Makes me wonder if there should be a TaskSpecification superclass with two methods: `to_ir` and `ir_types` that respectively serialize the specification and tell the user the available IR types; this could close that conceptual gap between spec and IR formats.

src/braket/emulation/passes/generic/program_set_validator.py
```diff
@@ -0,0 +1,60 @@
...
+            pset_action = pset_action.dict()
+        max_shots = pset_action["maximumTotalShots"]
+        max_exc = pset_action["maximumExecutables"]
+        if len(task_specification) > max_exc:
+            raise ValueError(
+                f"{len(task_specification)} is greater than "
+                f"the supported number of executables {max_exc}."
+            )
+
+        if len(task_specification) * task_specification.shots_per_executable > max_shots:
```
speller26: Use [total_shots](https://github.com/amazon-braket/amazon-braket-sdk-python/blob/811a8b6e7e449bedbcc282cdf098255a753422d0/src/braket/program_sets/program_set.py#L91) instead
```suggestion
        if task_specification.total_shots > max_shots:
```

src/braket/emulation/local_emulator.py
sesmart: I think it is okay as is - ProgramSet validator definitely needs the device action, not the specification.

src/braket/emulation/passes/validation_pass.py
sesmart: Added it back and imported ValidationPass

src/braket/emulation/passes/circuit_passes/result_type_validator.py
speller26, line 54: Should this be a list?

speller26 (approved):

PR #1172, opened 2025-11-29
fix: Increment classical bits when adding circuits

### LINKED ISSUE #1175: Bug in Measure.to_ir
**Describe the bug**

Double star in [Measure().to_ir()](https://github.com/amazon-braket/amazon-braket-sdk-python/blob/48f1dd3fbce7026180cfb4bce2f29197f6ff9495/src/braket/circuits/measure.py#L72) actually performs pow. 

Simple fix but don't actually think kwargs can be supplied to _to_openqasm...maybe to OpenQASMSerializationProperties?

**To reproduce**

`Measure().to_ir()`

**System information**
A description of your system. Please provide:
- **Amazon Braket Python SDK version**: 1.104.1
- **Amazon Braket Python Schemas version**: 1.26.1
- **Amazon Braket Python Default Simulator version**: 1.32.0
- **Python version**: 3.12

speller26:
Classical bits are actually stored in the `Measure` instruction rather than the circuit,  and `Circuit.measure` has no way to supply classical bits, taking only a qubit target. Each call to `measure` creates a `Measure` instruction with new classical bits, completely independent of the indices of the qubits; the classical bit indices are contiguous integers with the first measurement's starting from 0, and those of each subsequent measurement starting from [the current number of measured qubits in the circuit](https://github.com/amazon-braket/amazon-braket-sdk-python/blob/a76530ca3204e35453ea12c2bfd7541187febf01/src/braket/circuits/circuit.py#L762-L769).

*Issue #, if available:*

Also closes #1175 

*Description of changes:*

*Testing done:*

## Merge Checklist

#### General

#### Tests

### DIFF
--- a/src/braket/circuits/measure.py
+++ b/src/braket/circuits/measure.py
@@ -69,7 +69,7 @@ class Measure(QuantumOperator):
             return self._to_jaqcd()
         if ir_type == IRType.OPENQASM:
             return self._to_openqasm(
-                target, serialization_properties or OpenQASMSerializationProperties(), **kwargs
+                target, serialization_properties or OpenQASMSerializationProperties()
             )
         raise ValueError(f"supplied ir_type {ir_type} is not supported.")
 


test/unit_tests/braket/circuits/test_circuit.py
```diff
@@ -845,14 +847,24 @@ def test_measure_add_circuit_target_mapping():
...
 
 
+def test_add_circuit_with_measure():
+    circ1 = Circuit()
+    circ2 = Circuit()
+    for i in range(6):
+        circ1 += Circuit().h(i) + Circuit().measure(i)
+        circ2.h(i).measure(i)
+    assert circ1 == circ2
+    assert circ1.to_ir(IRType.OPENQASM) == circ2.to_ir(IRType.OPENQASM)
```
speller26, line 865: This is the only new test

src/braket/circuits/circuit.py
sesmart, line 540: Does this pattern need to change as well? 

Or rather, could the information be located here instead of in add_circuit as add_circuit could just call add_instruction?

src/braket/circuits/circuit.py
```diff
@@ -809,8 +797,19 @@ def measure(self, target_qubits: QubitSetInput) -> Circuit:
                 f"cannot repeat qubit(s) {', '.join(map(str, intersection))} "
                 "in the same measurement."
             )
-        self._add_measure(target_qubits=target_qubits)
 
+        for idx, target in enumerate(target_qubits):
```
sesmart, line 801: This is just removing the _add_measure subfunction?

sesmart: And if I understand correctly, if I piece together multiple measure instructions, the resulting classical bits should always be in the order in which measure is applied?

speller26: Yeah, just doing some consolidation, since `_add_measure` isn't used anywhere else, and `measure` is already short enough. And yes, the order will be in the order that measurements are applied.

src/braket/circuits/circuit.py
speller26: That makes sense too; I'll do that.

speller26: Ah, I remember why I did it this way; this is the bottom level that is called by `.measure` itself. I could untangle this a bit, although that may a be a bit trickier.

speller26: One thing that complicates this is that users may want to add a `Measure` instruction with a explicit classical bits, which would be overriden if the change is made here. Of course, this would also affect adding circuits with such instructions, but since adding circuits already doesn't work, things wouldn't really change for that subset.

The correct approach is to implement classical bits properly (for example, at least being able to access them in the circuit, and make them instruction targets), but that should be done in a dedicated feature.

sesmart: Okay, I think that makes sense. Although technically adding circuits with user-specified instructions / bits currently does work, and it would no longer work. It's adding circuits without user-specified bits that does not work. 

Kind of a funny case. Would be nice if there was a flag for user-assigned indexes, or if they defaulted to None from .measure() and were just assigned contiguously. 

Feel free to resolve though.

PR #1183, opened 2025-12-15
feature: update device.run to support experimentalCapabilities flag

Altanali:
*Issue #, if available:*

*Description of changes:*

Updating `AwsQuantumTask` and `device.run(...)` to support the new `experimentalCapabilities` argument in `Braket:CreateQuantumTask`. We introduce the `enabled_experimental_capabilities` argument which currently only accepts `["ALL"]`` as a valid input and adds 
```
{ 
 "experimentalCapabilities:: {
    "enabled": "ALL"
    }
}
```
to the quantum task arguments. 

Additionally, update the `EnableExperimentalCapabilities` context manager to support taking in a list of experimental capabilities. 

*Testing done:*
 - Added unittests for the context manager.
 - Updated unit tests for `AwsDevice`, `AwsQuantumTask`, `AwsQuantumTaskBatch`, and test utils to support the `enabled_experimental_capabilities` argument. 

## Merge Checklist

#### General

#### Tests

### DIFF
--- a/src/braket/experimental_capabilities/experimental_capability_context.py
+++ b/src/braket/experimental_capabilities/experimental_capability_context.py
@@ -45,8 +45,8 @@ class GlobalExperimentalCapabilityContext:
         """Enable all experimental capabilities. Default behavior is to enable all.
 
         Args:
-            experimental_capabilities (ExperimentalCapabilitiesEnabled | None): The experimental capabilities to enable.
-                Defaults to all capabilities.
+            experimental_capabilities (ExperimentalCapabilitiesEnabled | None):
+            The experimental capabilities to enable. Defaults to all capabilities.
         Raises:
             ExperimentalCapabilityContextError: If the experimental capabilities are not valid.
         """
@@ -63,7 +63,8 @@ class GlobalExperimentalCapabilityContext:
     def get_enabled_capabilities(self) -> ExperimentalCapabilitiesEnabled | None:
         """Get the set of currently enabled experimental capabilities.
         Returns:
-            ExperimentalCapabilitiesEnabled | None: The currently enabled experimental capabilities.
+            ExperimentalCapabilitiesEnabled | None: The currently enabled
+                experimental capabilities.
         """
         if self._all_enabled:
             return EXPERIMENTAL_CAPABILITIES_ALL
@@ -72,7 +73,8 @@ class GlobalExperimentalCapabilityContext:
     def set_state(self, experimental_capabilities: ExperimentalCapabilitiesEnabled | None) -> None:
         """Set the state of experimental capabilities.
         Args:
-            experimental_capabilities (ExperimentalCapabilitiesEnabled | None): The state of enabled experimental capabilities.
+            experimental_capabilities (ExperimentalCapabilitiesEnabled | None):
+                The state of enabled experimental capabilities.
         """
         if experimental_capabilities == EXPERIMENTAL_CAPABILITIES_ALL:
             self.enable()
@@ -101,8 +103,8 @@ class EnableExperimentalCapability:
         [1] https://docs.aws.amazon.com/braket/latest/developerguide/
         braket-experimental-capabilities.html
 
-        Args: experimental_capabilities (ExperimentalCapabilitiesEnabled | None): The experimental capabilities to enable.
-                Defaults to all capabilities.
+        Args: experimental_capabilities (ExperimentalCapabilitiesEnabled | None):
+            The experimental capabilities to enable. Defaults to all capabilities.
 
         Examples:
             >>> with EnableExperimentalCapability():
--- a/examples/debug_bell.py
+++ b/examples/debug_bell.py
@@ -27,8 +27,7 @@ device = AwsDevice(Devices.Amazon.SV1)
 bell = Circuit().h(0).cnot(0, 1)
 # pass in logger to device.run, enabling debugging logs to print to console
 logger.info(
-    device
-    .run(
+    device.run(
         bell,
         shots=100,
         poll_timeout_seconds=120,
--- a/src/braket/circuits/gate_calibrations.py
+++ b/src/braket/circuits/gate_calibrations.py
@@ -135,8 +135,7 @@ class GateCalibrations:
                     f"The key {calibration_key} does not exist in this GateCalibrations object."
                 )
             return (
-                self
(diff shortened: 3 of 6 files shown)


src/braket/experimental_capabilities/experimental_capability_context.py
```diff
@@ -24,40 +34,110 @@ class ExperimentalCapabilityContextError(Exception):
...
+        self,
+        experimental_capabilities: Any,
+    ):
+        raise TypeError(
+            "If provided, Experimental capabilities must be a single \
+                or list of ExperimentalCapability strings"
+        )
+
+    @_is_enabled.register
+    def _(self, experimental_capabilities: None):
```
yitchen-tim: This is ambiguous that None implies ALL here. Should we make  `_is_enabled(ALL)` explicit?

src/braket/experimental_capabilities/experimental_capability_context.py
```diff
@@ -24,40 +34,110 @@ class ExperimentalCapabilityContextError(Exception):
...
+    ) -> None:
         """Set the state of all experimental capabilities.
 
         Args:
-            state: The state to set. True to enable all capabilities, False to disable all.
+            enabled_capabilities: the state of the experimental capability context in terms of the
+            experimental capabilities to enable. If None, default to ["ALL"]
         """
-        self._is_enabled = bool(state)
+        self.enable(enabled_capabilities)
```
yitchen-tim: This is not setting the state. If the `_currently_enabled_capabilities={X}` and we `set_state([Y])`, `_currently_enabled_capabilities` becomes `{X, Y}` while we expect `{Y}`. 

This becomes important in nested context where 
```
with EnableExperimentalCapability([A,B]):
    # {A, B} enabled
    with EnableExperimentalCapability([C]):
           # {A, B, C} enabled
    # {A, B} enabled
```

yitchen-tim: Also, what if the previous state is 
```
_all_enabled = True
_currently_enabled_capabilities = {}
```
It won't set `_all_enabled`.

yitchen-tim: Now I see your implementation in `__exit__` below. I feel like we should either (1) fully implement `set_state()` or (2) replace the implementation in `__exit__` with
```
GLOBAL_EXPERIMENTAL_CAPABILITY_CONTEXT.disable(self._current_enabled_capabilities)
GLOBAL_EXPERIMENTAL_CAPABILITY_CONTEXT.enable(self._previous_enabled_capabilities)
```
and remove `def set_state()`.

Altanali: >  becomes {X, Y} while we expect {Y}.

Ah, for this I wasn't sure if the behavior for nested contexts should be to enable a union of the enabled capabilities (which is what I implemented) or to enable the capabilities _in only the most recent context_.

Assuming we're okay with taking the union of capabilities for nested contents, your option (2) makes sense to me (we remove `set_state` and replace `__exit__` with a `disable ==> enable` step)!

src/braket/experimental_capabilities/experimental_capability_context.py
Altanali: That's fair! I'll make `"ALL"` the default for `is_enabled` and raise an error if an explicit None is passed in.

^ Edit: Just to clarify, I agree that `_is_enabled(None) ==> _is_enabled(ALL)` is _wrong_. I'll update it so `is_enabled() ==> _is_enabled(ALL)` and `is_enabled(None) ==> Error`.

yitchen-tim (dismissed):

src/braket/experimental_capabilities/experimental_capability_context.py
yitchen-tim: sounds great, thank you

src/braket/aws/aws_quantum_task.py
```diff
@@ -199,6 +209,13 @@ def create(
                 ]
             })
 
+        if GLOBAL_EXPERIMENTAL_CAPABILITY_CONTEXT.is_enabled() or experimental_capabilities == [
+            ExperimentalCapability.ALL
+        ]:
```
speller26: Why a single-element list rather than just `ExperimentalCapabilities.All`?

Altanali: Hmm, do we maybe want `experimental_capabilities: list[ExperimentalCapability/] | ExperimentalCapability | None`?

The idea behind a list of experimental capabilities was to support changes to CreateQuantumTask where a user can enable specific features instead of the blanket "ALL" that's only supported right now.

speller26: I definitely agree with passing a list of capabilities; I'm just pointing out the `ALL` is a special value that I imagine will _never_ be in a list with other members.

Altanali: Aah, I agree with you! In fact, in the `EnableExperimentalCapability` context manager, supplying `ALL` with any other member will warn the user. 

I'll update this so that supplying `ExperimentalCapabilities.ALL` on its own is (not inside a list) is enough to enable all experimental capabilities.

yitchen-tim:
@speller26 Re: Changing to `ExperimentalCapability` to `ExperimentalCapabilities`, Do you mean including `EnableExperimentalCapability` and `GlobalExperimentalCapabilityContext`? `EnableExperimentalCapability` is public facing one so I am incline not to change it, but the rest I agree. 

For `EnableExperimentalCapability`, we could create an alias `EnableExperimentalCapabilities` and gradually phase out `EnableExperimentalCapability`.

src/braket/aws/aws_quantum_task.py
Altanali: Follow up question: would we still want to support users passing `"ALL"` into a list of capabilities or should we raise an error if a user tries to pass something like `["ALL"]` or `["ALL", "CapA"]` (or just warn and ignore "CapA" in the latter case)

Altanali:
Latest commit:

- Supports "ALL" and ["ALL"] as supported arguments; if "ALL" is in a list with any other capabilities, we'll ignore those and enable all (not really relevant at the moment as there are no other supported experimental capabilities, just pointing out in case that's isn't the behavior we're okay with).
- Changes the `ExperimentalCapability` enum to `ExperimentalCapabilities`.

Issue #1187, opened 2025-12-19
Keep examples in documentation up-to-date automatically


speller26:
**Describe the feature you'd like**

The [list of examples](https://amazon-braket-sdk-python.readthedocs.io/en/latest/index.html#examples) in the documentation is manually constructed, and is prone to going out-of-date with the addition of new notebooks in amazon-braket-examples.

<img width="1669" height="1377" alt="Image" src="https://github.com/user-attachments/assets/4886c6b7-9fc8-463d-a9f9-daa6fcb0d8da" />

This should be updated to automatically pull examples from the notebooks in amazon-braket-examples.

**How would this feature be used? Please describe.**
 
Having the documentation examples update automatically ensures that examples stay relevant and removes the mental load of constantly checking the list to add or remove entries.

**Requirements**

Update the doc generator to use the [entries](https://github.com/amazon-braket/amazon-braket-examples/blob/main/docs/ENTRIES.json) of amazon-braket-examples as the source of truth for the table of contents.

**About unitaryDESIGN 2026**
* This issue is part of unitaryDESIGN 2026; note that you have to be [registered](https://unitary.design/register/) to complete this issue. Learn more about the PR review process [here](https://unitary.design/participant-guide/). 
* A message from unitaryDESIGN about [AI Slop](https://en.wikipedia.org/wiki/AI_slop): "While we are open to collaboration with LLMs for unitaryDESIGN, please note that fully AI-generated PRs are not acceptable. It is up to the discretion of the individual maintainers whether or not LLM-generated PRs are the right fit for their issues, and they have the right to reject those that appear fully AI-generated."
* Learn more about unitaryDESIGN rules and opportunities [here](https://unitary.design/rules/)!

PR #1216, opened 2026-02-15
feat: Dynamically generate example documentation RST files

axif0:
*Issue #1187 , if available:*

*Description of changes:*
Automatically generate example documentation RST files from external data during the Sphinx build process. By running, 
```
python3.13 doc/generate_examples.py
```

*Testing done:*
No testing file edited

## Merge Checklist

#### General

- [X] I have read the [CONTRIBUTING](https://github.com/amazon-braket/amazon-braket-sdk-python/blob/main/CONTRIBUTING.md) doc
- [X] I used the PR title format described in [CONTRIBUTING](https://github.com/amazon-braket/amazon-braket-sdk-python/blob/main/CONTRIBUTING.md#PR-title-format)
- [X] I have updated any necessary documentation, including [READMEs](https://github.com/amazon-braket/amazon-braket-sdk-python/blob/main/README.md) and [API docs](https://github.com/amazon-braket/amazon-braket-sdk-python/blob/main/CONTRIBUTING.md#documentation-guidelines) (if appropriate)

#### Tests

### DIFF
--- a/doc/generate_examples.py
+++ b/doc/generate_examples.py
@@ -1,0 +1,161 @@
+# Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You
+# may not use this file except in compliance with the License. A copy of
+# the License is located at
+#
+#     http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+
+import json
+import os
+import re
+import urllib.request
+from typing import Dict, List, Any
+
+# Map directory names to RST filenames and Headers
+CATEGORY_MAPPING = {
+    "getting_started": {
+        "filename": "examples-getting-started.rst",
+        "header": "Getting started",
+        "description": "Get started on Amazon Braket with some introductory examples."
+    },
+    "braket_features": {
+        "filename": "examples-braket-features.rst",
+        "header": "Amazon Braket features",
+        "description": "Learn more about the individual features of Amazon Braket."
+    },
+    "advanced_circuits_algorithms": {
+        "filename": "examples-adv-circuits-algorithms.rst",
+        "header": "Advanced circuits and algorithms",
+        "description": "Learn more about working with advanced circuits and algorithms."
+    },
+    "hybrid_quantum_algorithms": {
+        "filename": "examples-hybrid-quantum.rst",
+        "header": "Hybrid quantum algorithms",
+        "description": "Learn more about hybrid quantum algorithms."
+    },
+    "pennylane": {
+        "filename": "examples-ml-pennylane.rst",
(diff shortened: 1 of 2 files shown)


doc/generate_examples.py
```diff
@@ -0,0 +1,161 @@
...
+                f.write(link_text + "\n")
+                f.write(underline + "\n\n")
+                
+                if content:
+                    f.write(content + "\n\n")
+
+if __name__ == "__main__":
+    entries = fetch_entries()
+    output_dir = os.path.dirname(os.path.abspath(__file__))
+    generate_rst(entries, output_dir)
```
rmshaffer, line 161: what is the output when you run this? Ideally this should be automatically invoked at build time, and all of the generated .rst files would be picked up by the sphinx build.

doc/generate_examples.py
```diff
@@ -0,0 +1,161 @@
...
+                
+                link_text = f"`{title} <{link}>`_"
+                underline = "*" * len(link_text)
+                
+                f.write(underline + "\n")
+                f.write(link_text + "\n")
+                f.write(underline + "\n\n")
+                
+                if content:
+                    f.write(content + "\n\n")
```
rmshaffer, line 156: what does this content look like? do you have an example?

if the .rst files are generated by the build, then the ReadTheDocs build in the PR will automatically include all of your rendered content.

doc/generate_examples.py
```diff
@@ -0,0 +1,161 @@
...
+# language governing permissions and limitations under the License.
+
+import json
+import os
+import re
+import urllib.request
+from typing import Dict, List, Any
+
+# Map directory names to RST filenames and Headers
+CATEGORY_MAPPING = {
```
rmshaffer, line 21: is there any way to auto-generate this from the entries.json? we want all categories from the json to be present, not limited to just these ones.

Missing from the json is a "friendly title" for each folder; so it may make sense to have some way to specify a custom title here (but with a smart fallback - so that we don't *have* to come update this every time new examples are added).

rmshaffer (commented):
This is a good start! Curious to see what the generated docs end up looking like, and getting this integrated into the sphinx build automatically.

doc/generate_examples.py
sesmart: There is a categories dict in the [_build_body.py](https://github.com/amazon-braket/amazon-braket-examples/blob/a4a16ac2dabf810cc64448d80982f02778111c5d/docs/_build_body.py#L6) file that maps the keyword to those titles....wondering if there is a better way to get that mapping easily. Could be another json file in the examples?

rmshaffer: Maybe, yes. Depends on how we want these docs to be organized. Currently they are organized by folder, which isn't quite 1:1 with category. So yes, possibly another json file in the examples that maps folder to display title?

sesmart: Great point - in the Examples, the folders are kind of all of over the place, and got a bit disorganized over time, but are harder to correct while preserving links, so we prioritized the category over the folder. 

Simulators is a good example, as there is no folder, but it's an important category for the "Test" paradigm.

sesmart: I would lean towards referencing the categories personally.

PR #1249, opened 2026-05-06
feat: Split program sets

speller26:
Introduce `split` method to `ProgramSet` to split program set into a list of program sets each with at most an allowed number of executables.

Also fixed coverage output when running `tox`.

*Issue #, if available:*

*Description of changes:*

*Testing done:*

## Merge Checklist

#### General

#### Tests


src/braket/program_sets/program_set.py
```diff
@@ -97,6 +97,114 @@ def total_shots(self) -> int:
...
+    def split(self, max_executables: int) -> list[ProgramSet]:
+        """
+        Split this program set into a list of program sets with
+        at most ``max_executables`` executables.
+
+        Sum Hamiltonians and lists of observables will not be broken into separate program sets;
+        consequently, this method will fail if the size of any Hamiltonian or observable list
+        exceeds ``max_executables``.
+
+        Adjacent triples originating from the same ``CircuitBinding`` are coalesced into
```
sesmart: Can you clarify what Adjacent means here? Triples isn't really defined before here.

src/braket/program_sets/program_set.py
```diff
@@ -97,6 +97,114 @@ def total_shots(self) -> int:
...
+        a single multi-parameter-set ``CircuitBinding`` in the resulting sub-program set.
+
+        Concatenating the executables of the program sets in the list in order reproduces
```
sesmart: This is very nice :)

src/braket/program_sets/program_set.py
```diff
@@ -97,6 +97,114 @@ def total_shots(self) -> int:
             raise ValueError("No per-executable shots defined")
         return self._shots_per_executable * self.total_executables
 
+    def split(self, max_executables: int) -> list[ProgramSet]:
+        """
+        Split this program set into a list of program sets with
+        at most ``max_executables`` executables.
+
+        Sum Hamiltonians and lists of observables will not be broken into separate program sets;
+        consequently, this method will fail if the size of any Hamiltonian or observable list
```
sesmart: or input set?

src/braket/program_sets/program_set.py
```diff
@@ -97,6 +97,114 @@ def total_shots(self) -> int:
...
+                ``Sum`` Hamiltonian is larger than allowed.
+
+        Examples:
+            >>> ps = ProgramSet([
+            ...     CircuitBinding(c1, inputs1, obs1),  # 100 param sets, 4 observables
+            ...     CircuitBinding(c2, inputs2, obs2),  # 50 param sets, 2 observables
+            ... ])
+            >>> sub = ps.split(120)
+            >>> [s.total_executables for s in sub]
+            [120, 120, 120, 80, 60]
```
sesmart: What dictates the filling here? You have 40 left over from CB1, then it splits the remaining 100 from CB2 as 40/60?

sesmart: Isn't it based on the observable? Here it's only size 2 so it would split 120, 120, 120, 120, 20?

sesmart (commented):
Very cool 😄, just some notes on usability. Any reason the input_set is taken as the major counting index before observables? Just coincidence of orderings?

test/unit_tests/braket/program_sets/test_program_set.py
```diff
@@ -534,3 +534,170 @@ def test_inequality(circuit_rx_parametrized):
...
+def test_split_non_positive_raises(circuit_rx_parametrized):
+    binding = CircuitBinding(circuit_rx_parametrized, input_sets=[{"theta": 1.23}])
+    program_set = ProgramSet(binding)
+    with pytest.raises(ValueError, match="must be positive"):
+        program_set.split(0)
+    with pytest.raises(ValueError, match="must be positive"):
+        program_set.split(-3)
+
+
+def test_split_oversize_class_raises(circuit_rx_parametrized):
```
yitchen-tim: Why can't we split it in this case?

src/braket/program_sets/program_set.py
yitchen-tim: +1 same question

src/braket/program_sets/program_set.py
```diff
@@ -97,6 +97,114 @@ def total_shots(self) -> int:
...
+            if class_size > max_executables:
+                raise ValueError(
+                    f"Program at index {prog_idx} has a single parameter-set index with "
+                    f"{class_size} executables, exceeding max_executables={max_executables}"
+                )
+            input_sets = prog.input_sets
+            if input_sets is None:
+                triples.append((prog_idx, None, class_size))
+            else:
+                triples.extend((prog_idx, i, class_size) for i in range(len(input_sets)))
```
yitchen-tim:
```suggestion
                triples.extend(
                    (prog_idx, input_index, executables_per_input)
                    for input_index in range(len(input_sets))
                )
```
for readability.

src/braket/program_sets/program_set.py
```diff
@@ -97,6 +97,114 @@ def total_shots(self) -> int:
...
+        return program_sets
+
+    def _enumerate_triples(self, max_executables: int) -> list[tuple[int, int | None, int]]:
+        triples = []
+        for prog_idx, prog in enumerate(self._programs):
+            if isinstance(prog, Circuit):
+                triples.append((prog_idx, None, 1))
+                continue
+            obs = prog.observables
+            class_size = max(1, len(obs)) if obs is not None else 1
```
yitchen-tim: what is a `class_size`? Can we call it something like `executables_per_input`?

speller26: Forgot to rename this; the name comes from treating triples as equivalence classes of executables that share an observable list (since we're not directly partitioning the set of executables, but rather the quotient by that equivalence relation).

test/unit_tests/braket/program_sets/test_program_set.py
speller26: Because of my choice to not split observables into different program sets. I can change this, although it'll be a bigger rework.

src/braket/program_sets/program_set.py
speller26: The example used to be different 😂 just forgot to change the output

src/braket/program_sets/program_set.py
speller26: Input sets are split up as necessary

src/braket/program_sets/program_set.py
yitchen-tim: maybe `n_executables_sharing_observables`?

test/unit_tests/braket/program_sets/test_program_set.py
yitchen-tim: Understood thanks for explaining, let's not expanding the scope right here.

speller26:
Results merging has been moved to #1254

src/braket/program_sets/program_set.py
```diff
@@ -97,6 +98,167 @@ def total_shots(self) -> int:
             raise ValueError("No per-executable shots defined")
         return self._shots_per_executable * self.total_executables
 
+    def enumerate_executables(self) -> Iterator[tuple[int, int, int]]:
```
sesmart, line 101: Hm...should this be a property? Like enumeration?

src/braket/program_sets/program_set.py
```diff
@@ -97,6 +98,167 @@ def total_shots(self) -> int:
...
+                continue
+            num_obs = len(prog.observables) if prog.observables is not None else 1
+            for ps_idx in range(len(prog.input_sets) if prog.input_sets is not None else 1):
+                for obs_idx in range(num_obs):
+                    yield binding_idx, ps_idx, obs_idx
+
+    def split(self, max_executables: int) -> tuple[list[ProgramSet], list[list[int]]]:
+        """Split this program set into program sets of at most ``max_executables`` executables,
+        alongside a map that records the position in the original program set of each executable
+        in each of the generated program sets.
```
sesmart, line 131: would add, i.e. indexed from (0, total_executables-1)

src/braket/program_sets/program_set.py
```diff
@@ -97,6 +98,167 @@ def total_shots(self) -> int:
...
+
+        When a single parameter set index of a ``CircuitBinding`` would by itself exceed
+        ``max_executables`` due to its observable list or ``Sum`` Hamiltonian being larger than
+        the budget, the observable list is split into chunks of at most ``max_executables`` entries
+        (``Sum`` summands are sliced with coefficients preserved). Observable splitting is only
+        performed when necessary; otherwise the full observable list or ``Sum`` is kept intact.
```
sesmart, line 137: Does this have consequences for max splitting? What is the worst case example here? I.e. CB of only observables that are at the border of being too big for 2 to fit in one? I.e. 10 executables but only a 19 ps split? This approaches ~0.5 filling efficiency for large N.

I think a flatten option might be nice, just so the user knows it has been split 1-1, though it adds complexity to the merge as well. Is there a simpler way to approach/think about this?

sesmart: Also, if it doesn't, that is okay - maybe we can just make it a bit more clear.

src/braket/program_sets/program_set.py
```diff
@@ -97,6 +98,167 @@ def total_shots(self) -> int:
...
+                current_size = 0
+            current.append(block)
+            current_size += block.size
+        sub, sub_map = self._build_program_set(current)
+        program_sets.append(sub)
+        index_map.append(sub_map)
+
+        return program_sets, index_map
+
+    def _executable_blocks(self, max_executables: int) -> list[_ExecutableBlock]:
```
sesmart, line 192: Sorry, this definitely needs a description 😢

sesmart: Rather, maybe a detailed description of the splitting approach.

src/braket/program_sets/program_set.py
sesmart: It also seems to occur on the other side of the approach - 10 executables, but only a 9 split limit causes -> 0.5 filling efficiency.

src/braket/program_sets/program_set.py
```diff
@@ -97,6 +98,167 @@ def total_shots(self) -> int:
...
+
+        if self.total_executables <= max_executables:
+            return [self], [list(range(self.total_executables))]
+
+        program_sets = []
+        index_map = []
+        current = []
+        current_size = 0
+        for block in self._executable_blocks(max_executables):
+            if current and current_size + block.size > max_executables:
```
sesmart, line 178: This is basically the majority of the logic, no? If A + next can fit, use it. Else, continue.

src/braket/program_sets/program_set.py
sesmart: It also just occured to me we kind of have a flatten...bind_observables_to_inputs would do something like this - but there is no, "unbind", or reorder. It would be cool if merge could take that as an input as well though, as it should also have the same ordering.

src/braket/program_sets/program_set.py
speller26: No strong opinion, but I did name it as a verb (i.e. method)

speller26:
Do not review program_set_quantum_task_result.py and test_program_set_quantum_task_result.py; those were approved in #1254 (I promise not to touch them).

src/braket/program_sets/program_set.py
speller26: The current implementation admittedly isn't maximally efficient, as "remainders" aren't combined together. I'd say this is good enough to get the job done though.

 As for explaining how it works in the docstring, I don't want to expose the details too much, since that can change (and in fact I'm already thinking of potential changes), and the interface is completely agnostic to the underlying algorithm.

src/braket/program_sets/program_set.py
speller26: Yep. It's not particularly sophisticated; I'm thinking of exploring better chunking algos later, but don't want to overhaul the current PR.

src/braket/program_sets/program_set.py
sesmart: Fair enough. I am of course biased towards a flatten 😁 but you could just as easily do that with an iterative loop right before the split.

src/braket/program_sets/program_set.py
```diff
@@ -97,6 +98,188 @@ def total_shots(self) -> int:
...
+        index_map.append(sub_map)
+
+        return program_sets, index_map
+
+    def _executable_blocks(self, max_executables: int) -> list[_ExecutableBlock]:
+        """Enumerate this program set's executables as a list of ``_ExecutableBlock``s in the order
+        of ``enumerate_executables``
+
+        Each block is a contiguous run of executables that share the same
+        ``(circuit, observable list/Sum Hamiltonian, single parameter assignment)`` and can thus be
```
sesmart, line 197:
```suggestion
        Each block is a contiguous run of executables generated from the same
        ``(circuit, observable list/Sum Hamiltonian, single parameter assignment)`` and can thus be
```

sesmart (approved):
Just a small tweak in the description. Otherwise looks good!

PR #1254, opened 2026-05-10
feat: Merge program set task results

speller26:
Introduce `merge` method to `ProgramSetQuantumTaskResult` that combines multiple results into one.

*Issue #, if available:*

*Description of changes:*

*Testing done:*

## Merge Checklist

#### General

#### Tests


speller26:
If this is approved before #1249, we can merge this cleanly into the `split` branch and assume the files are reviewed in that PR.

src/braket/tasks/program_set_quantum_task_result.py
```diff
@@ -370,6 +375,116 @@ def from_object(
...
+        if total_mapped != total_executables:
+            raise ValueError(
+                f"Index map covers {total_mapped} executables but the original program set "
+                f"has {total_executables}"
+            )
+
+        programs = [_binding_to_program(binding) for binding in program_set.entries]
+        executable_indices = list(program_set.enumerate_executables())
+        binding_executable_counts = [_count_executables(b) for b in program_set.entries]
+        shots_per_executable = results[0].entries[0].shots_per_executable
```
sesmart: Do we need to check this matches as well? Isn't this also at program_set.shots_per_executable?

src/braket/tasks/program_set_quantum_task_result.py
```diff
@@ -481,6 +596,85 @@ def _compute_num_executables(metadata: ProgramSetTaskMetadata) -> int:
         return counter
 
 
+def _binding_to_program(binding: CircuitBinding | Circuit) -> Program:
+    if isinstance(binding, Circuit):
+        return Program(source=binding.to_ir(IRType.OPENQASM).source, inputs=None)
+    return binding.to_ir()
+
+
+def _count_executables(binding: CircuitBinding | Circuit) -> int:
```
sesmart, line 689: Surprised CircuitBinding doesn't actually have a simple .executables property here.

src/braket/tasks/program_set_quantum_task_result.py
```diff
@@ -481,6 +596,85 @@ def _compute_num_executables(metadata: ProgramSetTaskMetadata) -> int:
...
+    if isinstance(binding, Circuit):
+        return 1
+    num_ps = len(binding.input_sets) if binding.input_sets is not None else 1
+    num_obs = len(binding.observables) if binding.observables is not None else 1
+    return num_ps * num_obs
+
+
+def _buffer_result(
```
sesmart: Can we add a description to this function? The other ones are straightforward but this has more complexity and `_buffer_result` is a bit vague.

src/braket/tasks/program_set_quantum_task_result.py
```diff
@@ -481,6 +596,85 @@ def _compute_num_executables(metadata: ProgramSetTaskMetadata) -> int:
...
+    program_set: ProgramSet,
+    programs: list[Program],
+    executable_indices: list[tuple[int, int, int]],
+    buffer: list[MeasuredEntry | ProgramSetExecutableFailure | None],
+) -> None:
+    j = 0
+    for composite in result.entries:
+        for entry in composite.entries:
+            if j >= len(map_k):
+                raise ValueError(
```
sesmart, line 710: Can this realistically happen? I guess if you have two programsets, where each one has the same total number of different numbers in the constituents?

src/braket/tasks/program_set_quantum_task_result.py
```diff
@@ -370,6 +375,116 @@ def from_object(
...
+                    additional_metadata=None,
+                )
+            )
+            start += count
+
+        metas = [r.task_metadata for r in results]
+        return ProgramSetQuantumTaskResult(
+            entries=entries,
+            task_metadata=ProgramSetTaskMetadata(
+                id=";".join(meta.id for meta in metas),  # Better way to do this?
```
sesmart, line 547: I wonder if there is a way to denote that this is just synthetic metadata, or an aggregate, and these are just convenience attributes. Unfortunately the best way for that would probably be a new schema, which seems excessive.

sesmart: Or if task_metadata could have a union type with `Sequence[ProgramSetTaskMetadata]`

src/braket/tasks/program_set_quantum_task_result.py
```diff
@@ -481,6 +596,85 @@ def _compute_num_executables(metadata: ProgramSetTaskMetadata) -> int:
...
+        return 1
+    num_ps = len(binding.input_sets) if binding.input_sets is not None else 1
+    num_obs = len(binding.observables) if binding.observables is not None else 1
+    return num_ps * num_obs
+
+
+def _buffer_result(
+    k: int,
+    result: ProgramSetQuantumTaskResult,
+    map_k: list[int],
```
sesmart: this is like the composite_entry_map?

src/braket/tasks/program_set_quantum_task_result.py
```diff
@@ -481,6 +596,85 @@ def _compute_num_executables(metadata: ProgramSetTaskMetadata) -> int:
...
+    for composite in result.entries:
+        for entry in composite.entries:
+            if j >= len(map_k):
+                raise ValueError(
+                    f"t=Task {result.task_metadata.id} at index {k} "
+                    "produced more executables than index map expects"
+                )
+            orig_idx = map_k[j]
+            binding_idx, ps_idx, obs_idx = executable_indices[orig_idx]
+            buffer[orig_idx] = _convert_measured_entry(
```
sesmart: So `map_k` is some contiguous list between [0] and the composite entry number of executables, right? j is essentially a subindex, or the rank-2 index. So why is the buffer at along this index when the original buffer is len(total_executables), i.e. unfolded?

src/braket/tasks/program_set_quantum_task_result.py
```diff
@@ -370,6 +375,116 @@ def from_object(
...
+                program_set=program_set,
+                programs=programs,
+                executable_indices=executable_indices,
+                buffer=buffer,
+            )
+
+        entries = []
+        start = 0
+        for binding_idx, binding in enumerate(program_set.entries):
+            count = binding_executable_counts[binding_idx]
```
sesmart: I think you can just calculate this directly from the binding here -`binding_executable_counts` isn't used elsewhere as far as I can see.

src/braket/tasks/program_set_quantum_task_result.py
speller26: I can add a flag that triggers a warning the first time the metadata is accessed.

src/braket/tasks/program_set_quantum_task_result.py
speller26: Mainly because there isn't really a class that encapsulates a single executable, so such a method would just return a list of tuples

src/braket/tasks/program_set_quantum_task_result.py
speller26: No, `map_k`'s sublist actually starts where the previous sublist ended; it's the index in the _parent_ program set. I'll rename accordingly.

src/braket/tasks/program_set_quantum_task_result.py
speller26: Map to the parent program set index

src/braket/tasks/program_set_quantum_task_result.py
speller26: Yeah, if you pass in the wrong merge list; this is just defensive, like the `if j != len(parent_indices)` below.

src/braket/tasks/program_set_quantum_task_result.py
```diff
@@ -141,7 +146,6 @@ def expectation(self) -> float | None:
         return self._expectation
 
 
-@dataclass
```
sesmart, line 144: Hm...I am okay with this, but also am just curious why a dataclass is no longer needed.

src/braket/tasks/program_set_quantum_task_result.py
```diff
@@ -370,6 +458,117 @@ def from_object(
...
+        results: Sequence[ProgramSetQuantumTaskResult],
+        program_set: ProgramSet,
+        index_map: list[list[int]],
+    ) -> ProgramSetQuantumTaskResult:
+        """Reconstruct a ``ProgramSetQuantumTaskResult`` from the task results produced by running
+        each program set of ``program_set.split(...)``.
+
+        ``index_map`` is the per-executable map returned alongside the program sets by
+        ``ProgramSet.split``: ``index_map[k][j]`` gives the index, in the order of ``program_set``,
+        of the executable that the jth executable of the kth task represents. The kth task's
```
sesmart, line 472: This is the k-th program, no? I.e. the k-th binding?

src/braket/tasks/program_set_quantum_task_result.py
```diff
@@ -370,6 +458,117 @@ def from_object(
...
+                programs=programs,
+                executable_indices=executable_indices,
+                buffer=buffer,
+            )
+
+        entries = []
+        start = 0
+        for binding_idx, binding in enumerate(program_set.entries):
+            count = _count_executables(binding)
+            program = programs[binding_idx]
```
sesmart, line 529: Why don't we just calculate the ir on the fly here? I.e. _binding_to_program(binding)? Feel like you don't need programs variable above.

src/braket/tasks/program_set_quantum_task_result.py
sesmart: Ah, no I see that it is actually the k-program set, so k-th task is appropriate.

src/braket/tasks/program_set_quantum_task_result.py
```diff
@@ -370,6 +458,117 @@ def from_object(
...
+                    ProgramMetadata(
+                        executables=[
+                            ProgramSetExecutableResultMetadata()
+                            for _ in range(_count_executables(b))
+                        ]
+                    )
+                    for b in program_set.entries
+                ],
+                deviceParameters=None,  # TODO: find a way to fill this in
+                createdAt=min(m.createdAt for m in metas if m.createdAt),
```
sesmart, line 561: This apparently fails on local sim, i.e. when this list is empty? 
```ValueError: min() iterable argument is empty```

File: src/braket/tasks/program_set_quantum_task_result.py [hunk 13]
```diff
@@ -370,6 +458,117 @@ def from_object(
...
+                endedAt=max(m.endedAt for m in metas if m.endedAt),
```
sesmart, line 562: Same as above

test/unit_tests/braket/tasks/test_program_set_quantum_task_result.py
```diff
@@ -437,3 +441,462 @@ def test_dispatch_executable_result_with_none_inputs(execution_measurement_proba
...
+    c1 = circuit_rx_parametrized_fixture
+    c2 = Circuit().rx(0, FreeParameter("phi"))
+    b1 = CircuitBinding(c1, {"theta": [0.1, 0.2, 0.3]}, observables=[X(0), Y(0)])  # 6 execs
+    b2 = CircuitBinding(c2, {"phi": [0.4, 0.5]})  # 2 execs, no observables
+    ps = ProgramSet([b1, b2])
+    assert ps.total_executables == 8
+
+    subs, index_map = ps.split(5)
+    # Greedy pack with max=5: b1 classes (sizes 2,2,2) fill [2+2=4, +2>5 flush], so
+    # sub 0 = 2 b1 classes (4 execs), sub 1 = 1 b1 class (2 execs) + b2 (2 execs) = 4 execs.
```
sesmart, line 679: I'm gonna look at this more with the split - I noticed inputs will not do this sort of greedy packing, so it's very interesting.

test/unit_tests/braket/tasks/test_program_set_quantum_task_result.py
```diff
@@ -437,3 +441,462 @@ def test_dispatch_executable_result_with_none_inputs(execution_measurement_proba
...
+    assert [s.total_executables for s in subs] == [4, 4]
+
+    def _failure(inputs_index):
```
sesmart, line 682: Is there any reasonable way to fix a failure like this btw?

src/braket/tasks/program_set_quantum_task_result.py
speller26: Because I want to emit a warning when someone tries to call `additional_metadata`; I need it to be a method rather than just a dataclass field.

src/braket/tasks/program_set_quantum_task_result.py
speller26: This is only really intended for `AwsDevice`; the local simulator doesn't actually need splitting

test/unit_tests/braket/tasks/test_program_set_quantum_task_result.py
speller26: Fix in what sense? Retry?

src/braket/tasks/program_set_quantum_task_result.py
sesmart: Hm....it might still be nice for validating / emulator usage, or use in other routines? Is it easy enough to catch?

test/unit_tests/braket/tasks/test_program_set_quantum_task_result.py
sesmart: Yeah, though I think my question applies more broadly to PS generally - I've yet to actually encounter a mid task failure. 

The scenario I've actually hit is that one of the split program sets can't validate and then the whole execution raises an error. Not quite related though.

src/braket/tasks/program_set_quantum_task_result.py
speller26: Not completely convinced that this would be useful for validation/emulation, since the local simulator also never returns this information; I could be convinced if we find a need to add that functionality to the local simulator, and I'm happy to add it then

src/braket/tasks/program_set_quantum_task_result.py
speller26: We use the `programs` variable in `_reorder_executable_results`

speller26: (which used to be called `buffer_results`; the new name should be a lot clearer)

sesmart: ohh, I missed that, okay.

PR #1278, opened 2026-06-04
feature: count circuit operation on circuit or given qubits

skushagra:
Closes Issue #1235 

*Description of changes:*
I have added a `count_instruction` functions that gives you count of an instrcution on a circuit. I have added test cases for the required flow.

#### Features
- Count instructions by operator type.
- Optional filtering by:
  - Operator name, operator class, or operator instance.
  - Qubits involved in an instruction.
  - Moment types (`GATE`, `GATE_NOISE`, `MEASURE` etc.).
- Supports multi-qubit filtering via:
  - `QubitMatch.ANY` — instruction on any specified qubit.
  - `QubitMatch.ALL` — instruction on all specified qubits.
- When both operator and qubit filters are provided, instructions must satisfy both filters (AND).

## Merge Checklist

#### General

#### Tests


skushagra:
I did not understand much around what can we do for calculating density of operations. I would need more clarification on how should we exactly define this density, and since there can be multiple ways with respect to which density can be calculated like, operations per layer, per qubit etc. 

Hence I extracted this into a seprate helper file to prevent bloating the Circuit class. Perhaps more clarification on this would help me to add such a features.

src/braket/circuits/circuit_analysis.py
```diff
@@ -0,0 +1,94 @@
...
+    operators: OperatorIdentifier | Iterable[OperatorIdentifier] | None,
+) -> list[str]:
+    if operators is None:
+        return []
+    if isinstance(operators, (str, type)) or isinstance(operators, Operator):
+        return [_normalize_operator_name(operators)]
+    return [_normalize_operator_name(op) for op in operators]
+
+
+def count_instructions(
```
aniksd-braket: why not make this a private function instead, or I do not mind moving it all to circuit.py

src/braket/circuits/circuit_analysis.py
```diff
@@ -0,0 +1,94 @@
...
+        include_types (Iterable[MomentType]): Moment types to count. Default: GATE only.
+            Pass additional MomentType values to include noise, measures, etc.
+
+    Returns:
+        Counter[str]: Operator names mapped to occurrence counts.
+    """
+    include_types_set = set(include_types)
+    operator_names_set = set(_to_operator_names(operators))
+    _qs = QubitSet(qubits) if qubits is not None else None
+    filter_qubits = _qs if _qs else None  # empty QubitSet treated as no filter
```
aniksd-braket: check for case when qubit specified is not part of the circuit

src/braket/circuits/circuit.py
```diff
@@ -243,6 +243,34 @@ def parameters(self) -> set[FreeParameter]:
...
+            >>> circ.count_instructions()
+            Counter({'H': 1, 'CNot': 1, 'Rx': 1})
+            >>> circ.count_instructions("h")
+            Counter({'H': 1})
+            >>> circ.count_instructions(["H", "CNot"])
+            Counter({'H': 1, 'CNot': 1})
+            >>> circ.count_instructions(qubits=0)
+            Counter({'H': 1, 'CNot': 1, 'Rx': 1})
+        """
+        from braket.circuits.circuit_analysis import count_instructions as _count_instructions
```
aniksd-braket: move all imports to the top

src/braket/circuits/circuit.py
```diff
@@ -243,6 +243,34 @@ def parameters(self) -> set[FreeParameter]:
         """
         return self._parameters
 
+    def count_instructions(self, operators=None, **kwargs) -> Counter[str]:
```
aniksd-braket: The function name should ideally be count()

src/braket/circuits/circuit_analysis.py
skushagra: we can move it all to circuit.py but this keeping(or renaming circuit_analysis.py) seems to be a very nice option that will keep the circuit class clean. But if it goes against the normal practives that we follow, I'll move it back.

aniksd-braket: Yes, I would prefer this to be in the circuit.py

aniksd-braket:
@skushagra could you also make sure to resolve all comments for the next updated CR

aniksd-braket:
@skushagra are you still working on this?

skushagra:
Yes I am, will resolve these comments in a bit

src/braket/circuits/circuit_analysis.py
skushagra: Moved all logic to circuit.py

src/braket/circuits/circuit.py
skushagra: this import is no longer required

src/braket/circuits/circuit.py
skushagra: renamed function to count

src/braket/circuits/circuit_analysis.py
skushagra: tested this, a empty counter object will be returned in this case.

skushagra:
@aniksd-braket fixed the pending comments, please have a look and let me know if anything else is needed

aniksd-braket (commented):
This looks good, getting closer to the final product. I am assuming you used AI for creating tests? some of them seem repetitive/unnecessary. can you trim it down and/or consolidate them?

aniksd-braket:
@skushagra just a bump on this

skushagra:
Yes, I'll consolidate and remove the unnecessary tests.

skushagra:
Hi @aniksd-braket 

I have updated the test file to keep only those that I feel were important and those that were discussed in your review. Let me know if there are any other instances that you'd like to capture here, will add them accordingly.

skushagra:
Fixed all codecov, linter and formatter issues

aniksd-braket:
Thank you for your effort @skushagra! One last request, could add some checks to ensure that errors are raised if an invalid qubit or operator name is passed in as an argument. For eg: a qubit that is not part of the circuit or a operator that is not defined. If the qubit is part of the circuit but has no gates on it, it should just return an empty counter. Similarly, if the operation is a valid braket operation but does not exist in the circuit it should return an empty counter

skushagra:
Sure, will add this

skushagra:
@aniksd-braket 

One question, how should I validate the operator names, qubit range is simple enough to check, but for operators etc. should i use `src/braket/circuits/translations.py` or something like `Operator.__subclasses__()` to find all at runtime?

aniksd-braket:
@skushagra yes translations should be a good reference

skushagra:
what about when users want to count some custom implementaiton, I think that will be a common usecase? 

Edit: Added the quibit check, working on how should I validate operators

aniksd-braket:
> what about when users want to count some custom implementaiton, I think that will be a common usecase?

Ah good point. Considering this, looking at subclasses at runtime might be the way to go

aniksd-braket:
@skushagra bump on this. I will have to assign winners to the issue today

skushagra:
Yep, finishing this in a bit.

skushagra:
I tried two ways, one keeping the names in the `Operator` class and other was to keep eveything in `circuit.py`.  Right now I am keeping everything in the circuit.py, let me know if the other approach looks better.

aniksd-braket:
Hi @skushagra! Looks good, couple of minor points
  - Test for passing an empty list to operators or qubits  
  - Test for include_types=[] (should return empty counter)
  - Test for circuits with compiler directives (like barrier)
  - Test for creating operator at runtime, and then using that as argument to count()

skushagra:
@aniksd-braket updated the tests as needed

aniksd-braket:
Looks great. I will be assigning you the issue. Congratulations!

skushagra:
Thanks! Looking forward to making more contributions!

aniksd-braket (approved):

PR #1323, opened 2026-07-15
feat: Per-task shots in `run_batch`

speller26:
*Issue #, if available:*

*Description of changes:*

*Testing done:*

## Merge Checklist

#### General

#### Tests

### DIFF
--- a/src/braket/aws/aws_quantum_task_batch.py
+++ b/src/braket/aws/aws_quantum_task_batch.py
@@ -14,6 +14,7 @@
 from __future__ import annotations
 
 import time
+from collections.abc import Sequence
 from concurrent.futures.thread import ThreadPoolExecutor
 from itertools import repeat
 from typing import Any
@@ -31,6 +32,12 @@ from braket.pulse.pulse_sequence import PulseSequence
 from braket.registers.qubit_set import QubitSet
 from braket.tasks.quantum_task import TaskResult, TaskSpecification
 from braket.tasks.quantum_task_batch import QuantumTaskBatch
+
+_TaskInputsGateDefinitions = tuple[
+    TaskSpecification,
+    dict[str, float],
+    dict[tuple[Gate, QubitSet], PulseSequence],
+]
 
 
 class AwsQuantumTaskBatch(QuantumTaskBatch):
@@ -53,7 +60,7 @@ class AwsQuantumTaskBatch(QuantumTaskBatch):
         device_arn: str,
         task_specifications: TaskSpecification | list[TaskSpecification],
         s3_destination_folder: AwsSession.S3DestinationFolder,
-        shots: int,
+        shots: int | Sequence[int],
         max_parallel: int,
         max_workers: int = MAX_CONNECTIONS_DEFAULT,
         poll_timeout_seconds: float = AwsQuantumTask.DEFAULT_RESULTS_POLL_TIMEOUT,
@@ -79,10 +86,11 @@ class AwsQuantumTaskBatch(QuantumTaskBatch):
             s3_destination_folder (AwsSession.S3DestinationFolder): NamedTuple, with bucket
                 for index 0 and key for index 1, that specifies the Amazon S3 bucket and folder
                 to store quantum task results in.
-            shots (int): The number of times to run the quantum task on the device. If the device is
-                a simulator, this implies the state is sampled N times, where N = `shots`.
-                `shots=0` is only available on simulators and means that the simulator
-                will compute the exact results based on the quantum task specification.
+            shots (int | Sequence[int]): The number of times to run each quantum task on the
+                device. If an integer, this value is used for every task. If a sequence, each
+                element is used for the corresponding task. On a simulator, `shots=0` means
+                that the simulator will compute the exact results based on the quantum task
+                specification.
             max_parallel (int): The maximum number of quantum tasks to run on AWS in parallel.
                 Batch creation will fail if this value is greater than the maximum allowed
                 concurrent quantum tasks on the device.
@@ -108,18 +116,19 @@ class AwsQuantumTaskBatch(QuantumTaskBatch):
                 capability context will be used if active. Default: None.
             **aws_quantum_task_kwargs (Any): Arbitrary kwargs for `QuantumTask`.,
         """  # noqa: E501
+        tasks_inputs_gatedefs = AwsQuantumTaskBatch._tasks_inputs_gatedefs(
+            task_specifications, inputs, gate_definitions
+        )
         self._tasks = AwsQuantumTaskBatch._execute(
             aws_session,
             device_arn,
-            task_specifications,
+            tasks_inputs_gatedefs,
             s3_destination_folder,
             shots,
             max_parallel,
             max_workers,
             poll_timeout_seconds,
             poll_interval_seconds,
-            inputs,
-            gate_definitions,
             reservation_arn,
             experimental_capabilities,
             *aws_quantum_task_args,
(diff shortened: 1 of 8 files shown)


src/braket/aws/aws_quantum_task_batch.py
```diff
@@ -32,6 +33,12 @@
 from braket.tasks.quantum_task import TaskResult, TaskSpecification
 from braket.tasks.quantum_task_batch import QuantumTaskBatch
 
+_TaskInputsGateDefinitions = tuple[
```
laurencap: Should this be a [type alias](https://typing.python.org/en/latest/spec/aliases.html)?

Also, since you're making a type, maybe comment what these two `dict`s are supposed to be so that the type can be understood without searching around.

Separate question: why not use a dataclass here?

src/braket/aws/aws_quantum_task_batch.py
```diff
@@ -108,18 +116,19 @@ def __init__(
                 capability context will be used if active. Default: None.
             **aws_quantum_task_kwargs (Any): Arbitrary kwargs for `QuantumTask`.,
         """  # noqa: E501
+        tasks_inputs_gatedefs = AwsQuantumTaskBatch._tasks_inputs_gatedefs(
+            task_specifications, inputs, gate_definitions
```
laurencap, line 120: I like the idea of having more structure for the task spec since there are so many parameters all being passed around.

The `task_specification` is essentially the program description, right? (like OpenQASM or ProgramSet)

I think it's out of scope to address this here -- but it does feel a little backwards. It seems as though a `TaskSpecification` should be a container with all the task parameters, including inputs and gate defs, and maybe shots and device arn for that matter.

Instead we have this "task program and inputs and gatedefs" container. It should probably be called the `TaskSpecification` but that name is already taken.

What do you think? Would it be better to later refactor all the parameters together into a single container, but for now keep these parameters separate and remove `_TaskInputsGateDefinitions`?

laurencap: Not a blocking comment.

test/unit_tests/braket/aws/test_aws_device.py
```diff
@@ -1609,6 +1609,29 @@ def test_run_batch_with_shots(
...
+    aws_session_mock,
+    boto_session_mock,
+    device,
+):
+    task_mock = Mock()
+    task_mock.state.return_value = "COMPLETED"
+    aws_quantum_task_mock.return_value = task_mock
+
+    shots = [10, 20, 30]
+    device("arn:aws:braket:::device/quantum-simulator/amazon/sim").run_batch(
```
laurencap, line 1626: So, this doesn't use `ProgramSet`s at all, right? It creates 3 separate single tasks?

The code doesn't touch the relevant lines so it's hard to tell from this change alone.

test/unit_tests/braket/aws/test_aws_quantum_task_batch.py
```diff
@@ -73,6 +73,88 @@ def test_creation_with_experimental_capabilities_string(mock_create):
...
+        AwsQuantumTaskBatch(
+            Mock(),
+            "foo",
+            _circuits(2),
+            S3_TARGET,
+            [100, 200],
+            max_parallel=1,
+        )
+
+    normalize_mock.assert_called_once()
```
laurencap: Not checking the result? What is this test validating?

test/unit_tests/braket/aws/test_aws_quantum_task_batch.py
```diff
@@ -73,6 +73,88 @@ def test_creation_with_experimental_capabilities_string(mock_create):
...
+    batch = AwsQuantumTaskBatch(
+        Mock(),
+        "foo",
+        _circuits(len(shots)),
+        S3_TARGET,
+        shots,
+        max_parallel=1,
+    )
+
+    assert batch.size == len(shots)
```
laurencap, line 93: Isn't this trivial from the test construction? What is this testing? The `size` method wasn't added in this PR, right?

test/unit_tests/braket/aws/test_aws_quantum_task_batch.py
```diff
@@ -73,6 +73,88 @@ def test_creation_with_experimental_capabilities_string(mock_create):
...
+    assert [call.args[4] for call in mock_create.call_args_list] == shots
+
+
+@patch("braket.aws.aws_quantum_task.AwsQuantumTask.create")
+def test_creation_with_shot_sequence_mismatch(mock_create):
+    task_mock = Mock()
+    task_mock.state.return_value = "COMPLETED"
+    mock_create.return_value = task_mock
+
+    with pytest.raises(ValueError, match="Multiple shots and quantum tasks"):
```
laurencap: This doesn't sound like a useful error message. Shouldn't it be "mismatch in the number of shots and tasks" or similar?

test/unit_tests/braket/aws/test_aws_quantum_task_batch.py
```diff
@@ -178,6 +260,35 @@ def test_retry(mock_create):
         batch.retry_unsuccessful_tasks()
 
 
+@patch("braket.aws.aws_quantum_task.AwsQuantumTask.create")
+def test_retry_with_shot_sequence(mock_create):
```
laurencap, line 243: This looks like it's testing retry instead of the new shots logic. Can you help me understand how these two processes interact?

laurencap (changes requested):

src/braket/aws/aws_quantum_task_batch.py
speller26: That refactor does make sense; I'll think about what other things might be worth packaging up. For the purposes of this PR though, I'll keep this as a private method rather than introducing a new type.

test/unit_tests/braket/aws/test_aws_quantum_task_batch.py
speller26: This is just to ensure that the retry uses the correct shot count for the retried task

test/unit_tests/braket/aws/test_aws_device.py
speller26: This isn't just for program sets; you can run individual circuits with different shot counts now.

test/unit_tests/braket/aws/test_aws_quantum_task_batch.py
speller26: Yeah, it's pretty redundant; it's just a sanity check to ensure that the shot vector didn't screw up the batch size (for example, running each task `len(shots)` times)

test/unit_tests/braket/aws/test_aws_quantum_task_batch.py
speller26: I can change it to the full "Multiple shots and quantum tasks must be equal in number." message

test/unit_tests/braket/aws/test_aws_quantum_task_batch.py
speller26: Just another sanity check that we're not multipying the shot vector across the tasks; this test might be redundant

src/braket/aws/aws_quantum_task_batch.py
speller26: No strong opinions, but a dataclass does make sense; I'll do that.

speller26: Oh right, now I remember; it's to make iterating easier via zipping.

laurencap (approved):

Issue #1322, opened 2026-07-13
Inconsistency with FreeParameterExpression


sesmart:
**Describe the bug**
FreeParameter expression can output to functions like `log(alpha)`, but cannot parse them, as [it is missing support for](https://github.com/amazon-braket/amazon-braket-sdk-python/blob/39de577dbf81ba23c113bc29230b75d9ebeaf49f/src/braket/parametric/free_parameter_expression.py#L83) `Ast.Call`

**To reproduce**

```
from braket.parametric import exp, FreeParameterExpression, FreeParameter

from_sympy = FreeParameterExpression(exp(FreeParameter('alpha')))
from_str   = FreeParameterExpression(from_sympy.__str__())
```

**Expected behavior**

Parity between the two. 

**Screenshots or logs**

```ValueError: Unsupported string detected: <ast.Call object at 0x10e571190>```

**System information**
A description of your system. Please provide:
- **Amazon Braket Python SDK version**:  1.123.0.post0
- **Amazon Braket Python Schemas version**:  1.29.1
- **Amazon Braket Python Default Simulator version**: 1.39.5
- **Python version**: 3.13

**Additional context**

sesmart:
Closed by #1328.

PR #1328, opened 2026-07-23
fix: parse function calls in FreeParameterExpression

CleanDev-Fix:
## Summary

Fix parsing of OpenQASM-compatible function calls in `FreeParameterExpression`.

## Changes

- Map the existing OpenQASM math helper names to their SymPy functions.
- Recursively evaluate positional call arguments while rejecting unknown functions and keyword calls.
- Add regression coverage for `exp`, `log`, `sqrt`, `mod`, and the unsupported-function boundary.

## Testing/Verification

- `PYTHONPATH=src .venv/Scripts/python.exe -m pytest -o addopts='' test/unit_tests/braket/parametric/test_free_parameter_expression.py -q` (58 passed, 4 xfailed)
- `ruff format --check` on modified files (pass)
- `ruff check src/braket/parametric/free_parameter_expression.py` (pass)
- `python -m compileall -q` on modified files (pass)
- `git diff --check` (pass)
- The full unit suite was attempted but collection is blocked in this isolated environment by missing generated/runtime packages (`braket.ir`, `braket.task_result`) after the editable test install hit Windows long-path limits; no full-suite failure is attributed to this patch.

Related issue: #1322

### DIFF
--- a/src/braket/parametric/free_parameter_expression.py
+++ b/src/braket/parametric/free_parameter_expression.py
@@ -15,6 +15,7 @@ from __future__ import annotations
 
 import ast
 import operator
+from collections.abc import Callable
 from functools import reduce
 from numbers import Number
 from typing import Any, ClassVar
@@ -97,6 +98,21 @@ class FreeParameterExpression:
         else:
             raise NotImplementedError
 
+    _STRING_FUNCTIONS: ClassVar[dict[str, Callable[..., Any]]] = {
+        "sin": sympy.sin,
+        "cos": sympy.cos,
+        "tan": sympy.tan,
+        "arcsin": sympy.asin,
+        "arccos": sympy.acos,
+        "arctan": sympy.atan,
+        "exp": sympy.exp,
+        "log": sympy.log,
+        "sqrt": sympy.sqrt,
+        "mod": sympy.Mod,
+        "ceiling": sympy.ceiling,
+        "floor": sympy.floor,
+    }
+
     @property
     def expression(self) -> Number | sympy.Expr:
         """Gets the expression.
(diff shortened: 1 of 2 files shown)


src/braket/parametric/free_parameter_expression.py
```diff
@@ -97,6 +98,21 @@ def __init__(self, expression: FreeParameterExpression | Number | sympy.Expr | s
         else:
             raise NotImplementedError
 
+    _STRING_FUNCTIONS: ClassVar[dict[str, Callable[..., Any]]] = {
```
sesmart: I would replace this and _FN_MAP around line 30 with a single dict at the top and then a one-line inversion of the dictionary. Probably doesn't need to be interior to the class.

src/braket/parametric/free_parameter_expression.py
```diff
@@ -144,6 +160,15 @@ def _eval_operation(self, node: Any) -> FreeParameterExpression:
             return FreeParameterExpression(node.value)
         if isinstance(node, ast.Name):
             return FreeParameterExpression(sympy.Symbol(node.id))
+        if isinstance(node, ast.Call):
+            if not isinstance(node.func, ast.Name) or node.keywords:
+                raise ValueError(f"Unsupported string detected: {node}")
+            function = self._STRING_FUNCTIONS.get(node.func.id)
+            if function is None:
+                raise ValueError(f"Unsupported string function: {node.func.id}")
+            return FreeParameterExpression(
```
sesmart: Maybe raise an error if a kwarg is supplied? I think only supplying args is fine, but wouldn't want it to miss a kwarg.

src/braket/parametric/free_parameter_expression.py
```diff
@@ -144,6 +160,15 @@ def _eval_operation(self, node: Any) -> FreeParameterExpression:
             return FreeParameterExpression(node.value)
         if isinstance(node, ast.Name):
             return FreeParameterExpression(sympy.Symbol(node.id))
+        if isinstance(node, ast.Call):
+            if not isinstance(node.func, ast.Name) or node.keywords:
+                raise ValueError(f"Unsupported string detected: {node}")
+            function = self._STRING_FUNCTIONS.get(node.func.id)
+            if function is None:
+                raise ValueError(f"Unsupported string function: {node.func.id}")
```
sesmart: Can we make this a (slightly) more descriptive error message? And above?

sesmart (commented):
Thanks for the PR @CleanDev-Fix  - looks good, just a few small comments.

sesmart (approved):

PR #1352, opened 2026-09-12
fix: treat a single OpenQASM string or Circuit as one ProgramSet entry

alyabouzaid:
### Problem

`ProgramSet.__init__` wraps only `CircuitBinding` and `Program` in a list. A bare OpenQASM string is therefore treated as a sequence and iterated character by character, silently:

```python
qasm = "OPENQASM 3.0;\nbit[2] b;\nqubit[2] q;\nh q[0];\ncnot q[0], q[1];\nb = measure q;"
ps = ProgramSet(qasm)
ps.total_executables        # 75  (== len(qasm))
ps.to_ir().programs[0]      # Program(source='O')
```

No error is raised; the malformed program set is only caught at the service. A bare `Circuit` hits the same line and raises `TypeError: 'Circuit' object is not iterable`.

Both types are already valid list entries, and a standalone `Program` is already supported (`test_standalone_openqasm_program`), so a standalone string or circuit is a natural thing to write.

### Fix

Include `Circuit` and `str` in the `isinstance` check that wraps a single entry, and widen the type hint/docstring accordingly. `ProgramSet(qasm)` now equals `ProgramSet([qasm])`, and a single `Circuit` with result types now raises the intended `ValueError: Circuit cannot have result types`.

### Verification

- Added `test_standalone_openqasm_string` and `test_standalone_circuit`, mirroring the existing `test_standalone_openqasm_program`. Both fail on `main` (75-character split / `TypeError`) and pass with the fix.
- `pytest test/unit_tests`: 3492 passed, 131 xfailed.
- `ruff check src`, `ruff format --check .`, `mypy`: clean.
- Exercised manually: `entries`, `len`, `total_executables`, `total_shots`, `to_ir`, `enumerate_executables`, `split`, `__eq__`, `__add__` on a standalone string and a standalone circuit, plus the unchanged `CircuitBinding` / `Program` / mixed-list paths.

AI disclosure: this PR was prepared with the assistance of Claude Code (Anthropic's Claude); I review and test all changes.


sesmart (approved):
Simple fix, nice!

